PR #1417, opened 2016-04-30
Issue 1303 fix unicode static

### LINKED ISSUE #1303: error_page.404 utf-8 handling error

---

Hi, i have russian chars encoded with utf-8 in 404.html

When referring to a nonexistent page get an "500 Internal Server Error"

```
#!python

The server encountered an unexpected condition which prevented it from fulfilling the request.

  File "C:\Work\Internet\library\python\lib\site-packages\cherrypy\_cprequest.py", line 664, in respond
    inst.set_response()
  File "C:\Work\Internet\library\python\lib\site-packages\cherrypy\_cperror.py", line 376, in set_response
    message=self._message).encode('utf-8')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 149: ordinal not in range(128)
Powered by CherryPy 3.2.5
```

Config:

```
#!python

cherrypy.config.update({
        'global': {
             ...
            'error_page.404': r'%s/public/internet/errors/404.html' % WORKING_DIRECTORY
        }
    })
```

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1303

jaraco:
Pull request ported from https://bitbucket.org/cherrypy/cherrypy/pull-requests/119.

---

There's a bug related to serving files with Unicode names. I've fixed it and added corresponding test.

P.S. Test filename and its contents are in Ukrainian, since it's my native language. It includes Cyrillic symbols, which are present in lots of Slavic languages (like Belarusian, Russian etc.)

Tested against 2.7 and 3.5:

```console
wk@wk-laptop /home/wk/src/cherrypy/cherrypy/test                                                                                                                                               
⚡ . ~/src/pyenv-2.7/bin/activate                  
(env) <= Python 2.7

wk@wk-laptop /home/wk/src/cherrypy/cherrypy/test                                                                                                                                               
⚡ nosetests -s test_static:StaticTest.test_unicode
.
----------------------------------------------------------------------
Ran 1 test in 0.118s

OK
(env) <= Python 2.7

wk@wk-laptop /home/wk/src/cherrypy/cherrypy/test                                                                                                                                               
⚡ . ~/src/pyenv-3.5/bin/activate                  
(env) <= Python 3.5

wk@wk-laptop /home/wk/src/cherrypy/cherrypy/test                                                                                                                                               
⚡ nosetests -s test_static:StaticTest.test_unicode
.
----------------------------------------------------------------------
Ran 1 test in 0.120s

OK
```


jaraco:
I have some concerns with the switch from ISO-8859-1 to UTF-8, not because I think the world should use ISO-8859-1 (I don't), but because IIRC the choice of that encoding was based on w3c standards. I would feel more comfortable accepting this change if it could be shown that newer standards recommend UTF-8 where they once recommended ISO-8869-1 or that UTF-8 is clearly the preferred encoding today and will minimally impact real-world applications.

One factor to consider is that ISO-8859-1 will decode all bytestreams, whereas many bytestreams will result in a UnicodeDecodeError when decoded in UTF-8.

How can we have confidence that this change won't cause massive disruption in existing applications?

webknjaz:
Hi @jaraco,
That is a good point. Still, do you have other suggestions on how should this issue be addressed?

jaraco:
I recommend researching into the relevant HTTP specs or at least referencing other frameworks that encode using UTF-8 where cherrypy currently encodes as Latin-1.

Alternately, provide an analysis of why this approach is sound and preferable and a description of which users might be impacted and what they must do to account for the change.

webknjaz:
Well, at django they fallback to latin-1 if an exception happens:
https://github.com/django/django/blob/master/django/http/request.py#L368-L391

jaraco:
After looking at this again, I'm pretty sure it's not acceptable. The switch of the `ntob` function to use UTF-8 in place of latin-1 is going to have an impact beyond the scope indicated in this pull request, and I'm not comfortable that the existing test suite has sufficient coverage of the unicode handling.

On the other hand, I do see that the tests were failing, which does suggest that there is at least some test coverage around the encoding.

Especially given that CherryPy now has a unified wsgiserver (which conflicts with this patch), I'm going to decline this patch, but I want to encourage you to give it another go. I see two possible approaches that I'd prefer to see:
- Surgically address the issue of Unicode support of static files (only). This approach may require some hacks to workaround encoding assumptions at different layers.
- Re-write CherryPy to prefer UTF-8 and all Unicode values by default. This backward-incompatible change should include instructions to upgrading applications on what changes would be required to their application to accommodate this change.

webknjaz:
@jaraco okay, I'll try crafting a new PR, taking in account your advices :)

PR #1418, opened 2016-05-01
Add a README.md

nicklasb:
Add a first version of a github-style readme.

Table of contents is generated using [doctoc](https://github.com/thlorenz/doctoc), to update:
```
doctoc README.md --notitle
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Welcome to the GitHub-repository of CherryPy!
 
 # Help
 
-What are my options if I feel I need help?
+What are my options if I feel I need help? 
 
 ### I don't understand the documentation
 While CherryPy is one of the easiest and most intuitive frameworks out there, the prerequisite for understanding the [CherryPy documentation](http://docs.cherrypy.org/en/latest/) is that you have a general understanding of Python and web development.<br />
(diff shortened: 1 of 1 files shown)


jaraco:
I can't help but feel like there's a lot of repetition here between this file and the documentation and the web site. I'd hate to create yet another place where documentation is maintained. Perhaps it would be better if the readme simply provided references to the web site and docs?

Also, the creation of this new file runs in conflict with an existing README.txt (which one should a user read?). Why not delete that other file?

Furthermore, use of markdown, while preferred on Github, isn't particularly useful for referencing in a PyPI long description or Sphinx docs, which require reStructuredText. How about instead renaming README.txt to README.rst, so it's rendered by Github, and update that document with the recommended intro material?

nicklasb:
The only documentation it contains is a short section describing the installation of the library.
I see no reason to reference the unrendered github frontpage, which the readme really is instead of an actual readme, from any documentation.
For me following convention weighs very heavily(when in Rome...).
But I am certainly open for discussion.

nicklasb:
Thing is that it seems to me that the github frontpage is the most important page on github based projects. The actual home page plays sort of second fiddle. In CherryPy:s case, that doesn't seem to be the case yet, that's why I thought it was important to update.
At least it has been so on the projects I am involved with.

As you reach the issues and everything from that page anyway, it seems to be the main entrance page for users when interacting with the project.

Just to make that clear, I do not want to be too redundant either, that is why only things that almost never change is appropriate on the github front page. So having the lots of non-generated documentation there would cause a lot of non-technical spelling fix and layout pollution in the commit history.

PR #1425, opened 2016-05-01
Add bitbucket-variation and hello world

nicklasb:
I just realized that the text at bitbucket is a really nice intro, varied it a bit and added it.
Also, the hello world example is both brief and shows the simple elegance of CherryPy, and instructive as it really is that easy to work with.

However, I used decorators instead of manually assigning the value, for two reasons:
1. I think it looks better, and the decorators are with the function that is exposed.
2. Feels like a more modern approach and suggests that more decorators add more functionality.

Not sure, though. Any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,22 +62,21 @@ Questions and their answers have great value for the community, and a tip is to 
 Examples are strongly encouraged.
 
 ### I have found a bug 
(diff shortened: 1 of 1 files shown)


nicklasb:
I have made some more fixes now, and I feel pretty good about it text-wise.

WRT decorators, the official docs use them, so that feels actually pretty correct now.

Issue #1420, opened 2016-05-01
LimitRequestQueueTests.test_gc failing intermittently


jaraco:
Here's [an run](https://travis-ci.org/cherrypy/cherrypy/jobs/126996462) where test_gc failed.

PR #1428, opened 2016-05-02
Fix #1311: decorator for registering tools.

### LINKED ISSUE #1311: Decorator for registering tools.
Originally reported by: **coady (Bitbucket: [coady](http://bitbucket.org/coady), GitHub: [coady](http://github.com/coady))**

----------------------------------------

For simple function tools, they're almost always registered immediately after defining them, using the same name.  Even the example in the docs:
```
#!python
def protect(users):
    if cherrypy.request.login not in users:
        raise cherrypy.HTTPError("401 Unauthorized")
cherrypy.tools.protect = Tool('on_start_resource', protect)
```
This pattern seems like a perfect use case for a decorator.  Such as:
```
#!python
@cherrypy.tools.register('on_start_resource')
def protect(users):
    if cherrypy.request.login not in users:
        raise cherrypy.HTTPError("401 Unauthorized")
```

----------------------------------------
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1311

coady:


jaraco:
LGTM. I like it.

@coady, can you add some documentation? Since this is a user-facing feature, it at least needs to be described in the API docs. Better would be to update the examples to make use of this feature when possible.

Also, could you add a note in the CHANGES.rst describing user notes for the planned release?

coady:
Docs updated.  ~~Test failure appears to be #1420.~~

jaraco:
Love it! Thank you.

PR #1430, opened 2016-05-04
Add TravisCI badge to README

webknjaz:
Is this a right place to place this badge?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
-
-
 # CherryPy
 
 Welcome to the GitHub-repository of [CherryPy](http://cherrypy.org/)! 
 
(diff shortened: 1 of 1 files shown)


nicklasb:
Hm, well no. It is usually at the top; here: 
image

Like it is here: https://github.com/json-schema-form/angular-schema-form
Not in the middle if the text like that, and no "CI:" (if I am looking at the right repo).

And also, it feels sort of annoying that it is added right now when the build is failing. :-)

webknjaz:
Okay, I've upgraded the location of the badge. You may accept that once tests go green :)

nicklasb:
Haha, that's good, I'll do that. :-)

Issue #1194, opened 2012-12-21
[Python3.2] [CherryPy 3.2.2] encoding of the request path error on Windows 7


ghost:
Originally reported by: **ywzhaiqi (Bitbucket: [ywzhaiqi](http://bitbucket.org/ywzhaiqi), GitHub: [ywzhaiqi](http://github.com/ywzhaiqi))**

---

When the url contains Chinese, like "你", an error occurs.

Error:
    404, "The path '/ä½\xa0' was not found."

I added 2 lines on cherrypy/_cpwsgi.py line 310, then it works ok.
I do not know why.

```
#!python
if py3k:
            # This isn't perfect; if the given PATH_INFO is in the wrong encoding,
            # it may fail to match the appropriate config section URI. But meh.
            old_enc = self.environ.get('wsgi.url_encoding', 'ISO-8859-1')
            new_enc = self.cpapp.find_config(self.environ.get('PATH_INFO', ''),
                                             "request.uri_encoding", 'utf-8')
            if new_enc.lower() != old_enc.lower():
                # Even though the path and qs are unicode, the WSGI server is
                # required by PEP 3333 to coerce them to ISO-8859-1 masquerading
                # as unicode. So we have to encode back to bytes and then decode
                # again using the "correct" encoding.
                try:
                    u_path = path.encode(old_enc).decode(new_enc)
                    u_qs = qs.encode(old_enc).decode(new_enc)
                    #  two lines added by me
                    u_path = u_path.encode(old_enc).decode(new_enc)
                    u_qs = u_qs.encode(old_enc).decode(new_enc)
                except (UnicodeEncodeError, UnicodeDecodeError):
                    # Just pass them through without transcoding and hope.
                    pass
                else:
                    # Only set transcoded values if they both succeed.
                    path = u_path
                    qs = u_qs

```

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1194

PR #1431, opened 2016-05-09
Fix unicode characters in path on Python 3

poofeg:
Variables PATH_INFO and SCRIPT_NAME are already ISO-8859-1 string. Old code makes these variables double encoded. Issue #1194


jaraco:
I feel a little uneasy about these changes as they seem to remove the explicit intention to honor WSGI 1.x. On the other hand, when I search for WSGI u.0, I don't find anything useful to reference. Furthermore, the tests seem to pass without all of this special handling.

So after some more consideration, I'm ready to push this out.

Thanks for the contrib.

PR #1432, opened 2016-05-10
Fix unicode url in HTTPRedirect on Python 3

poofeg:
If Location header is unicode string, it will be encoded with rules of RFC 2047 later. Browsers do not understand Location header like =?utf-8?b?....


jaraco:
This defect probably demands a unit test to capture the failure and prove the fix.

poofeg:
Done.

jaraco:
LGTM

PR #1433, opened 2016-05-12
Correct popargs example

sapid:
Match the code to the explanation of the code:
`take the first segment and store it into a parameter name band`

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -164,20 +164,20 @@ understand the structure of your URL.
 
     import cherrypy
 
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks @sapid. I think I'd prefer to correct the documentation, as the existing implementation is clearer. Here's my rationale - the argument is an attribute of the class handling the request. An `Album` has a `title` and similarly a `Band` has a `name`. It's awkward for a `Band` to have a `band` or to reference the `band` of a `Band`. Can you instead update the documentation to be more clear about the code?

PR #1434, opened 2016-05-18
Fix decoding of non-ASCII filenames in multipart/form-data payloads

ngrilly:
This commit fixes decoding of non-ASCII filenames in multipart/form-data payloads.

This commit also removes the class attribute Part.attempt_charsets 
for two reasons:
- The default encodings, US-ASCII and UTF-8, are essentially redundant, 
  since UTF-8 is ASCII-compatible.
- There is no way to replace the default encodings with something
  specific to the application or request. The encoding tool overrides
  the default in Entity.attemp_charsets, but not the default in
  Part.attempt_charsets.

### DIFF
--- a/cherrypy/_cpreqbody.py
+++ b/cherrypy/_cpreqbody.py
@@ -210,7 +210,7 @@ def process_multipart(entity):
 
     # Read all parts
     while True:
-        part = entity.part_class.from_fp(entity.fp, ib)
+        part = entity.part_class.from_fp(entity.fp, ib, entity.default_charset)
         entity.parts.append(part)
         part.process()
         if part.fp.done:
@@ -320,6 +320,17 @@ class Entity(object):
     in an ``on_start_resource`` tool.
     """
 
+    default_charset = 'utf-8'
+    """A string representing the character set encoding used for multipart/form-data payloads.
+    Defaults to utf-8.
+
+    References:
+    https://tools.ietf.org/html/rfc7578#section-5.1.3
+    https://www.w3.org/TR/html5/forms.html#multipart-form-data
+    """
+    # TODO: `default_charset` could probably be used to decode application/x-www-form-urlencoded
+    # as well, and maybe even replace `attempt_charsets`.
+
     # http://tools.ietf.org/html/rfc2046#section-4.1.2:
     # "The default character set, which must be assumed in the
     # absence of a charset parameter, is US-ASCII."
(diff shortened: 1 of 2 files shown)


cherrypy/_cpreqbody.py
```diff
@@ -589,18 +585,19 @@ class Part(Entity):
     module in Python's standard library.
     """
 
-    def __init__(self, fp, headers, boundary):
+    def __init__(self, fp, headers, boundary, charset):
```
jaraco, line 588: Here you've changed the signature of `Part` and `Part.from_fp`, adding a new, required parameter, but I don't see anywhere in the code invocation of these functions is adapted.

cherrypy/_cpreqbody.py
```diff
@@ -553,20 +563,6 @@ class Part(Entity):
 
     """A MIME part entity, part of a multipart entity."""
 
-    # "The default character set, which must be assumed in the absence of a
```
jaraco, line 556: You've deleted this comment, which based on the quotes leads me to believe it's a reference in an RFC, so it seems relevant.

cherrypy/_cpreqbody.py
```diff
@@ -553,20 +563,6 @@ class Part(Entity):
...
-    # charset parameter, is US-ASCII."
-    attempt_charsets = ['us-ascii', 'utf-8']
-    """A list of strings, each of which should be a known encoding.
```
jaraco, line 559: As `attempt_charsets` is now defined in `__init__`, this docstring is now lost, so the meaning of this attribute is now replaced with an explicit definition to an implicit one. As complicated as this issue is, it's important to be as explicit and descriptive as possible.

cherrypy/_cpreqbody.py
```diff
@@ -589,18 +585,19 @@ class Part(Entity):
     module in Python's standard library.
     """
 
-    def __init__(self, fp, headers, boundary):
+    def __init__(self, fp, headers, boundary, charset):
+        self.attempt_charsets = [charset]
```
jaraco, line 589: What's the purpose of having a list of attempted charsets if there's always exactly one?

cherrypy/_cpreqbody.py
```diff
@@ -616,11 +613,11 @@ def read_headers(cls, fp):
 
             if line[0] in ntob(' \t'):
                 # It's a continuation line.
-                v = line.strip().decode('ISO-8859-1')
+                v = line.strip().decode(charset, 'replace')
```
jaraco, line 616: This looks dangerous to me - using the 'replace' directive in decode is lenient, but almost never what the user intended.

jaraco:
This PR seems really close to an acceptable implementation - it includes a test capturing the expected behavior, and it passes other, existing checks. I made some comments inline with the changes that raise concerns for me.

My biggest concern, though, is I feel like the change is backward-incompatible in subtle and hard-to-document ways (and obviously in ways that the test suite hasn't captured). Maybe my sense is wrong. How can we have confidence that this change won't affect users during an upgrade, or how can someone upgrading account for the new behavior?

jaraco:
Please feel free to revive this PR if you have time to consider some of my concerns.

cherrypy/_cpreqbody.py
```diff
@@ -210,7 +210,7 @@ def process_multipart(entity):
 
     # Read all parts
     while True:
-        part = entity.part_class.from_fp(entity.fp, ib)
+        part = entity.part_class.from_fp(entity.fp, ib, entity.default_charset)
```
pawciobiel, line 213: Would it work if I send iso02 encoded file name?

pawciobiel:
IMO if we don't know what encoding is used in POST we shouldn't guess but rather simply pass it on as string/bytes and lave the decoding job to handler.

PR #1435, opened 2016-05-27
Fix signal handler registration.

thezbyg:
Using configuration options like these:

```
'engine.SIGHUP': None
'engine.SIGTERM': None
```

causes exceptions to be thrown when handling SIGTERM or SIGHUP signals
because listeners are registered without entries in the `_priorities`
dictionary.

In `_cpconfig.py`, the following code registers a SIGTERM handler:

```
engine.listeners['SIGTERM'] = set([v])
```

In `process/wspbus.py`, the publish method gets called when signal is
received and executes the following code:

```
items = [(self._priorities[(channel, listener)], listener)
     for listener in self.listeners[channel]]
```

`self._priorities` dictionary does not have the necessary key and KeyError
is thrown.

### DIFF
--- a/cherrypy/_cpconfig.py
+++ b/cherrypy/_cpconfig.py
@@ -282,9 +282,11 @@ def _engine_namespace_handler(k, v):
     elif k == 'deadlock_poll_freq':
         engine.timeout_monitor.frequency = v
     elif k == 'SIGHUP':
-        engine.listeners['SIGHUP'] = set([v])
+        if v is not None:
+            engine.subscribe('SIGHUP', v)
     elif k == 'SIGTERM':
-        engine.listeners['SIGTERM'] = set([v])
+        if v is not None:
+            engine.subscribe('SIGTERM', v)
     elif "." in k:
         plugin, attrname = k.split(".", 1)
         plugin = getattr(engine, plugin)


jaraco:
Your changes appear in a method that appears to be [for backward compatibility](https://github.com/cherrypy/cherrypy/blob/499ae61/cherrypy/_cpconfig.py#L255), and since None wasn't supported for these listeners before, I don't think it needs to be supported now.

I'm not sure what functionality supersedes this compatibility layer, but could you first explore to see if that functionality would suit your needs? Which by the way, what is it you're seeking to accomplish by passing None for these values?

thezbyg:
I am using `embedded` environment which sets signal handlers to None, and then enabling default signal handlers by calling:

```
if hasattr(cherrypy.engine, 'signal_handler'):
    cherrypy.engine.signal_handler.subscribe()
```

As the SIGTERM None handler gets added to the handler list, the KeyError is thrown when SIGTERM signal is received.

I see that `production` environment is equal to `embedded` environment except for the signal handlers, is `embedded` environment deprecated?

jaraco:
> I see that production environment is equal to embedded environment except for the signal handlers, is embedded environment deprecated?

I believe embedded is meant for an app which has an embedded web server, but isn't entirely a web server, so you don't want the web server portion of the application to be responding to those signals (but rather let the host application handle them as appropriate). It's not deprecated as far as I know; it just seems to be using a deprecated technique for disabling the signal handlers.

jaraco:
> is embedded environment deprecated?

I'm beginning to think maybe it is. Why not just start up the server and not configure any signal handlers?

jaraco:
I'm seeing now why you not only check for `None` in your patch, but you also call subscribe rather than setting the listeners - the KeyError you encounter is not triggered because None is passed, but because setting listeners explicitly bypasses the setting of priorities, which means that this code segment would cause crashes under any circumstances (prior to your patch).

jaraco:
On further consideration, I think I was wrong about the 'engine' namespace handler being only for compatibility. There are some config values that are there for compatibility, but I don't think the signal handlers are part of that backward compatibility.

Still, what I don't understand is why is specifying None preferable to simply not setting anything for the signal handlers?

Given that this patch conflates two issues - supporting None for signal handlers and ensuring that priorities are properly configured for a given channel when configured through a config file, I'm going to decline it. I'll file a ticket for the second issue. For the first, you're welcome to file a ticket or another pull request, but I do believe it deserves an explanation as to why it's necessary versus simply omitting that config.

Issue #1352, opened 2015-01-10
Text returned as byte object when using 'enctype="multipart/form-data"' in forms


bb-migration:
Originally reported by: **Juern Brodersen (Bitbucket: [juern_uni](http://bitbucket.org/juern_uni), GitHub: Unknown)**

----------------------------------------

If you are using 'enctype="multipart/form-data"' in forms and putting more than around 1000 characters in a text field that text field will be returned as a byte object.
Using less characters in the text field will return a string.

```
#!python

import cherrypy

class StringGenerator(object):
    @cherrypy.expose
    def index(self):
        return """<html>
          <head></head>
          <body>
            <form method="post" action="show" enctype="multipart/form-data">
              <textarea type="text" name="textarea" cols="50" rows="10"></textarea>
              <button type="submit">Give it now!</button>
            </form>
          </body>
        </html>"""

    @cherrypy.expose
    def show(self, **kw):
        a = str(type(kw['textarea']))
        return a[1:-1]

if __name__ == '__main__':
    cherrypy.quickstart(StringGenerator())
```
Less than 1000 characters will return "class 'str'" more returns "class 'bytes'"

Tested with:

python 3.4

cherrypy 3.6

----------------------------------------
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1352

jaraco:
Fix released in 5.5.0.

PR #1438, opened 2016-06-03
Patch for #1352

JB26:
Fix for #1352
I moved the part responsible for encoding a field value. This way the value gets encoded no matter if it was cached in a file or not.
A test case is included. Is setting `cherrypy._cpreqbody.Part.maxrambytes` inside a test case ok?

### DIFF
--- a/cherrypy/_cpreqbody.py
+++ b/cherrypy/_cpreqbody.py
@@ -520,7 +520,25 @@ class Entity(object):
             self.file.seek(0)
         else:
             value = self.value
+        value = self.decode_entity(value)
         return value
+
+    def decode_entity(self , value):
+        """Return a given byte encoded value as a string"""
+        for charset in self.attempt_charsets:
+            try:
+                value = value.decode(charset)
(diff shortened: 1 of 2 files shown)


jaraco:
This change strikes me as a smart refactor. Thanks for it.

> Is setting cherrypy._cpreqbody.Part.maxrambytes inside a test case ok?

Yes, I think so, especially since you unset it (assuming an exception doesn't occur). I'll provide a follow-up commit that demonstrates how to more robustly patch a value like this using mock (I would use pytest's monkeypatch, which is even more elegant, but CherryPy relies on Nose).

jaraco:
Also, if I may make one other suggestion - with your commit messages, if you could add a little more detail to the messages describing briefly the purpose of the change (what was changed, why was it changed, how was it changed), that would be helpful to others including your future self when reviewing those commits. Obviously, the link to the relevant issue is the most important piece, as it links to the deep detail where all of that information could be derived.

Issue #1146, opened 2012-05-16
Problems handling uploaded files which contain commas and spaces in filename


ghost:
Originally reported by: **Allan Crooks (Bitbucket: [the_allanc](http://bitbucket.org/the_allanc), GitHub: Unknown)**

---

If you try uploading a file to CherryPy which contains a comma followed by space, it can have problems handling the uploaded content because of the filename.

Using the example found here:
http://docs.cherrypy.org/stable/progguide/files/uploading.html

1) Uploading a simple file called "boop.csv", I get this:

```
myFile length: 7007
myFile filename: boop.csv
myFile mime-type: text/csv 
```

2) Uploading a file called "foo, bar.csv", I get this:

```
myFile length: 325
myFile filename: "foo
myFile mime-type: text/csv
```

3) Uploading a file called "bar, xxxx.csv", I get this:

```
404 Not Found

Missing parameters: myFile

Traceback (most recent call last):
<snip>
HTTPError: (404, 'Missing parameters: myFile')
```

The behaviour does seem somewhat arbitrary, but some filenames seem to result in a truncated filename being submitted (ilke in the second case) or a broken request altogether.

(Recreated with Python 2.6.6 on a 64 bit install of Linux using CherryPy 3.2.2.)

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1146

Issue #1397, opened 2015-12-17
Not respecting quotes in headers


ghost:
Originally reported by: **Shy Pike (Bitbucket: [shypike](http://bitbucket.org/shypike), GitHub: @shypike**

---

When a header contains a value between quotes and that value contains a semicolon 
";", the header will not be parsed properly.

Example from a "Content-Disposition" header:

```
#!text

form-data; name="name"; filename="one;word.txt"
```

The cause is cherrypy/lib/httputil.py, line 148, function parse().

```
#!python

atoms = [x.strip() for x in elementstr.split(";") if x.strip()]
```

That code is too simple, it splits in the middle of the file name.

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1397

Safihre:
Circumvented that in SABnzbd in a little clumsy way:
https://github.com/sabnzbd/sabnzbd/commit/69ce6e33118c72492f4d0d6c42f25533b5d446ad

JB26:
In case anyone else is wondering how double quotes in filenames (e.g. `file"name.csv`) are handled here are my findings.
Firefox escapes them like this `\"` which works fine with cherrypy.
Chrome escapes them using `%22` which makes it impossible to differentiate between `file"name.csv` and `file%22name.csv`.
A discussion about this on w3: [multipart/form-data filename encoding: unicode and special characters](https://lists.w3.org/Archives/Public/public-whatwg-archive/2012May/0003.html)

PR #1439, opened 2016-06-05
Issue #1397

JB26:
If you don't like importing `cgi` for this. The regex used to fix issue #1146 will probably work here, too.

I also kinda hijacked the test for issue #1146, hope that's ok :)


jaraco:
Looks good to me. I suspect the cgi module didn't have this functionality or wasn't as robust as the CP implementation at the time this code was written. Thanks for addressing it.

PR #1442, opened 2016-06-05
Convert request params based on function annotations.  Fixes #1441.

### LINKED ISSUE #1441: Convert request params based on function annotations.
With the increasing focus on [type hints](https://docs.python.org/3.5/library/typing.html) in Python, having a tool which automatically converted query params (with exception handling) would be a nice feature.

It's common for apps to have boilerplate code for [converting request params](http://stackoverflow.com/questions/32774024/), and returning 400-levels errors on failure.

```python
@cherrypy.tools.params(exception=ValueError)
def resource(self, limit: int):
    assert isinstance(limit, int) # a ValueError would have already raised a 400
```

coady:


jaraco:
This idea is a very cool one. And the implementation is so short and concise, signs of an elegant implementation. I have a couple of comments to make about the specific implementation, but I don't see why this change couldn't be rolled into an upcoming release in short order.

CHANGES.txt
```diff
@@ -3,6 +3,7 @@
 
 * Issue #1444: Correct typos in ``@cherrypy.expose``
   decorators.
+* #1441: Added tool to automatically convert request params.
```
jaraco: Thanks for the changelog entry. Can you move it into a new section with an 0.1 bump?

cherrypy/test/test_params.py
```diff
@@ -0,0 +1,34 @@
...
+
+class ParamsTest(helper.CPWebCase):
+    @staticmethod
+    def setup_server():
+        class Root:
+            @cherrypy.expose
+            @cherrypy.tools.params(exception=ValueError)
+            def resource(self, limit=None, sort=None):
+                return '{0}: {1}'.format(limit, type(limit).__name__)
+            # for testing on Py 2
```
jaraco, line 14: I like that this can be tested on Python 2, but I'd like also if on the appropriate Python versions if it used the native syntax (to validate that the native syntax does what you expect, but also to make the syntax visible). I'm not sure it's possible or won't be so ugly to implement that it violates the goals, so this is just something I'd like to see, but don't require.

cherrypy/lib/cptools.py
```diff
@@ -630,3 +630,21 @@ def set_response_header():
         v.sort()
         resp_h['Vary'] = ', '.join(v)
     request.hooks.attach('before_finalize', set_response_header, 95)
+
+
+def convert_params(exception, error=400):
+    """Convert request params based on function annotations, with error handling.
+
+    exception
+        Exception class to catch.
```
jaraco, line 639: When would the exception class ever not be a TypeError? If there aren't such cases, perhaps this detail could be omitted. If there are such cases, perhaps it would be suitable to supply `TypeError` as the default for `exception` as it's likely the most common usage.

cherrypy/test/test_params.py
```diff
@@ -0,0 +1,34 @@
...
+        self.assertStatus(200)
+        self.assertBody('None: NoneType')
+
+        self.getPage('/resource?limit=0')
+        self.assertStatus(200)
+        self.assertBody('0: int')
+
+    def test_error(self):
+        self.getPage('/resource?limit=')
+        self.assertStatus(400)
```
jaraco, line 30: Probably worth asserting the body here too.

cherrypy/test/test_params.py
coady, line 14: It's a `SyntaxError`, so it will have to be an `exec` statement like in `test_objectmapping`.

cherrypy/lib/cptools.py
coady, line 639: Normally I'm wary of implicit error catching, but since this is really error translating, sounds ok to me.

It's `ValueError` though; that would cover the common cases: `int`, `float`, `json.loads`, etc.

Issue #1444, opened 2016-06-06
Import of cherrypy.lib.cpstats fails


cachedout:
Using CherryPy 4.0.0 downloaded from PyPi:

```
>>> from cherrypy.lib import cpstats
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/cherrypy/lib/cpstats.py", line 438, in <module>
    class StatsPage(object):
  File "/usr/lib/python2.7/site-packages/cherrypy/lib/cpstats.py", line 478, in StatsPage
    @cherrypy.exposed
AttributeError: 'module' object has no attribute 'exposed'
>>> 
```

PR #1445, opened 2016-06-06
fixup 5e1a48f: exposed -> expose

jfindlay:
Proposed fix for #1444.  Updates 5e1a48f.  You are welcome to reject and reimplement this if it is not quite correct.  I only tracked down the problem as far as this.


jaraco:
Very nice. Thanks for putting this together. I had missed the docs updates in my own edit.

PR #1450, opened 2016-07-01
Try to do the right thing when autoreloading a process invoked with 'python -m'

### LINKED ISSUE #1411: Autoreload can't handle module / __main__

----------------------------------------

I try to run `python -m mymodule` (and `mymodule/` contains `__main__.py`). The module then spawns a CherryPy server. Everything works until it tries to reload itself, where it tries to run `/path/to/python mymodule/__main__.py`, without importing `mymodule`. This makes my module to throw various `ImportError`s that claim `mymodule` cannot be found.

Could you share any tips how to organize the code, or change the way I run the server, to make reloading work?

----------------------------------------
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1411

batterseapower:
Appears to fix #1411

### DIFF
--- a/cherrypy/process/wspbus.py
+++ b/cherrypy/process/wspbus.py
@@ -379,6 +379,20 @@ class Bus(object):
         args = sys.argv[:]
         self.log('Re-spawning %s' % ' '.join(args))
 
+        # If sys.path[0] is an empty string, we were (probably) invoked with -m and
+        # the effective path is about to change on re-exec.  Add the current directory
+        # to $PYTHONPATH to ensure that the new process sees the same path we did.
+        #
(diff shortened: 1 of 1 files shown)


jaraco:
Excellent - thanks!

saioia:
Thanks a lot! This will really help my workflow.

sevmonster:
While it does recognize the right files to reload now, I still think local imports are messing up -- I will investigate later this week.

sevmonster:
Forgot to post back, I messed with it but was unable to find the source of the issue... Local imports kinda work? Sometimes they load correctly and sometimes they don't. It's a little out of my expertise, import system has always confounded me.

PR #1458, opened 2016-07-21
Systemd socket activation

xaiki:
This tiny patch implements systemd's socket activation mechanism[0] for
cherrypy servers, it's based on work sponsored by Endless Computers[1] to
package kalite[2] as a flatpak[3].

Socket Activation allows to setup a system so that systemd will sit on a
port and start services 'on demand' (a little bit like inetd and xinetd
used to do). This patch makes that mechanism work for any cherrypy
service without modification.

[0] http://0pointer.de/blog/projects/socket-activation.html
[1] https://endlessm.com
[2] https://learningequality.org/ka-lite/
[3] https://flatpak.org

https://phabricator.endlessm.com/T4489

Signed-off-by: Niv Sardi xaiki@evilgiggle.com

### DIFF
--- a/cherrypy/wsgiserver/__init__.py
+++ b/cherrypy/wsgiserver/__init__.py
@@ -1925,7 +1925,10 @@ class HTTPServer(object):
         interface" (INADDR_ANY), and '::' is the similar IN6ADDR_ANY for
         IPv6. The empty string or None are not allowed.
 
-        For UNIX sockets, supply the filename as a string.""")
+        For UNIX sockets, supply the filename as a string.
+
+        Systemd socket activation is automatic and doesn't require tempering
+        with this variable""")
 
     def start(self):
         """Run the server forever."""
--- a/cherrypy/wsgiserver/__init__.py
+++ b/cherrypy/wsgiserver/__init__.py
@@ -1939,7 +1939,11 @@ class HTTPServer(object):
             self.software = "%s Server" % self.version
 
         # Select the appropriate socket
-        if isinstance(self.bind_addr, six.string_types):
+        self.socket = None
+        if os.getenv('LISTEN_PID', None):
+            # systemd socket activation
+            self.socket = socket.fromfd(3, socket.AF_INET, socket.SOCK_STREAM)
+        elif isinstance(self.bind_addr, six.string_types):
             # AF_UNIX socket
 
             # So we can reuse the socket...
@@ -1973,21 +1977,21 @@ class HTTPServer(object):
                     info = [(socket.AF_INET, socket.SOCK_STREAM,
                              0, "", self.bind_addr)]
 
-        self.socket = None
-        msg = "No socket could be created"
-        for res in info:
-            af, socktype, proto, canonname, sa = res
-            try:
-                self.bind(af, socktype, proto)
-            except socket.error as serr:
-                msg = "%s -- (%s: %s)" % (msg, sa, serr)
-                if self.socket:
-                    self.socket.close()
-                self.socket = None
-                continue
-            break
         if not self.socket:
-            raise socket.error(msg)
+            msg = "No socket could be created"
+            for res in info:
+                af, socktype, proto, canonname, sa = res
+                try:
+                    self.bind(af, socktype, proto)
+                    break
+                except socket.error as serr:
+                    msg = "%s -- (%s: %s)" % (msg, sa, serr)
+                    if self.socket:
+                        self.socket.close()
+                    self.socket = None
+
+            if not self.socket:
+                raise socket.error(msg)
 
         # Timeout so KeyboardInterrupt can be caught on Win32
         self.socket.settimeout(1)
--- a/cherrypy/process/servers.py
+++ b/cherrypy/process/servers.py
@@ -113,6 +113,7 @@ Please see `Lighttpd FastCGI Docs
 an explanation of the possible configuration options.
 """
 
+import os
 import sys
 import time
 import warnings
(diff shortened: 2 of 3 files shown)


cherrypy/wsgiserver/__init__.py
```diff
@@ -1939,7 +1939,11 @@ def start(self):
             self.software = "%s Server" % self.version
 
         # Select the appropriate socket
-        if isinstance(self.bind_addr, six.string_types):
+        self.socket = None
+        if os.environ.get('LISTEN_PID', None):
```
webknjaz: Why set default value to `None`? It's not necessary.
Moreover, [`os.getenv('LISTEN_PID')`](https://docs.python.org/2/library/os.html#os.getenv) would be more verbose.

cherrypy/wsgiserver/__init__.py
```diff
@@ -1973,21 +1977,21 @@ def start(self):
...
+                af, socktype, proto, canonname, sa = res
+                try:
+                    self.bind(af, socktype, proto)
+                except socket.error, serr:
+                    msg = "%s -- (%s: %s)" % (msg, sa, serr)
+                    if self.socket:
+                        self.socket.close()
+                    self.socket = None
+                    continue
+                break
```
webknjaz: I'd move `break` to line 1986, just next to self.bind() and remove `continue`.

cherrypy/wsgiserver/__init__.py
```diff
@@ -1973,21 +1977,21 @@ def start(self):
...
-                continue
-            break
         if not self.socket:
-            raise socket.error(msg)
+            msg = "No socket could be created"
+            for res in info:
+                af, socktype, proto, canonname, sa = res
+                try:
+                    self.bind(af, socktype, proto)
+                except socket.error, serr:
```
webknjaz: This outdated exception handling syntax breaks TravisCI tests. Plz fix

cherrypy/wsgiserver/__init__.py
xaiki: fixed

cherrypy/wsgiserver/__init__.py
xaiki: fixed

cherrypy/wsgiserver/__init__.py
xaiki: fixed

xaiki:
@webknjaz thanks for the review, i fixed your comments and pushed !

webknjaz:
LGTM, but there's some issue w/ [TravisCI](https://travis-ci.org/cherrypy/cherrypy/builds/146641684). Could you plz try restarting builds?
I think we can now ping @jaraco and see if he would merge this.

jaraco:
I tried restarting the builds, but the result is the same. Why would the tests not run on Travis with this patch?

webknjaz:
@jaraco I think, there's some issue with TravisCI itself. Look at the `pip install` output. It differs for different jobs. Did you mention this?
I have a strong feeling that it gets cut/halted at some point.

webknjaz:
@jaraco I've found the same issue, which happened in the past: https://travis-ci.org/cherrypy/cherrypy/jobs/139925725

webknjaz:
Okay.. I've tried running `nosetests -s -v` myself and I also get return code 70. So the issue is not in Travis

webknjaz:
``` sh
$ .tox/py27/bin/nosetests -s -v
nose.config: INFO: Set working dir to ~/src/cherrypy-xaiki/cherrypy
nose.config: INFO: Working directory ~/src/cherrypy-xaiki/cherrypy is a package; adding to sys.path
nose.config: INFO: Ignoring files matching ['^\\.', '^_', '^setup\\.py$']
nose.selector: INFO: ~/src/cherrypy-xaiki/cherrypy/cherryd is executable; skipped
nose.selector: INFO: ~/src/cherrypy-xaiki/cherrypy/daemon.py is executable; skipped
nose.selector: INFO: ~/src/cherrypy-xaiki/cherrypy/test/sessiondemo.py is executable; skipped
nose.selector: INFO: ~/src/cherrypy-xaiki/cherrypy/test/test_session.py is executable; skipped
$ echo $?
70
```

webknjaz:
It looks there's something malformed in this patch. Even if I generate diff and apply it to working master it breaks nose command.

xaiki:
@webknjaz it was the removal of `self.socket = None` i added it back and it now unbreaks nose, i have no idea why ....

webknjaz:
@xaiki okay, seems good

Could you plz include some documentation/examples of running cherrypy app with use of this feature?

Also it would be nice to have some tests covering it.

xaiki:
a test is a bit hard as you need to have systemd setup, 

we (endlessm.com) use that patch to enable kalite systemd activation, here's our system part:
https://github.com/mariospr/eos-kalite-system-helper

basically, you need to setup these files:
https://github.com/mariospr/eos-kalite-system-helper/blob/master/debian/eos-kalite-system-helper.service
https://github.com/mariospr/eos-kalite-system-helper/blob/master/debian/eos-kalite-system-helper.socket

to make it work

webknjaz:
@xaiki Could you plz describe this in [docs/deploy.rst](/cherrypy/cherrypy/blob/master/docs/deploy.rst)?
Ability of others to use this feature seamlessly will make this PR more valuable.
Also add some information about the feature to the docstring of `start()` method.

xaiki:
docs added to deploy.rst
for start you mean start of the wcgiserver ? or the dameon start ? i'm not that familiar with those uses =)

xaiki:
added to the bind_addr docstring as it seemed more relevant.

webknjaz:
LGTM

@jaraco I thinks this PR is ready for merge

xaiki:
good, i had missed a patch from @mariospr just got it in the PR. thanks for a great review @webknjaz

jaraco:
Excellent. Thanks everyone for putting this together.

jaraco:
Pending release in 7.1.0.

mariospr:
@xaiki, @jaraco: that was fast! Thanks for accepting my patch :-)

PR #1462, opened 2016-07-25
Fix running PR#1461 test under python 3.2

webknjaz:
// hotfix for 3.2

/cc: @jaraco


jaraco:
I wanted to keep the code as readable as possible, so I did 74b36988470 instead.

jaraco:
Thanks, though, for being on top of the test failures and developing a patch.

PR #1464, opened 2016-07-25
Add testing against loosely supported Python versions in Travis

webknjaz:
@jaraco I've mentioned classifiers declaring support of Python 3.1in `setup.py`, which isn't tested at all.
Also, there are several extra unsupported or unstable environments, which are useful for testing against.

As an experiment, I've added testing against `3.1`, `nightly`, `pypy` and `pypy3` targets. I've marked them as allowed to fail, so it won't affect overall build status. Let's try it out if you don't mind. This will help cherrypy detect possible issues with future versions of Python.

Plz let me know your opinion on this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,11 @@ python:
 - nightly
 matrix:
   allow_failures:
-    - 3.1
-    - pypy
-    - pypy3
-    - nightly
+    python:
+      - 3.1
+      - pypy
+      - pypy3
+      - nightly
 env:
 - WEBTEST_INTERACTIVE=false
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ python:
 - nightly
 matrix:
   allow_failures:
-    python:
+    - python:
       - 3.1
       - pypy
       - pypy3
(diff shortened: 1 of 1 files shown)


jaraco:
Last I checked, Python 3.1 and nightly weren't possible targets on Travis. Glad to hear they are now. Definitely worth trying to add that assurance. I'm pretty sure pypy3 will fail horribly in lots of ways. IIIRC, pypy mostly works, but fails on a couple of edge cases. Glad though to include the tests even if just to establish a failing baseline, especially since they can be allowed to fail but not break the build.

jaraco:
I'll wait until the Travis tests light up green, but I'm +1 on the idea.

webknjaz:
@jaraco 

Well, it looks they don't really support 3.1 (I've double checked the docs and now see this in build log). Of course it could be worked around with tox, but I'm not sure whether it's worth it. Btw, it seems everyone is dropping 3.1 and 3.2 support. Django has even dropped 3.3.
Some tests really fail under pypy3, pypy target stucks.
And the whole test suite passes under nightly build (3.6.0a3+ currently).
I've also found 3.5-dev target in docs. So I think of dropping 3.1 from config and adding 3.5-dev with allowed_failures

jaraco:
LGTM

PR #1476, opened 2016-08-21
Store bytes in memcached session, but always return native strings

### LINKED ISSUE #1453: class MemcachedSession returns session ids as bytes, and browsers can not handle them
This commit 

https://github.com/cherrypy/cherrypy/commit/0952bcd8b44dc4f9cfc50e7e179b4bd4e625ad56

reimplemented the _set_id method to encode the session ids (as the comment says, to allow memcached to index the key).

However:
1. Version 1.58 of python-memcached does not require bytes anymore. It accepts bytes or unicode strings.
2. I realized cherrypy is returning the session id as bytes in the response headers. Of course, browsers don't know what to do with that, so they modify them wrongly, sending back a different "bytes string" in the next request headers. This is provoking cherrypy to create a new session for each request.

**Solutions**
- On python 3.4, with cherrypy 6.0.2 and python-memcached 1.58, removing the lines

```
if isinstance(value, unicodestr):
     value = value.encode('utf-8')
```

solves the problem.
- Maybe it would be more convenient to modify the _get_id_ method instead, decoding the id first if it is instance of bytes, so this way it would be retrocompatible with old versions of python-memcached.

**Edit:** 
https://github.com/cherrypy/cherrypy/issues/1447 is a temporary fix for this problem, although i'm not getting any exception...

webknjaz:
Fix #1453


jaraco:
`ntob` and `tonative` are deprecated constructs, so probably shouldn't be used here. An the very least, I think the change to _set_id is no better than the implementation that's in place now. Why ensure it's unicode just to convert it to bytes?

I can see the value in the change to `_get_id` as it provides backward compatibility for older versions of python-memcached on Python 2.

Personally, I'd be inclined to drop support for older versions of python-memcachd, and remove all of this workaround code altogether. What do you think?

webknjaz:
Well, I think to do so we need to add dependency restriction.
I suggest adding `extras` to `setup.py`.
It would be useful for end-users as well. If they want to use memcached backed sessions they'll just specify (e.g.) `cherrypy[memcached-sessions]` and won't need to track `memcached` client version themselves.

webknjaz:
@jaraco I think now it's OK to merge it

jaraco:
Looks good to me. Can you add a CHANGES.rst entry as well?

webknjaz:
Sure, done. Merging PR :)

Issue #1473, opened 2016-08-21
Exception handling utility.


coady:
In most cherrypy apps, I end up with a utility like this:

``` python
def HTTPError(status, *exceptions):
    "Interpret exceptions as an HTTPError with given status code."
    try:
        yield
    except exceptions as exc:
        raise cherrypy.HTTPError(status, str(exc))
```

Trivial, but it's handy in a framework like CherryPy that allows raising a response anywhere.  Compare that to:
- Django's `get_object_or_404`.  Convenient, but not extensible and coupled to its ORM.
- Flask's application-level error handlers.  The problem with global handlers is they tend towards a proliferation of custom single-use exception classes.

Whereas something like:

``` python
with cherrypy.NotFound(IndexError):
    # db query [0]
```

allows error handling at any level of granularity.

webknjaz:
Excuse me, I didn't quite understand what you are asking for in this issue. Could you please make your point more clear?

coady:
It's a feature request for more elegant exception handling, via context managers.  Opened PR  #1478 for discussion.  Typical boilerplate like:

``` python
try:
    obj = db_query.get()
except db_excs.NotFound:
    raise cherrypy.HTTPError(404)
```

would become:

``` python
with cherrypy.HTTPError.handle(db_excs.NotFound, 404):
    obj = db_query.get()
```

Open question:  should `HTTPError` _have_ a context manager, _be_ a context manager, or just a separate utility?

webknjaz:
@jaraco how do you like @coady's idea? It looks interesting to me

jaraco:
Overall I like it, although the implementation is key. Here are some other possibilities to consider:
- additionally allow rewriting exceptions in a decorator so it doesn't indent the main body of the code, something like

``` python
@cherrypy.HTTPError.trap(db_excs.NotFound, 404)
def items(self):
    obj = db.query.get()
    return render(obj)
```
- Allow for a non-error response as well, like

``` python
with cherrypy.HTTPResponse.handle(db_excs.NotFound, "<span>Didn't find it</span>"):
    obj = db_query.get()
```

But most important is going to be the updates to the docs that explain how this is used. If the docs can present a coherent description of the functionality, then I see no problem with providing an implementation.

pawciobiel:
Knowing that "a context manager is an object that defines the runtime context to be established when executing a with statement. " the above code can be confusing.

webknjaz:
[...]

PR #1478, opened 2016-08-21
HTTPError provides context manager for handling exceptions.

coady:
Implementation for #1473.
- [x] tests
- [x] implementation
- [x] docs

### DIFF
--- a/cherrypy/_cperror.py
+++ b/cherrypy/_cperror.py
@@ -115,6 +115,7 @@ Note that you have to explicitly set
 and not simply return an error message as a result.
 """
 
+import contextlib
 from cgi import escape as _escape
 from sys import exc_info as _exc_info
 from traceback import format_exception as _format_exception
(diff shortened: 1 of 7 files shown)


webknjaz:
Could you please upgrade this PR in accordance to [@jaraco's suggestions](/cherrypy/cherrypy/issues/1473#issuecomment-241396807)?

webknjaz:
@coady ping

coady:
Added docs, and examples in `cherrypy.lib`, just to demonstrate how common the pattern is.

webknjaz:
And what about `trap` and other stuff?

jaraco:
@webknjaz What would you like to see for this to be acceptable?

webknjaz:
I'd like to see your suggestions with trap decorator and non-error response handler: https://github.com/cherrypy/cherrypy/issues/1473#issuecomment-241396807. But they could be implemented in separate PR.

jaraco:
That sounds like a good plan to me.

PR #1479, opened 2016-08-22
Added Content-Length handling in webtest

conradoplg:
The webtest client does not add the Content-Length header, unless you don't provide Content-Type in which case it adds `Content-Type: application/x-www-form-urlencoded` and the Content-Length. This makes cumbersome to test JSON requests, since then you need another Content-Type and have to add Content-Length manually.

This PR changes the logic in order to add Content-Length if missing.

However, this prevents testing requests with no headers (i.e. the changed test in test_request_obj). I'm not sure if this patch is acceptable but since I had already written it after I realized this issue, here it is.

### DIFF
--- a/cherrypy/test/test_request_obj.py
+++ b/cherrypy/test/test_request_obj.py
@@ -761,7 +761,7 @@ class RequestObjectTests(helper.CPWebCase):
         # Provide a C-T or webtest will provide one (and a C-L) for us.
         h = [("Content-Type", "text/plain")]
         self.getPage("/method/reachable", headers=h, method="PUT")
-        self.assertStatus(411)
+        self.assertStatus(200)
 
         # Request a custom method with a request body
         b = ('<?xml version="1.0" encoding="utf-8" ?>\n\n'
--- a/cherrypy/test/test_json.py
+++ b/cherrypy/test/test_json.py
@@ -78,6 +78,11 @@ class JsonTest(helper.CPWebCase):
         self.assertBody('ok')
 
         body = '[13, "c"]'
+        headers = [('Content-Type', 'application/json')]
+        self.getPage("/json_post", method="POST", headers=headers, body=body)
+        self.assertBody('ok')
+
+        body = '[13, "c"]'
         headers = [('Content-Type', 'text/plain'),
                    ('Content-Length', str(len(body)))]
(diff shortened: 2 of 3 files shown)


cherrypy/test/test_request_obj.py
```diff
@@ -761,7 +761,7 @@ def test_basic_HTTPMethods(self):
         # Provide a C-T or webtest will provide one (and a C-L) for us.
         h = [("Content-Type", "text/plain")]
         self.getPage("/method/reachable", headers=h, method="PUT")
-        self.assertStatus(411)
```
jaraco, line 764: This change isn't appropriate. The test is explicitly testing the server to show that it returns a 411 Content Length Required when no content length is supplied.

jaraco:
After reviewing the code, I'm thinking that cleanHeaders is too overreaching anyway. I'd rather see the removal of the adding of headers (even if they're required). They should be extracted out into a separate helper that should be called as necessary. We have an application where we do a lot of JSON tests, and in that application, we subclass WebTest and add these methods:

```
    def postJSON(self, url, entity, headers=None):
        jsonbody = simplejson.dumps(entity, cls=shoji.JSONEncoder,
                                    use_decimal=True)
        h = [("Content-Type", "application/json"),
             ("Content-Length", str(len(jsonbody)))]
        for k, v in (headers or []):
            h.append((k, v))
        self.getPage(url, method="POST", headers=h, body=jsonbody)

    def putJSON(self, url, entity, headers=None):
        jsonbody = simplejson.dumps(entity, cls=shoji.JSONEncoder,
                                    use_decimal=True)
        h = [("Content-Type", "application/json"),
             ("Content-Length", str(len(jsonbody)))]
        for k, v in (headers or []):
            h.append((k, v))
        self.getPage(url, method="PUT", headers=h, body=jsonbody)
```

Perhaps that would serve your needs. The question is, should CherryPy supply something like this? I'm not sure.

For now, I'm closing this PR, as it breaks existing test expectations and almost certainly does the same for other tests in the wild using WebTest.

conradoplg:
Yes, I agree that you shouldn't break things. I ended up using a helper function like you mentioned.

Still, I think the current approach can be a little confusing, specially for someone using CherryPy for the first time. But I don't know the best approach for this. Maybe just adding this issue to the documentation? (I'll try to patch it)

Issue #1487, opened 2016-08-29
Allow use own custom session storage class


michal-juranyi:
Currently, there is quite limited range of supported session storage backends. Even for supported backends there may be need for advanced features of particular backend (i.e. custom Pickler/Unpickler in Memcached). I propose adding a config option for specifying fully qualified class name of custom class to be used instead of predefined *Session class.

webknjaz:
Implemented via #1495

PR #1488, opened 2016-08-29
Allow use custom session storage class (Fixes cherrypy/cherrypy#1487)

michal-juranyi:


cherrypy/lib/sessions.py
```diff
@@ -94,6 +94,7 @@
 import os
 import time
 import threading
+import importlib
```
webknjaz, line 97: We need some workaround for Python 2.6

cherrypy/lib/sessions.py
```diff
@@ -715,8 +716,8 @@ def close():
 close.priority = 90
 
 
-def init(storage_type='ram', path=None, path_header=None, name='session_id',
-         timeout=60, domain=None, secure=False, clean_freq=5,
+def init(storage_type='ram', storage_class=None, path=None, path_header=None,
```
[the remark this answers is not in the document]
webknjaz: I'd add it in the end of the var list

webknjaz:
Okay, this PR needs:
0. to be documented
1. not to break existing tests (see Travis CI)
1.1. not break existing public API
2. introduce new tests covering the feature and proving it works
3. be added to the changelog

webknjaz:
Also, it should conform current cherrypy principles. I'd like to see this as extension to existing session configuration.

cherrypy/lib/sessions.py
jaraco: Just add it as a marked dependency, e.g. `extras_require={':python_version=="2.6"': ['importlib']}`

jaraco:
I'm slightly opposed to this approach. It adds a new, different way to initialize sessions, adding new parameters to the existing init function. So now it's possible to configure sessions as [jaraco.mongodb does](https://github.com/jaraco/jaraco.mongodb/blob/master/jaraco/mongodb/sessions.py), installing itself into CherryPy using the existing API, or to follow a different model and pass the class path.

Your approach has some advantages. In particular, it doesn't require each Sessions subclass to implement an install routine. It also doesn't rely on the mutation of another package's globals. If this technique is adopted, I'd like to see deprecate the `storage_type` parameter, perhaps setting its default to `None` and instead set the default parameter for `storage_class` to `'cherrypy.lib.sessions.RamSessions'`. That way, we set up the library to transition to this new API rather than having two mechanisms.

I think the primary motivation here is not to require that all Session implementations need to be present in the `globals()`.

Other approaches I'd like to consider (in my order of preference):
- Rely on entry points to advertise and resolve CherryPy sessions and still allow them to be indicated by `storage_type`.
- Solicit the storage class directly (and not a string indicating it) (e.g. `storage_class=RamSession`).
- Use `Sessions.__subclasses__` to locate session implementations and still resolve them by `storage_type`.

jaraco:
> - Solicit the storage class directly (and not a string indicating it) (e.g. storage_class=RamSession)

The more I think about it, the more I prefer this option. Only, I'm not sure a config file could supply this value. If it can, then I think that's the best approach and should be pursued. It should additionally deprecate the `storage_type` parameter.

Issue #1485, opened 2016-08-28
[Python3.4] CherryPy >= 5.5.0 vpath encoding is wrong


ikus060:
Seams to be related to #1248 or #1194, but I can't identify exactly which commit broke my testcases. Basically, starting at 5.5.0, some testcases are not working any more.

Sending HTTP request:
GET /browse/testcases/R%C3%A9pertoire%20Supprim%C3%A9/ is received as str: 'Répertoire Supprimé'. The url is UTF8 valid.

Sending HTTP request:
GET /browse/testcases/R%E9pertoire%20Supprim%E9/ is also received as str: 'Répertoire Supprimé'. While this one is ISO-8859-1 valid.

Cherrypy seams to guess the encoding. I don't know how CherryPy figure out the encoding, but it's not working for my specific scenario since I need the bytes representation of the URL.

i.e.: 
`R%C3%A9pertoire%20Supprim%C3%A9` vs `R%E9pertoire%20Supprim%E9` are two different path for me and both of them are valid.

Any help on this subject ?

ikus060:
After some research, it seams changes for #1431 (commit 4a23073) is creating a side effect.
1. Contrary to poofeg comments, the value of PATH_INFO is unicode decoded using ISO-8859-1.
   - %E9 > é
   - %C3%A9 > Ã©
2. PATH_INFO is sent as unicode and reach _cpwsgi.py AppResponse#run() untouched
3. Then the line `path.encode(old_enc).decode(new_enc)` raise a UnicodeDecodeError or not depending of the URL content.

**Result:**
4. A URL with %E9 will raise an exception, thus PATH_INFO as unicode containing ISO-8859-1 data is sent to application...
5, A URL with %C3%A9 will not raise an exception, this PATH_INFO is properly converted from ISO-8859-1 then converted from UTF-8 to unicode.

**Expected Results:**
1. Some warning should be shown if a UnicodeDecodeError is raised. It would have help me debug this.
OR
2. Decoding the PATH_INFO should use "surrogateescape".

jaraco:
@ikus060: Since you have test cases, would you mind sharing one here or in a PR (preferably integrated with the CherryPy test suite)? It's going to take some work to figure out what's the proper thing to do here. I think surrogateescape might be appropriate if CherryPy has already done the best it can do to decode the content but still cannot.

ikus060:
[...]

PR #1490, opened 2016-08-31
WIP: Use "surrogateescape" to decode invalid URL.

ikus060:
See ticket #1485

### DIFF
--- a/cherrypy/_cpwsgi.py
+++ b/cherrypy/_cpwsgi.py
@@ -337,8 +337,8 @@ class AppResponse(object):
                 # masquerading as unicode. So we have to encode back to
                 # bytes and then decode again using the "correct" encoding.
                 try:
-                    u_path = path.encode(old_enc).decode(new_enc)
-                    u_qs = qs.encode(old_enc).decode(new_enc)
+                    u_path = path.encode(old_enc).decode(new_enc, "surrogateescape")
+                    u_qs = qs.encode(old_enc).decode(new_enc, "surrogateescape")
                 except (UnicodeEncodeError, UnicodeDecodeError):
                     # Just pass them through without transcoding and hope.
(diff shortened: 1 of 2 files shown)


jaraco:
Thanks! The test case looks great. And I think the fix is on the right track, but as you observed, it causes other parts of the test suite to fail.

Additionally, I think we should drop the "Just pass them through and hope". If there's a UnicodeEncodeError, then the application is violating the WSGI spec and an error should be raised. And with the surrogate escape, the UnicodeDecodeError should never happen.

I'm going to accept this PR but move it into another branch so we can iterate on it.

jaraco:
I've moved that commit into the `stricter-url-handling` branch.

ikus060:
Will it someday be part of master branches ? Whats is your plan about "stricter-url-handling" ?

jaraco:
Before it can go into master, the tests have to pass.

I started looking at `test_objectmapping.ObjectMappingTest.testObjectMapping`, but it's a mess. If I use a UTF-8 encoded URL for that test, the test start passing on Python 3 and failing on Python 2. The ISO-8859-1 encoded URL is actually wrong, but it probably works because of the generous ISO-8859-1 decoding in `ntob` and `ntou`.

The encode/decode behavior in `AppResponse.run` only applies to Python 3, which seems wrong. It should probably apply to Python 2 and 3, but if I enable it for Python 2, it breaks a lot of things. This is going to be a tough nut to crack, and will probably require a fairly massive overhaul of the Unicode handling in CherryPy.

It's conceivable we could just disable the failing tests and kick this can down the road, causing disruption in the codebase and in our user's apps. It's also conceivable we could just drop support for Python 2.

I'd rather, however, just iterate on this branch and work toward a solution with proper passing tests.

PR #1493, opened 2016-09-03
Finish fix #1489 (WSGI should emit bytes)

### LINKED ISSUE #1489: WSGI applications are required to return bytes.
Needs @jaraco advice about:
fgallaire/wsgiserver#2

fgallaire:


jaraco:
Looks good to me. I suspect the test failure is a spurious one since tests passed on 2.7.

Issue #1497, opened 2016-09-07
Connecting to HTTPS port using HTTP when using 'builtin' SSL


Safihre:
We (SABnzbd) are investigating if we need pyOpenSSL at all, therefore we tried running with `server.ssl_module: 'builtin'` on CherryPy 8.1.0.
However, when connecting to the HTTPS port via HTTP we get an ugly error that doesn't happen when using pyOpenSSL as `ssl_module`.

```
2016-09-07 16:25:57,895::ERROR::[_cplogging:216] [07/Sep/2016:16:25:57] ENGINE Error in HTTPServer.tick
Traceback (most recent call last):
  File "C:\Python27\Lib\site-packages\cherrypy\wsgiserver\wsgiserver2.py", line 1948, in start
    self.tick()
  File "C:\Python27\Lib\site-packages\cherrypy\wsgiserver\wsgiserver2.py", line 2015, in tick
    s, ssl_env = self.ssl_adapter.wrap(s)
  File "C:\Python27\Lib\site-packages\cherrypy\wsgiserver\ssl_builtin.py", line 54, in wrap
    ssl_version=ssl.PROTOCOL_SSLv23)
  File "C:\Python27\lib\ssl.py", line 933, in wrap_socket
    ciphers=ciphers)
  File "C:\Python27\lib\ssl.py", line 601, in __init__
    self.do_handshake()
  File "C:\Python27\lib\ssl.py", line 830, in do_handshake
    self._sslobj.do_handshake()
SSLError: [SSL: HTTP_REQUEST] http request (_ssl.c:590)
```

Using pyOpenSSL the browser will show:

```
The client sent a plain HTTP request, but this server only speaks HTTPS on this port.
```

jaraco:
I think you may have a bug in your installation. CherryPy 8.1.0 doesn't have a wsgiserver2 module. I'd check your site-packages directory and make sure you have only one installation of cherrypy 8.1.0. Then report back if the issue persists.

Safihre:
Sorry about that, old log. 
Retested and used `print cherrypy.__version__` to confirm the version when starting the server:

```
V: 8.1.0

2016-09-07 20:40:02,651::ERROR::[_cplogging:217] [07/Sep/2016:20:40:02] ENGINE Error in HTTPServer.tick
Traceback (most recent call last):
  File "C:\Python27\Lib\site-packages\cherrypy\wsgiserver\__init__.py", line 2019, in start
    self.tick()
[...]

PR #1499, opened 2016-09-13
Fixes for builtin SSL provider

Safihre:
- Fixes error #1497 by not only checking at end of the error.
- Also fixes another error where the builtin provider will throw `The handshake operation timed out` even though a response was already send to the non-HTTPS client. I will admit that am not sure why this happens, but it does.

Tested on Ubuntu and Windows 10, both producing the errors described above.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,9 @@
+8.1.1
+-----
+
+* #1497: Handle errors thrown by ``ssl_module: 'builtin'``
+  when client opens connection to HTTPS port using HTTP.
+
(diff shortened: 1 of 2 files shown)


jaraco:
Looks reasonable. Could you add a new entry in the CHANGES.txt for the next bugfix version and describe for other users how the change impacts them? Then it'll be ready for a quick release.

Safihre:
Rebased changes to CHANGES.txt on last commit :)

webknjaz:
LGTM, merging.

Issue #1330, opened 2014-07-09
Move _escape to _cpcompat to avoid DeprecationWarning in Python3


bb-migration:
Originally reported by: **saaj (Bitbucket: [saaj](http://bitbucket.org/saaj), GitHub: [saaj](http://github.com/saaj))**
---

`cgi.escape` is [deprecated](https://docs.python.org/3/library/cgi.html#cgi.escape) since Python 3.2. `html.escape` is suggested replacement. `cgi.escape` is referenced from `_cperror.get_error_page`.
---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1330

PR #1513, opened 2016-10-26
Deprecated cgi in python3 #1330

paulohrpinheiro:
Use import especific to py2 or py3 to escape function from older cgi or newest html module in python3, conform #1330.

### DIFF
--- a/cherrypy/_cpcompat.py
+++ b/cherrypy/_cpcompat.py
@@ -341,7 +341,10 @@ except ImportError:
 
 # html module needed the argument quote=False because in cgi the default
 # is False. With quote=True the results differ.
+
 def escape_html(s, escape_quote=False):
-    """Replace special characters "&", "<" and ">" to HTML-safe sequences. When
-    escape_quote=True, escape (') and (") chars."""
+    """Replace special characters "&", "<" and ">" to HTML-safe sequences.
+
+    When escape_quote=True, escape (') and (") chars.
+    """
     return escape(s, quote=escape_quote)
--- a/cherrypy/_cpcompat.py
+++ b/cherrypy/_cpcompat.py
@@ -341,7 +341,7 @@ except ImportError:
 
 # html module needed the argument quote=False because in cgi the default
 # is False. With quote=True the results differ.
-def escape_html(s, quote=False):
+def escape_html(s, escape_quote=False):
     """Replace special characters "&", "<" and ">" to HTML-safe sequences. When
-    quote=True, escape (') and (") chars."""
-    return escape(s, quote=quote)
+    escape_quote=True, escape (') and (") chars."""
+    return escape(s, quote=escape_quote)
--- a/cherrypy/_cpcompat.py
+++ b/cherrypy/_cpcompat.py
@@ -333,12 +333,14 @@ except ImportError:
 
         return args
 
-if six.PY2:
-    from cgi import escape as escape_html
-else:
+# html module come in 3.2 version
+try:
     from html import escape
+except:
+    from cgi import escape
 
-    def escape_html(s):
-        """In py3 version is needed the argument quote=False to produce same results"""
-
-        return escape(s, quote=False)
+# html module needed the argument quote=False because in cgi the default
+# is False. With quote=True the results differ.
+def escape_html(s):
+    """Replace special characters "&", "<" and ">" to HTML-safe sequences."""
+    return escape(s, quote=False)
--- a/cherrypy/_cpcompat.py
+++ b/cherrypy/_cpcompat.py
@@ -334,7 +334,7 @@ except ImportError:
         return args
 
 if six.PY2:
-    from cgi import escape as _escape
+    from cgi import escape as escape_html
 else:
-    from html import escape as _escape
+    from html import escape as escape_html
 
--- a/cherrypy/_cpcompat.py
+++ b/cherrypy/_cpcompat.py
@@ -336,5 +336,9 @@ except ImportError:
 if six.PY2:
     from cgi import escape as escape_html
 else:
-    from html import escape as escape_html
+    from html import escape
 
+    def escape_html(s):
+        """In py3 version is needed the argument quote=False to produce same results"""
+
+        return escape(s, quote=False)
--- a/cherrypy/_cpcompat.py
+++ b/cherrypy/_cpcompat.py
@@ -336,11 +336,12 @@ except ImportError:
 # html module come in 3.2 version
 try:
     from html import escape
-except:
+except ImportError:
     from cgi import escape
 
 # html module needed the argument quote=False because in cgi the default
 # is False. With quote=True the results differ.
-def escape_html(s):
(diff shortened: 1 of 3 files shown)


cherrypy/_cpcompat.py
```diff
@@ -332,3 +332,13 @@ def _args_from_interpreter_flags():
             args.append('-W' + opt)
 
         return args
+
+if six.PY2:
```
webknjaz: I'd use try/except ImportError approach. `html.escape()` appeared in 3.2, so it will fail for 3.0-3.1.

cherrypy/_cpcompat.py
```diff
@@ -332,3 +332,13 @@ def _args_from_interpreter_flags():
...
+    from cgi import escape as escape_html
+else:
+    from html import escape
+
+    def escape_html(s):
```
webknjaz: You break function signature here. It's now impossible to pass custom `quote` argument explicitly. You should accept  `quote=False` and bypass it on call below.

cherrypy/_cpcompat.py
```diff
@@ -332,3 +332,13 @@ def _args_from_interpreter_flags():
             args.append('-W' + opt)
 
         return args
+
+if six.PY2:
+    from cgi import escape as escape_html
```
webknjaz: you **must not** fool user with another function name. especially when it's behavior differs. escape_html would be expected to have `quote=True`, whereas you break this expectation.

cherrypy/test/test_compat.py
```diff
@@ -21,3 +21,14 @@ def test_ntob_non_native(self):
...
+
+
+class EscapeTester(unittest.TestCase):
+    """Class to test escape_html function from _cpcompat.
+    """
+
+    def test_escape_quote(self):
+        """test_escape_quote - Verify the output for &<>" chars
+        """
+        self.assertEqual('xx&amp;&lt;&gt;"aa', compat.escape_html('xx&<>"aa'))
```
webknjaz: Add tests with passing different `quote=` explicitly.

cherrypy/_cpcompat.py
webknjaz: Yet, I'm still unsure which default should we fall back to.
@jaraco, shouldn't we escape html safely?

webknjaz (changes requested):
I appreciate the overall idea, but you should fix the pitfalls, I've mentioned.

paulohrpinheiro:
Hello!

Em 30-10-2016 18:00, Sviatoslav Sydorenko escreveu:

> I appreciate the overall idea, but you should fix the pitfalls, I've
> mentioned.

First of all, I thank the indicated corrections. I hope to make a good 
contribution.

> In cherrypy/_cpcompat.py
> (...)
> +if six.PY2:
> 
> I'd use try/except ImportError approach. |html.escape()| appeared in
> 3.2, so it will fail for 3.0-3.1.

OK.

> In cherrypy/_cpcompat.py
> https://github.com/cherrypy/cherrypy/pull/1513#pullrequestreview-6370151:
> (...)
> +if six.PY2:
> -    from cgi import escape as escape_html
>   +else:
> -    from html import escape
>   +
> -    def escape_html(s):
> 
> You break function signature here. It's now impossible to pass custom
> |quote| argument explicitly. You should accept |quote=False| and bypass
> it on call below.

See my remarks on the quote parameter below.

> In cherrypy/_cpcompat.py
> https://github.com/cherrypy/cherrypy/pull/1513#pullrequestreview-6370151:
> (...)
> +if six.PY2:
> -    from cgi import escape as escape_html
> 
> you _must not_ fool user with another function name. especially when
> it's behavior differs. escape_html would be expected to have
> |quote=True|, whereas you break this expectation.

OK. Now, escape_html is a function, the try/catch is only to the import. 
The quote argument is a problem. With quote=True I get different results:

[paulohrpinheiro@localhost cherrypy]$ python2
Python 2.7.12 (default, Sep 29 2016, 13:30:34)
[GCC 6.2.1 20160916 (Red Hat 6.2.1-2)] on linux2

> > > import cgi
> > > cgi.escape("""xx&<>"aa'""", quote=True)
> > > "xx&amp;&lt;&gt;&quot;aa'"

[paulohrpinheiro@localhost cherrypy]$ python3
Python 3.5.1 (default, Sep 19 2016, 10:16:17)
[GCC 6.1.1 20160621 (Red Hat 6.1.1-3)] on linux

> > > import html
> > > html.escape("""xx&<>"aa'""", quote=True)
> > > 'xx&amp;&lt;&gt;&quot;aa&#x27;'
> 
> In cherrypy/test/test_compat.py
> https://github.com/cherrypy/cherrypy/pull/1513#pullrequestreview-6370151:
> (...)
> Add tests with passing different |quote=| explicitly.

As I said above, quote = True produces different results between PY2 and 
PY3. Because this option is not used, and the current code uses quote = 
False, it is not better to force always quote = False?

I changed to always use quote = False explicitly, but if you think 
otherwise I can change that.

Thanks!

## 

Paulo Henrique Rodrigues Pinheiro
https://about.me/paulohrpinheiro

cherrypy/_cpcompat.py
```diff
@@ -332,3 +332,15 @@ def _args_from_interpreter_flags():
             args.append('-W' + opt)
 
         return args
+
+# html module come in 3.2 version
+try:
+    from html import escape
+except:
```
webknjaz: catch only `ImportError` here.

cherrypy/_cpcompat.py
```diff
@@ -332,3 +332,15 @@ def _args_from_interpreter_flags():
...
+    from cgi import escape
+
+# html module needed the argument quote=False because in cgi the default
+# is False. With quote=True the results differ.
+def escape_html(s):
```
webknjaz: This still needs to have `quote=False`

cherrypy/_cpcompat.py
```diff
@@ -332,3 +332,15 @@ def _args_from_interpreter_flags():
...
+    """Replace special characters "&", "<" and ">" to HTML-safe sequences."""
+    return escape(s, quote=False)
```
webknjaz: And bypass `quote=quote` into `escape()`

webknjaz (changes requested):

cherrypy/_cpcompat.py
paulohrpinheiro: Done.

cherrypy/_cpcompat.py
paulohrpinheiro: Done

cherrypy/_cpcompat.py
paulohrpinheiro: Done.

paulohrpinheiro:
@webknjaz, the fail in travis is a timeout in another piece of code. Please, what I do now?

webknjaz (approved):

webknjaz:
It's okay, I've re-run that build and it succeeded.

Issue #1322, opened 2014-05-26
HTTP instead of HTTPS


bb-migration:
Originally reported by: **Anonymous**

---

Source: cherrypy/process/servers.py
Class: ServerAdapter
Routine: _get_base
Line: if getattr(self.httpserver, 'ssl_certificate', None):

self.httpserver doesn't appear to have ssl_certificate.

Switching the line to either of the following will work:
if getattr(self, 'ssl_certificate', None):
if getattr(self.httpserver.ssl_adapter, 'certificate', None):

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1322

500swapnil:
I would like to help fix this issue

webknjaz:
Thanks, waiting for the PR then. Plz don't forget to add tests and make TravisCI build green :)

jaraco:
This issue was addressed in c8f07984 and d1be8384. If there are ongoing issues with the implementation, please file a new report explaining under what conditions the current logic fails expectations.

PR #1519, opened 2016-11-19
HTTP instead of HTTPS issue #1322 resolved

500swapnil:
Changed the line from
if getattr(self.httpserver, 'ssl_certificate', None):
 to
if getattr(self, 'ssl_certificate', None):

### DIFF
--- a/cherrypy/process/servers.py
+++ b/cherrypy/process/servers.py
@@ -185,7 +185,7 @@ class ServerAdapter(object):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if sys.version_info[0] > 2:    
+        if sys.version_info[0] > 2:
             if getattr(self.httpserver.ssl_adapter, 'certificate', None):
                 scheme = 'https'
                 if port != 443:
@@ -193,7 +193,7 @@ class ServerAdapter(object):
             else:
                 scheme = 'http'
                 if port != 80:
-                    host += ':%s' % port    
+                    host += ':%s' % port
         else:
             if getattr(self,'ssl_certificate', None):
                 scheme = 'https'
--- a/cherrypy/process/servers.py
+++ b/cherrypy/process/servers.py
@@ -185,14 +185,24 @@ class ServerAdapter(object):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if getattr(self,'ssl_certificate', None):
-            scheme = 'https'
-            if port != 443:
-                host += ':%s' % port
+        if sys.version_info[0] > 2:    
+            if getattr(self.httpserver.ssl_adapter, 'certificate', None):
+                scheme = 'https'
+                if port != 443:
+                    host += ':%s' % port
+            else:
+                scheme = 'http'
+                if port != 80:
+                    host += ':%s' % port    
         else:
-            scheme = 'http'
-            if port != 80:
-                host += ':%s' % port
+            if getattr(self,'ssl_certificate', None):
+                scheme = 'https'
+                if port != 443:
+                    host += ':%s' % port
+            else:
+                scheme = 'http'
+                if port != 80:
+                    host += ':%s' % port
 
         return '%s://%s' % (scheme, host)
 
--- a/cherrypy/process/servers.py
+++ b/cherrypy/process/servers.py
@@ -185,11 +185,11 @@ class ServerAdapter(object):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if getattr(self.httpserver.ssl_adapter, 'certificate', None):
+        if getattr(self, 'ssl_certificate', None):
             scheme = 'https'
             if port != 443:
                 host += ':%s' % port
-        else
+        else:
 	    scheme = 'http'
             if port != 80:
                 host += ':%s' % port
--- a/cherrypy/process/servers.py
+++ b/cherrypy/process/servers.py
@@ -185,7 +185,7 @@ class ServerAdapter(object):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if getattr(self.httpserver, 'ssl_adapter', None):
+        if getattr(self,'ssl_certificate', None):
             scheme = 'https'
             if port != 443:
                 host += ':%s' % port
--- a/cherrypy/process/servers.py
+++ b/cherrypy/process/servers.py
@@ -185,25 +185,14 @@ class ServerAdapter(object):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if sys.version_info[0] > 2:
-            if getattr(self.httpserver.ssl_adapter, 'certificate', None):
-                scheme = 'https'
-                if port != 443:
-                    host += ':%s' % port
-            else:
-                scheme = 'http'
-                if port != 80:
-                    host += ':%s' % port
-        else:
-            if getattr(self,'ssl_certificate', None):
-                scheme = 'https'
-                if port != 443:
-                    host += ':%s' % port
-            else:
-                scheme = 'http'
-                if port != 80:
-                    host += ':%s' % port
-
+        if getattr(self.httpserver.ssl_adapter, 'certificate', None):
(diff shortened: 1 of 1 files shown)


cherrypy/process/servers.py
```diff
@@ -185,7 +185,7 @@ def _get_base(self):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if getattr(self.httpserver, 'ssl_adapter', None):
+        if getattr(self,'ssl_certificate', None):
```
webknjaz: Please add/fix corresponding test for this.

webknjaz (changes requested):
Tests should not fail

cherrypy/process/servers.py
500swapnil: Only the test with Python:nightly build ( python 3.7 ) is failing. Tests with all other versions give positive results. How do I fix this?

webknjaz: I've restarted that build to check whether it's a temporary failure. Still, you need to add test for this case checking that what was failing before is fixed now.

500swapnil: It passed the tests now. However I will try to check the cause for previous failure.

500swapnil: I have committed the changes and added a check that will avoid that temporary failure.

cherrypy/process/servers.py
```diff
@@ -185,14 +185,24 @@ def _get_base(self):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if getattr(self.httpserver, 'ssl_adapter', None):
-            scheme = 'https'
-            if port != 443:
-                host += ':%s' % port
+        if sys.version_info[0] > 2:
```
webknjaz: what is this?

webknjaz (changes requested):
Please explain your changes

cherrypy/process/servers.py
webknjaz: 1) why is that check needed?
2) we use `six` for writing multi python code
3) where are new tests?

500swapnil: since the code was failing for certain 3.x versions of python. That check is needed for running the alternative code for python versions greater than 2 (i.e. 3.x). 
Instead of six.PY2 or six.PY3 , sys.version_info can also be used to find the current version of python.
And I didnt know what new tests to add. I just put that check to fix the error in 3.x versions.

cherrypy/process/servers.py
500swapnil: This condition checks whether the version of python is greater than 2 or not. Since the code failed temporarily failed the tests with certain python 3.x versions. The alternative code is executed after being checked by sys.version_info which works on python 3.x. The original code works fine with python 2.x

webknjaz: That's weird. Anyway, there's lot's of code duplication you introduced.

cherrypy/process/servers.py
webknjaz: Yeah, I know what the check does, but I won't accept such patch. Also, please explain me **why** this issue happens and **prove** with tests, that patch doesn't break anything. This is what I want basically,

webknjaz:
@500swapnil any luck?

500swapnil:
@webknjaz I tried finding the log file of the failure with the first build. But I cant find it anywhere. So I am not able to find out what the bug was. This patch just changes the server call method based on version so it cannot break anything. I can look into removing the duplication of code if you want but I am not that rigorous with servers on python so I dont know how or what tests to add to show that it wont break anything. I am trying my best to understand though.

webknjaz:
@500swapnil click on tick or crossing near commits @ https://github.com/cherrypy/cherrypy/pull/1519/commits

500swapnil:
@webknjaz I did. It shows a successful build from travis.ci for my first commit as well. Even though I had a failure the first time I sent PR. Shows no error with python: nightly build.

webknjaz:
Oh, it was a temporary issue with timeouts, so I just restarted it and it passed

500swapnil:
Okay so should I remove the extra test and duplication of code. Since my first change works fine for all versions?

webknjaz:
Version check is not a test. A test is a way to check your functionality. Like ones in [`cherrypy/test`](https://github.com/cherrypy/cherrypy/tree/master/cherrypy/test) folder.

500swapnil:
Okay. I have removed the version check. And you want a separate test file by the name test_servers.py?
Because I dont know how to create test files.

webknjaz:
I don't see it removed

webknjaz:
I think, you should find existing test module testing this method you change and add extend tests found there.

cherrypy/process/servers.py
```diff
@@ -185,15 +185,14 @@ def _get_base(self):
...
             return ''
         host, port = self.bind_addr
-        if getattr(self.httpserver, 'ssl_adapter', None):
+        if getattr(self.httpserver.ssl_adapter, 'certificate', None):
             scheme = 'https'
             if port != 443:
                 host += ':%s' % port
-        else:
-            scheme = 'http'
+        else
```
webknjaz: BUG.

cherrypy/process/servers.py
```diff
@@ -185,15 +185,14 @@ def _get_base(self):
...
+	    scheme = 'http'
```
webknjaz, line 193: BUG

cherrypy/process/servers.py
```diff
@@ -185,15 +185,14 @@ def _get_base(self):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if getattr(self.httpserver, 'ssl_adapter', None):
+        if getattr(self.httpserver.ssl_adapter, 'certificate', None):
```
webknjaz: what if there's no `ssl_adapter` attribute at all?

cherrypy/process/servers.py
```diff
@@ -185,15 +185,14 @@ def _get_base(self):
...
             scheme = 'https'
             if port != 443:
                 host += ':%s' % port
-        else:
-            scheme = 'http'
+        else
+	    scheme = 'http'
             if port != 80:
                 host += ':%s' % port
-
```
webknjaz, line 196: DON'T

webknjaz (changes requested):

cherrypy/process/servers.py
```diff
@@ -185,15 +185,14 @@ def _get_base(self):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if getattr(self.httpserver, 'ssl_adapter', None):
+        if getattr(self, 'ssl_certificate', None):
```
webknjaz, line 188: why do you do this?

500swapnil:
Wait I will remove all changes and try again

cherrypy/process/servers.py
```diff
@@ -185,15 +185,14 @@ def _get_base(self):
...
             scheme = 'https'
             if port != 443:
                 host += ':%s' % port
         else:
-            scheme = 'http'
+	    scheme = 'http'
```
webknjaz, line 193: indentation is broken here

cherrypy/process/servers.py
```diff
@@ -185,15 +185,14 @@ def _get_base(self):
...
             if port != 80:
                 host += ':%s' % port
-
```
webknjaz, line 196: don't remove this line

webknjaz (changes requested):

500swapnil:
I am closing this PR and opening a new one since I have messed up this one.

webknjaz:
Well, anyway I would merge it with Squash (combining all your changes into a single commit), so mess wouldn't get into master :)

500swapnil:
Okay I will try and find out why the issue appears. Are you going to merge this pull request?

webknjaz:
This PR is already closed. No one can merge it.
I'll merge the new one if I understand what are the consequences (that is why I asked you for analysis). Also, having test cases is a must.

Issue #1523, opened 2016-11-22
[enhancement]Request to add attribute "ssl_ciphers" to "cherrypy._cpserver.Server" as "ssl_certificate".


WXITWangPeng:
Hi Team,
    Our security team got a request recently, need to dynamicly set cipher suites of https connections.
   Since Python(version >= 2.7.9) already supports API 

"context.set_ciphers()", then we can add attribute "ssl_ciphers", and transfer it to "ssl_builtin.py". Thus, if cherrypy use builtin ssl lib, it can dynamicly configure the cipher suites.
   
   I have a solution to add the attribute as the attachment, could you please help to review it?

[cherry_modify.zip](https://github.com/cherrypy/cherrypy/files/605285/cherry_modify.zip)

Thanks,
Cindy

webknjaz:
Please create a Pull Request for this. It's not 90-s: we don't exchange emails with patches nowadays.

webknjaz:
Okay, you've sent PR #1524. Closing the issue as a duplicate. Let's move discussion to that PR.

PR #1524, opened 2016-11-22
Add attribute "ssl_ciphers" to support configuration cipher suites by cherrypy

WXITWangPeng:
Hi Team,
    Our security team got a request recently, need to dynamiclly set cipher suites of https connections.

     Since Python(version >= 2.7.9)support "context.set_ciphers()", then we can add attribute "ssl_ciphers", and transfer it to "ssl_builtin.py". Thus, if cherrypy use builtin ssl lib, it can dynamicly configure the cipher suites.
   
   I have a solution to add the attribute, could you please merge it once you pass the review?

Thanks,
Cindy

UPD: cc #1523

### DIFF
--- a/cherrypy/_cpserver.py
+++ b/cherrypy/_cpserver.py
@@ -113,6 +113,9 @@ class Server(ServerAdapter):
     ssl_private_key = None
     """The filename of the private key to use with SSL."""
 
+     ssl_ciphers = None
+    """The ciphers list of SSL"""
+    
     if six.PY3:
         ssl_module = 'builtin'
         """The name of a registered SSL adaptation module to use with
--- a/cherrypy/_cpserver.py
+++ b/cherrypy/_cpserver.py
@@ -113,7 +113,7 @@ class Server(ServerAdapter):
     ssl_private_key = None
     """The filename of the private key to use with SSL."""
 
-     ssl_ciphers = None
+    ssl_ciphers = None
     """The ciphers list of SSL"""
     
     if six.PY3:
--- a/cherrypy/_cpserver.py
+++ b/cherrypy/_cpserver.py
@@ -115,6 +115,7 @@ class Server(ServerAdapter):
 
     ssl_ciphers = None
     """The ciphers list of SSL"""
+    
     if six.PY3:
         ssl_module = 'builtin'
         """The name of a registered SSL adaptation module to use with
--- a/cherrypy/_cpserver.py
+++ b/cherrypy/_cpserver.py
@@ -115,7 +115,7 @@ class Server(ServerAdapter):
 
     ssl_ciphers = None
     """The ciphers list of SSL"""
-    
+
     if six.PY3:
         ssl_module = 'builtin'
         """The name of a registered SSL adaptation module to use with
--- a/cherrypy/_cpserver.py
+++ b/cherrypy/_cpserver.py
@@ -115,7 +115,6 @@ class Server(ServerAdapter):
 
     ssl_ciphers = None
     """The ciphers list of SSL"""
-    
     if six.PY3:
         ssl_module = 'builtin'
         """The name of a registered SSL adaptation module to use with
--- a/cherrypy/wsgiserver/__init__.py
+++ b/cherrypy/wsgiserver/__init__.py
@@ -1771,7 +1771,7 @@ class SSLAdapter(object):
           socket file object``
     """
 
-    def __init__(self, certificate, private_key, ciphers, certificate_chain=None):
+    def __init__(self, certificate, private_key, certificate_chain=None, ciphers):
         self.certificate = certificate
         self.private_key = private_key
         self.ciphers = ciphers
--- a/cherrypy/wsgiserver/__init__.py
+++ b/cherrypy/wsgiserver/__init__.py
@@ -1771,7 +1771,7 @@ class SSLAdapter(object):
           socket file object``
     """
 
-    def __init__(self, certificate, private_key, certificate_chain=None, ciphers):
+    def __init__(self, certificate, private_key, certificate_chain=None, ciphers=None):
         self.certificate = certificate
         self.private_key = private_key
         self.ciphers = ciphers
--- a/cherrypy/wsgiserver/__init__.py
+++ b/cherrypy/wsgiserver/__init__.py
@@ -1771,9 +1771,10 @@ class SSLAdapter(object):
           socket file object``
     """
 
-    def __init__(self, certificate, private_key, certificate_chain=None):
+    def __init__(self, certificate, private_key, ciphers, certificate_chain=None):
         self.certificate = certificate
         self.private_key = private_key
+        self.ciphers = ciphers
         self.certificate_chain = certificate_chain
 
     def wrap(self, sock):
--- a/cherrypy/wsgiserver/ssl_builtin.py
+++ b/cherrypy/wsgiserver/ssl_builtin.py
@@ -36,17 +36,21 @@ class BuiltinSSLAdapter(wsgiserver.SSLAdapter):
 
     certificate_chain = None
     """The filename of the certificate chain file."""
+    
+    ciphers = None
+    """The ciphers list of SSL"""
 
     """The ssl.SSLContext that will be used to wrap sockets where available
     (on Python > 2.7.9 / 3.3)
     """
     context = None
 
-    def __init__(self, certificate, private_key, certificate_chain=None):
+    def __init__(self, certificate, private_key, ciphers, certificate_chain=None):
         if ssl is None:
             raise ImportError('You must install the ssl module to use HTTPS.')
         self.certificate = certificate
         self.private_key = private_key
+        self.ciphers = ciphers
         self.certificate_chain = certificate_chain
         if hasattr(ssl, 'create_default_context'):
             self.context = ssl.create_default_context(
@@ -54,6 +58,7 @@ class BuiltinSSLAdapter(wsgiserver.SSLAdapter):
                 cafile=certificate_chain
             )
             self.context.load_cert_chain(certificate, private_key)
+            self.context.set_ciphers(ciphers)
 
     def bind(self, sock):
         """Wrap and return the given socket."""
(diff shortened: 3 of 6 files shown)


cherrypy/wsgiserver/ssl_pyopenssl.py
```diff
@@ -157,6 +157,7 @@ class pyOpenSSLAdapter(wsgiserver.SSLAdapter):
 
     private_key = None
     """The filename of the server's private key file."""
+
```
webknjaz: @LenovoCindywang Please get rid of this issue

cherrypy/_cpserver.py
```diff
@@ -113,6 +113,9 @@ def _set_socket_host(self, value):
     ssl_private_key = None
     """The filename of the private key to use with SSL."""
 
+    ssl_ciphers = None
+    """The ciphers list of SSL"""
+
```
webknjaz: @LenovoCindywang plz fix this

cherrypy/wsgiserver/ssl_builtin.py
```diff
@@ -37,23 +37,28 @@ class BuiltinSSLAdapter(wsgiserver.SSLAdapter):
...
+    ciphers = None
+    """The ciphers list of SSL"""
+
     """The ssl.SSLContext that will be used to wrap sockets where available
     (on Python > 2.7.9 / 3.3)
     """
     context = None
 
-    def __init__(self, certificate, private_key, certificate_chain=None):
+    def __init__(self, certificate, private_key, ciphers, certificate_chain=None):
```
webknjaz: When changing function signature, make sure you find all its invocations. In this case, identifying **all of them** is way too hard. At least don't add new argument somewhere in the middle of existing ones. Adding it to the end would be safer.

webknjaz (changes requested):

cherrypy/wsgiserver/__init__.py
```diff
@@ -1771,9 +1771,10 @@ class SSLAdapter(object):
           socket file object``
     """
 
-    def __init__(self, certificate, private_key, certificate_chain=None):
+    def __init__(self, certificate, private_key, certificate_chain=None, ciphers):
```
webknjaz: This is an invalid function signature. `ciphers` should have a default value.

cherrypy/wsgiserver/ssl_builtin.py
```diff
@@ -37,23 +37,27 @@ class BuiltinSSLAdapter(wsgiserver.SSLAdapter):
     certificate_chain = None
     """The filename of the certificate chain file."""
 
+    ciphers = None
```
webknjaz, line 40: Why do you add it only to this module?

cherrypy/wsgiserver/ssl_builtin.py
```diff
@@ -37,23 +37,27 @@ class BuiltinSSLAdapter(wsgiserver.SSLAdapter):
...
+    """The ciphers list of SSL"""
     """The ssl.SSLContext that will be used to wrap sockets where available
     (on Python > 2.7.9 / 3.3)
     """
     context = None
 
-    def __init__(self, certificate, private_key, certificate_chain=None):
+    def __init__(self, certificate, private_key, certificate_chain=None, ciphers):
```
webknjaz: Same

cherrypy/wsgiserver/__init__.py
webknjaz: If you want it to be mandatory, just add a corresponding check.

cherrypy/_cpserver.py
```diff
@@ -113,6 +113,8 @@ def _set_socket_host(self, value):
     ssl_private_key = None
     """The filename of the private key to use with SSL."""
 
+    ssl_ciphers = None
```
webknjaz, line 116: I'm not sure whether it should be here.

webknjaz (changes requested):
Existing tests fail, new test cases are missing.

cherrypy/wsgiserver/ssl_builtin.py
WXITWangPeng: I used to add it to ssl_pyopenssl.py, however it will cause build error. And also, this requirement is in the case of using "built-in" ssl module.

cherrypy/wsgiserver/ssl_builtin.py
```diff
@@ -37,23 +37,27 @@ class BuiltinSSLAdapter(wsgiserver.SSLAdapter):
...
 
+    ciphers = None
+    """The ciphers list of SSL"""
     """The ssl.SSLContext that will be used to wrap sockets where available
     (on Python > 2.7.9 / 3.3)
     """
     context = None
 
-    def __init__(self, certificate, private_key, certificate_chain=None):
+    def __init__(self, certificate, private_key, certificate_chain=None, ciphers=None):
```
WXITWangPeng, line 48: Why this error will generated since I only add a parameter to this function..

webknjaz: Codacy bot is only static analysis tool.

cherrypy/wsgiserver/ssl_builtin.py
```diff
@@ -37,23 +37,28 @@ class BuiltinSSLAdapter(wsgiserver.SSLAdapter):
...
         self.private_key = private_key
+        self.ciphers = ciphers
         self.certificate_chain = certificate_chain
         if hasattr(ssl, 'create_default_context'):
             self.context = ssl.create_default_context(
                 purpose=ssl.Purpose.CLIENT_AUTH,
                 cafile=certificate_chain
             )
             self.context.load_cert_chain(certificate, private_key)
+            if(self.ciphers != None):
```
webknjaz: 1) This violates the code style: you should have space after `if` and parentheses are normally omitted.
2) In Python checking against `None` would be done with `self.ciphers is None`.

cherrypy/_cpserver.py
webknjaz: What is the point of having it here? Where is it being filled in?

webknjaz (changes requested):
Could you please check for a way of enabling the HTTPS test in the CI?

cherrypy/_cpserver.py
WXITWangPeng: Since cherrypy and python are alll opensource code, we can't add many private behavior to them.
Sometimes, we need to dynamically controled the cipher suites as customer's request.
Like below:
```python
CIPHERS = (
    'ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+HIGH:'
    'DH+HIGH:RSA+AESGCM:RSA+AES:RSA+HIGH:!aNULL:'
    '!eNULL:!MD5:!DSS:!RC4:!DES:!3DES:!BF:!BF+CFB:!BF+ECB:!BF+OFB'
)
              cherrypy.config.update({'tools.secureheaders.on' : True})
              cherrypy.config.update({'server.socket_port': 443})
              cherrypy.config.update({'server.ssl_module': "builtin"})
              cherrypy.config.update({'server.ssl_certificate':"/cfg/cert/management-server.cert"})
              cherrypy.config.update({'server.ssl_private_key':"/cfg/cert/management-server.key"})
              cherrypy.config.update({'server.ssl_ciphers':CIPHERS})
```

cherrypy/wsgiserver/ssl_builtin.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, certificate, private_key, certificate_chain=None, ciphers=Non
                 cafile=certificate_chain
             )
             self.context.load_cert_chain(certificate, private_key)
-            if(self.ciphers != None):
+            if (self.ciphers != None):
```
webknjaz: You didn't fix the issue with parens and the issue with comparison.

webknjaz (changes requested):

cherrypy/_cpserver.py
```diff
@@ -113,6 +113,8 @@ def _set_socket_host(self, value):
...
+    """The ciphers list of SSL"""
```
webknjaz, line 117: Please add one empty line after this docstring.

cherrypy/wsgiserver/ssl_builtin.py
```diff
@@ -37,23 +37,28 @@ class BuiltinSSLAdapter(wsgiserver.SSLAdapter):
     certificate_chain = None
     """The filename of the certificate chain file."""
 
+    ciphers = None
+    """The ciphers list of SSL"""
```
webknjaz, line 41: Please add an empty line below this docstring

cherrypy/_cpserver.py
webknjaz: @LenovoCindywang you added lots of trailing whitespaces to this line (not just new line), please remove them.

jaraco:
@webknjaz: Are you happy with this PR except for the two stylistic issues? Do you want to withhold acceptance until tests are added?

cherrypy/wsgiserver/ssl_builtin.py
```diff
@@ -37,23 +37,29 @@ class BuiltinSSLAdapter(wsgiserver.SSLAdapter):
...
         self.private_key = private_key
+        self.ciphers = ciphers
         self.certificate_chain = certificate_chain
         if hasattr(ssl, 'create_default_context'):
             self.context = ssl.create_default_context(
                 purpose=ssl.Purpose.CLIENT_AUTH,
                 cafile=certificate_chain
             )
             self.context.load_cert_chain(certificate, private_key)
+            if self.ciphers != None :
```
webknjaz: There should be no whitespace before the colon

webknjaz: And the check must be `is not None`

webknjaz:
@jaraco 
I was going to fix styling myself, but it has got postponed..
My worst fear is to merge the PR without understanding all the consequences of doing this. HTTPS testing is disabled in CI and AFAIK there are several bugs with HTTPS which don't get resolved. So yes, I would prefer having tests.
On the other hand, we could merge this after several style fixes.

webknjaz (changes requested):

jaraco:
Hi Cindy. I have to apologize. This PR got hung up on some stylistic issues and never got accepted. And now with CherryPy 9, the conflicts with the changes to wsgiserver are irreconcilable. I'm going to have to ask you to re-formulate your PR, first to the [cheroot project](/cherrypy/cheroot) (where wsgiserver was moved to) and then the rest to this project once the cheroot changes are accepted and released. Thanks, and sorry for the trouble.

PR #1525, opened 2016-11-22
Fix issue with ssl_certificate lookup

### LINKED ISSUE #1322: HTTP instead of HTTPS

500swapnil:
Fix #1322

### DIFF
--- a/cherrypy/process/servers.py
+++ b/cherrypy/process/servers.py
@@ -185,7 +185,7 @@ class ServerAdapter(object):
         if not self.httpserver:
             return ''
         host, port = self.bind_addr
-        if getattr(self.httpserver, 'ssl_adapter', None):
+        if getattr(self, 'ssl_certificate', None):
             scheme = 'https'
             if port != 443:
                 host += ':%s' % port


webknjaz:
I need a full investigation/analysis on why this issue appears and why do you think this is the fix. Also, I'm waiting for test cases to be added.

500swapnil:
Okay I will try and find out why this issue appears

jaraco:
Thanks for the contrib, but when I looked into the issue, it seems to be already addressed in the current code. If you are having issues with the current version, please file a separate ticket explaining under what conditions the current implementation is failing.

PR #1531, opened 2016-12-16
Fix for issue 1530

lbolla:
The first commit fixes the issue with TypeError being swallowed by decorated handlers.
The other two commits are required in order to run the test suite correctly.


jaraco:
Thanks for this. Fix is rolling out as 8.1.3.

PR #1533, opened 2016-12-23
Popargs documentation: Use matching parameter name

tn-osimis:
The explanation for that piece of sample code below it refers to the parameter which represents the band name as "band", whereas the code uses "name".  We change the code to use "band" as well.

This also fixes a typo in the same explanation.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -185,7 +185,7 @@ understand the structure of your URL.
 This works similarly to `_cp_dispatch` but, as said above, is more
 explicit and localized. It says:
 
-- take the first segment and store it into a parameter name `band`
+- take the first segment and store it into a parameter name `name`
 - take again the first segment (since we removed the previous first) 
   and store it into a parameter named `title`
 
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -185,7 +185,7 @@ understand the structure of your URL.
 This works similarly to `_cp_dispatch` but, as said above, is more
 explicit and localized. It says:
 
-- take the first segment and store it into a parameter name `name`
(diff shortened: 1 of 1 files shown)


docs/advanced.rst
```diff
@@ -164,28 +164,28 @@ understand the structure of your URL.
...
-    @cherrypy.popargs('name')
+    @cherrypy.popargs('band')
     class Band(object):
         def __init__(self):
             self.albums = Album()
 
         @cherrypy.expose
-        def index(self, name):
-            return 'About %s...' % name
+        def index(self, band):
```
jaraco: Here, 'name' is a better word to describe the parameter. It's an attribute of the Band, the "name" of the band.

tn-osimis:
@jaraco Gotcha, doing the opposite then.

jaraco:
I agree. `band` is better than `name` here. After all, the Album could have a `name` and the track could have a `name`. The naming scheme is inconsistent.

This PR is exposing a fundamental challenge of computer science - naming things. Since what you're receiving in the parameter isn't in fact a band object, "band" doesn't seem quite right either, but you do need to disambiguate the band's name from the names of other things. I'd use 'band_name' to capture the band name... and "album_title" and "track_number" and "track_name".

jaraco:
Also, sorry for pushing conflicting changes. I thought I was making changes apart from where you had edits.

tn-osimis:
Okay. I did rebase to act on your initial comment (deleted I think) but now I cannot see where you pushed the changes you mention. Honestly, whatever works, feel free to close this if you're merging the changes from somewhere else.

PR #1542, opened 2016-12-29
Allow binding to an ephemeral port. Fixes #645.

### LINKED ISSUE #645: A value of 0 for socket_port should cause CP to bind to an available unused port

---

When socket_port is specified as 0, an exception is thrown as follows:

```
File
"/localhome/raghu/localwork/cherrypy/svn.cherrypy.org/trunk/cherrypy/_cpserver.py",
line 69, in quickstart
    self.start()
  File
"/localhome/raghu/localwork/cherrypy/svn.cherrypy.org/trunk/cherrypy/_cpserver.py",
line 95, in start
    self._start_http(httpserver)
  File
"/localhome/raghu/localwork/cherrypy/svn.cherrypy.org/trunk/cherrypy/_cpserver.py",
line 118, in _start_http
    self.wait(httpserver)
  File
"/localhome/raghu/localwork/cherrypy/svn.cherrypy.org/trunk/cherrypy/_cpserver.py",
line 159, in wait
    wait_for_occupied_port(*bind_addr)
  File
"/localhome/raghu/localwork/cherrypy/svn.cherrypy.org/trunk/cherrypy/_cpserver.py",
line 247, in wait_for_occupied_port
    raise IOError(msg)
IOError: Port 0 not bound on 'localhost'
```

I am using the latest code from the trunk. It appears to me that even though wsgiserver is properly passing 0 to socket.bind (and hence binding to an unused port), other parts of the code still use "configured" port which is 0. 

Reported by draghuram@gmail.com

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/645

jaraco:


cherrypy/process/servers.py
```diff
@@ -237,8 +243,22 @@ def wait(self):
             # Wait for port to be occupied if not running via socket-activation
             # (for socket-activation the port will be managed by systemd )
             if isinstance(self.bind_addr, tuple):
-                host, port = self.bind_addr
-                wait_for_occupied_port(host, port)
+                host, port = self.bound_addr
```
webknjaz: Perhaps `port` should be `_`

cherrypy/process/servers.py
```diff
@@ -182,13 +175,26 @@ def start(self):
 
         self.wait()
         self.running = True
-        self.bus.log('Serving on %s' % on_what)
+        self.bus.log('Serving on %s' % self.description)
     start.priority = 75
 
+    @property
+    def description(self):
```
webknjaz, line 182: Currently `self.description` access invokes `@property`'s body every time. Shouldn't we have some cache for its value?

cherrypy/wsgiserver/__init__.py
```diff
@@ -2052,11 +2052,13 @@ def bind(self, family, type, proto=0):
         if self.ssl_adapter is not None:
             self.socket = self.ssl_adapter.bind(self.socket)
 
+        host, port = self.bind_addr[:2]
```
webknjaz, line 2055: Nit: `port` -> `_`

webknjaz (approved):
LGTM, but it would be nice to add a test for port 0.

cherrypy/process/servers.py
jaraco: I personally think `port` is clearer. When I follow the codacy link, it says "Having unused variable in a project can increase the maintainance [sic] cost, proliferating dead code and increasing the cognitive burden of the programmer." In this regard, I believe Codacy is mistaken. The variable provides clarity as to what the other part of the pair is, even if it's unused.

I do wonder, though, if it was a mistake to have removed the 'wait_for_occupied_port'.

cherrypy/process/servers.py
jaraco: Sounds like a premature optimization, especially when the only known invocation only gets invoked once per start call.

cherrypy/process/servers.py
jaraco: Aha. 'host' is also unused. It was the `host, port = self.bound_addr` which is unused and needs to be removed.

cherrypy/wsgiserver/__init__.py
jaraco: In this case, port is used.

jaraco:
I'm looking into writing a test, but there appears to be no prior art on which to derive. I've tried this test:

```
class TestBinding:
    def test_bind_ephemeral_port(self, capsys):
        """
        A server configured to bind to port 0 will bind to an ephemeral
        port and indicate that port number on startup.
        """
        cherrypy.config.reset()
        bind_ephemeral_conf = {
            'server.socket_port': 0,
        }
        cherrypy.config.update(bind_ephemeral_conf)
        cherrypy.engine.start()
        import re, time
        time.sleep(3)
        cherrypy.engine.stop()
        out, err = capsys.readouterr()
        assert re.match('Serving on http://.*:(?P<port>\d+)/', out)
```

But at the assertion, both out and err are empty strings. I've tried using `capfd` instead of `capsys` but that's no different. It seems there's something in the test suite that's causing the output to be suppressed from the engine startup.

jaraco:
Thanks for the review, @webknjaz. I'm releasing this as v8.7.0.

cherrypy/wsgiserver/__init__.py
webknjaz: `host = self.bind_addr[0]` would be better, since port is not used in this method

Issue #1481, opened 2016-08-23
Synchronize with cheroot


jaraco:
I very much appreciate the work that cheroot has done to factor out the HTTP Server portion of CherryPy into a standalone HTTP Server.

The effort seems to have been abandoned. I take some blame in that as well.

I'd like to revive the effort, but I see a few obstactles.
1. Cheroot was released, but never officially. The [only releases are beta releases](https://pypi.python.org/simple/cheroot/).
2. The project appear to be a fork of CherryPy, but the code base diverged a long time ago and was never consumed by CherryPy, so reconciling the differences may be a massive effort.
3. The [project repo](/cherrypy/cheroot) has issues disabled.

Now that CherryPy allows for dependencies to be defined, I think it's at a better place to factor out the wsgiserver. 

Here's what I propose:
1. Abandon the existing cheroot codebase. Maybe there are some gems there, and they'll remain in the history in case someone wants to dig them up.
2. Replace the existing cheroot with wsgiserver from cherrypy. As a drop-in replacement, it will be compatible with CherryPy.
3. Keep the cheroot tests that pass but disable or remove the rest.
4. Document CherryPy tests passing as a temporary condition for acceptable stability.
5. Cut the first stable release (5.0) and iterate from there.
6. Re-enable issues in Cheroot.

Then, iterate on the goals and shortcomings.

Thoughts?

jaraco:
I discovered that the /cherrypy/cheroot project, although it's based on the same cherrypy history, due to the transition from Mercurial to Git, no longer is considered related to the cherrypy history as found here on Github. Therefore, what I've done is rebase the cheroot 4.0.0b2 changes onto a checkout of cherrypy (still against 3.2.2 as before), and pushed those changes up to the cheroot origin under the tag `archive/cheroot-4`. Now I'll push the latest cherrypy master (with a related git history) up as the cheroot master and re-enact the refactor into a separate project.

jaraco:
I've also tagged the old master as archive/old-cheroot-master, in case anybody ever wants to go find that history.

jaraco:
Now in the cheroot master, which is up-to-date with the cherrypy master, I'll review the refactor that [happened here](https://github.com/cherrypy/cheroot/commit/577b83acef2281d11537fb7974c913b4f7a5bd3c) and try to re-apply it to the current CherryPy project tree.

jaraco:
[...]

PR #1548, opened 2017-01-14
Use cheroot for server

jaraco:
Per the plan in #1481, this PR extracts the code from wsgiserver and implants it into the new 5.0 release of cheroot. Work still remaining to be done - update the tests in cheroot to run against cheroot, capturing the server functionality, and then remove those tests from the CherryPy suite. At least now, the main code exists in exactly one place so won't diverge as it did for cheroot 4.x and prior releases of CherryPy's wsgiserver.


CHANGES.rst
```diff
@@ -1,3 +1,9 @@
+v9.0.0
+-----
+
+* #1481: Move functionality from cherrypy.wsgiserver to
+  the `cheroot 5.0 <https://pypa.io/project/cheroot/5.0`_.
```
webknjaz: this link is broken

cherrypy/_cpnative_server.py
```diff
@@ -122,7 +123,7 @@ def __init__(self, server_adapter=cherrypy.server):
                        self.server_adapter.socket_file or
                        None)
 
-        wsgiserver.HTTPServer.__init__(
+        cheroot.server.HTTPServer.__init__(
```
webknjaz, line 126: why not just `super().__init__(...args)`?

cherrypy/_cpwsgi_server.py
```diff
@@ -31,7 +34,7 @@ def __init__(self, server_adapter=cherrypy.server):
                        None)
 
         self.wsgi_version = self.server_adapter.wsgi_version
-        s = wsgiserver.CherryPyWSGIServer
+        s = cheroot.wsgi.WSGIServer
```
webknjaz, line 37: m/b use `super()`?

webknjaz (dismissed):

CHANGES.rst
jaraco: Yep. Typo in the domain name. Should be pypi.org. I'll correct it.

cherrypy/_cpnative_server.py
jaraco: Probably super would be better, but I'll leave that as a change to follow. I'm aiming for minimal flux to adopt cheroot.

The-Compiler:
As someone who only uses `cherrypy.wsgiserver` (for a end-to-end testsuite with Flask and httpbin): I really appreciate this - thanks! 👍

Issue #1332, opened 2014-07-13
Use portend for port checking/waiting


bb-migration:
Originally reported by: **Jason R. Coombs (Bitbucket: [jaraco](http://bitbucket.org/jaraco), GitHub: [jaraco](http://github.com/jaraco))**

---

I've extracted and made more general purpose the port checking routines in cherrypy.process.servers in a new package called [portend](https://pypi.python.org/pypi/portend).

CherryPy should consider using that package.

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1332

webknjaz:
@jaraco to use it, first switch tests to run against all supported versions.
And there's a dependency for `tempora.timing`. Don't you think it's an overhead to depend on another package just to use a stopwatch?

jaraco:
It is extra overhead, and yes, it's worth the overhead. After several years of accepting that 'copy/paste' is acceptable, I'm of the opinion that if someone is willing to maintain a library with suitable functionality that it's better to use that library than to create a fork of that functionality (by inlining it). The maintenance costs of subtly divergent implementations of essentially the same functionality are far greater than the costs of incorporating dependencies.

webknjaz:
Okay, I agree with that

PR #1554, opened 2017-01-19
Use portend for port checks

jaraco:
Following up on the proposal in #1332, use portend, which was itself inspired by this CherryPy code, to perform the checks.

I've marked this as a work in progress, as I expect it to land after #1548, so the changelog still needs to be updated. Otherwise, I believe it's ready.


CHANGES.rst
```diff
@@ -1,3 +1,24 @@
...
+* #1332: CherryPy now uses `portend
+  <https://pypi.org/project/portend>`_ for checking and
+  waiting on ports for startup and teardown checks. The
+  following names are no longer present:
+
+  - cherrypy._cpserver.client_host
+  - cherrypy._cpserver.check_port
+  - cherrypy._cpserver.wait_for_free_port
+  - cherrypy._cpserver.wait_for_occupied_port
+  - cherrypy.process.servers.check_port
```
webknjaz, line 13: According to note below this one is available

cherrypy/process/servers.py
```diff
@@ -121,6 +121,14 @@ def index(self):
 import sys
 import time
 import warnings
+import contextlib
+
+import portend
+
+
+class Timeouts:
```
webknjaz, line 129: I don't think it's appropriate to use class for storing constants

cherrypy/test/test_states.py
```diff
@@ -110,7 +112,7 @@ def test_0_NormalStateFlow(self):
 
         host = cherrypy.server.socket_host
         port = cherrypy.server.socket_port
-        self.assertRaises(IOError, cherrypy._cpserver.check_port, host, port)
+        self.assertRaises(IOError, portend._check_port, host, port, timeout=0.1)
```
webknjaz: Codacy is right. `portend` should expose public API instead of private functions.

cherrypy/process/servers.py
jaraco: Why not? Zen of python says "Namespaces are one honking great idea -- let's do more of those!"

What would be better? A namedtuple?

```
timeouts = collections.namedtuple('timeouts', 'occupied free')(5, 1)
```

Nah, that's impossible to read.

In any other language (C, C++, Java), you would group constants in an enum or struct or class. It seems perfectly sane to do the same in Python, applying DRY to the scope which is "constants about timeouts". It's codacy that's at fault here, demanding docstrings on classes which are simple and self-documenting.

cherrypy/test/test_states.py
jaraco: Yes, this is definitely bad form. I allowed it since this was for a test and not in the code. It might be better to make that a public function. I didn't want to make it public in portend because it exposes somewhat obscure implementation details. But the need for its usage here and the fact that I've found the need for it elsewhere suggests that I should invest the time and expose a proper function in portend.

CHANGES.rst
jaraco: I don't follow. `portend._check_port` is used, but `cherrypy.process.servers.check_port` has been removed.

CHANGES.rst
```diff
@@ -1,3 +1,24 @@
...
+  - cherrypy.process.servers.wait_for_free_port
+  - cherrypy.process.servers.wait_for_occupied_port
+
+  For ``check_port``, use
+  ``cherrypy.process.servers.check_port`` directly.
```
webknjaz: I was talking about this line

cherrypy/process/servers.py
webknjaz: Well, I vote for having just two constants here. It's KISS. I really like this talk https://www.youtube.com/watch?v=o9pEzgHorH0, I think it's applicable, here. Two constants, which are not going to be imported somewhere externally don't need another container around them, just module itself.

cherrypy/test/test_states.py
webknjaz: Sounds like a plan :)

cherrypy/test/test_states.py
```diff
@@ -112,7 +112,7 @@ def test_0_NormalStateFlow(self):
 
         host = cherrypy.server.socket_host
         port = cherrypy.server.socket_port
-        self.assertRaises(IOError, portend._check_port, host, port, timeout=0.1)
+        portend.occupied(host, port, timeout=0.1)
```
webknjaz, line 115: Did it stop raising `IOError`?

cherrypy/process/servers.py
jaraco: I remember that talk; I asked the [second question](https://youtu.be/o9pEzgHorH0?t=21m19s). I certainly absorbed it and factor it into my philosophy. It's nice to review it again. I'm a big fan of functools and functional programming. My first formal training was in Scheme. I do apply KISS frequently. However, I don't think Diederich's criticisms apply here. The purpose is clear without having to trace to the origin.

I do abhore `ASPECT_ATTRIBUTE=val` constants in code. They are crude and clumsy. For example, if you want to inspect all attributes relating to `ASPECT`, you have to query all globals in the module and find the ones whose names match a pattern. With `class Timeout`, you can find all timeout constants with `dir(Timeout)`. And yet, the amount of typing to reference the constant is identical (`Timeouts.free` vs. `TIMEOUTS_FREE`). Actually, it's less because `_` requires a chord (on most keyboards) versus `.` which is available with a single stroke.

After your feedback, I was tempted to roll back to the previous declarations, but after considering this more in depth, I now feel strongly that this code should be incorporated, not only for the benefits that it provides, but also to challenge the reader's assumptions and demonstrate a valuable paradigm.

cherrypy/test/test_states.py
jaraco: I realized that what the test was doing here was not to test the behavior of _check_port, but was simply asserting that the port was occupied by the server above. `portend.occupied` does exactly this.

cherrypy/process/servers.py
jaraco: And [in this question](https://www.youtube.com/watch?v=o9pEzgHorH0&feature=youtu.be&t=26m34s), Deiderich specifically endorses this approach as "useful."

webknjaz: Sounds reasonable :)

webknjaz (approved):
Great job ;)

jaraco:
Thanks @webknjaz for the detailed critique.

PR #1555, opened 2017-02-07
Updated sphinx docs to reflect the current state of cherrypy

wolph:

The package documentation for cherrypy is currently very out of date and a bit broken which results in pages like these: http://docs.cherrypy.org/en/latest/pkg/cherrypy.wsgiserver.html

The actual module doesn't exist anymore so it should be removed from the docs.

Empty documentation pages such as these: http://docs.cherrypy.org/en/latest/pkg/cherrypy.wsgiserver.html

Functioning docs, such as: http://cherrypy-wolph.readthedocs.io/en/latest/pkg/cherrypy.lib.html#module-cherrypy.lib.auth

* **Other information**:

The new version was generated using the Sphinx APIDoc command (similar to the previous version I imagine):

```
sphinx-apidoc -o docs/pkg cherrypy
```


docs/requirements.txt
```diff
@@ -1 +1,5 @@
 rst.linker
+mock
```
webknjaz, line 2: I'd change this list to smth like `-e ..[doc,testing]`

webknjaz (approved):

docs/requirements.txt
wolph: I wasn't aware of that feature. Good idea :)

webknjaz: Remove `mock` as well

CHANGES.rst
```diff
@@ -169,8 +169,8 @@ v8.0.0
 * Removed unicode filename from the package, working around
   pip #3894 and setuptools #704.
 
-7.1.0
------
+v7.1.0
```
jaraco, line 172: We don't want to change these headings unless we also update the tags in the repo, as rst.linker uses these headings to inject date stamps into the changelog. If you could write up a script that will remove the existing tags and recreate them with the 'v' in them, then we could include this change.

cherrypy/lib/encoding.py
```diff
@@ -321,9 +321,9 @@ def gzip(compress_level=5, mime_types=['text/html', 'text/plain'],
     values in the mime_types arg before calling this function.
 
     The provided list of mime-types must be of one of the following form:
-        * type/subtype
-        * type/*
-        * type/*+subtype
+        * `type/subtype`
```
jaraco, line 324: In my experience, the single backtick has special meaning in rst. I'll trust that you've tested that this renders properly, but if you haven't, please do confirm it does.

docs/conf.py
```diff
@@ -66,7 +69,7 @@ def try_import(mod_name):
 project, version, url, author = output_bytes.decode('utf-8').strip().split('\n')
 
 # General information about the project.
-copyright = '2001-2016 ' + author
+copyright = '2001-%d %s' % (datetime.date.today().year, author)
```
jaraco, line 72: Ooh. I like this. I'll want to adopt this change in [jaraco/skeleton](/jaraco/skeleton) from where the original code came, but fine to use this here for now.

docs/requirements.txt
jaraco: Yeah, the code is unclear about what's the appropriate way to declare dependencies, in part because it's straddling two techniques (one which specifies the dependencies in `docs/requirements.txt` and another where they're specified in `.[doc]`). I dislike the `.[doc]` technique mainly because it requires that you install `.` to get doc dependencies, which isn't always necessary or desirable. I like the `.[doc]` technique because it keeps the dependencies declared in a single location (setup.py). I should point out that the `-e .` line also runs in conflict with the RTD option to install the package in a virtualenv, though presumably pip resolves that conflict in some reasonable way. In any case, it's messy and unclear in the original code, so I won't demand that someone trying to help with docs reconcile the mess.

jaraco (approved):

CHANGES.rst
jaraco: Oh, look. This change actually fixed the mismatch. Thanks for that!

jaraco:
Thanks for this PR. I made a few commits subsequently to get builds running again: 77023a, 7dc687, and fae0613. Docs look better and substantially more useful. Well done.

docs/requirements.txt
webknjaz: Sounds reasonable. Of course I wanted it to be consistent. I'll try checking whether there's some way to make pip install just extras w/o the package itself. (there's a possibility of bypassing arguments to `setup.py`'s of individual packages, I would play with that)

webknjaz: Oh.. And shouldn't `mock` be replaced with `backports.mock`?

jaraco: Well, sort-of. backports.unittest_mock has a wonderful magic power when running under pytest, but has to be explicitly invoked in other environments such as a docs build. But! Since we're able to build the docs under Python 3, I just disabled support for building docs under Python 2 and then relied on unittest.mock natively. Problem solved.

cherrypy/lib/encoding.py
wolph: My limited tests indicated that it works better than the original version. And at the very least, it doesn't generate errors.

docs/requirements.txt
webknjaz: (nice)

webknjaz:
@jaraco btw, normally we are able to push into PR branches directly
JFYI

cherrypy/lib/encoding.py
webknjaz: Single backticks produce *italic text* and double backticks make up `inline code`.

wolph: Aww crud, did I mix up rst and md again. My bad

PR #1559, opened 2017-02-10
Keep Content-Type for encoded files

### LINKED ISSUE #1558: Wrong 'Content-Type' when retrieving gzip:ed js
While running CherryPy 10.0, I encounted an issue while trying to retrieve some gzipped javascript.
In order to save space on an embedded device, I'm only installing the `*.js.gz` files, not the unpacked `*.js`. After changing our html to directly request the `*.js.gz` files, the webapp stopped working.

CherryPy is configured to server those files by enabling `tools.staticdir.on`
The reason is that CherryPy sets 'Content-Type: application/x-gzip' when retrieving those files.
I believe that CherryPy should set:
```
Content-Type: application/json
Content-Encoding: gzip
```

darander:
Prior to this a file foo.js.gz got its Content-Type set to
application/x-gzip, after this change it'll instead have the
following two headers set:

Content-Type: application/javascript
Content-Encoding: gzip

Fixes #1558


cherrypy/lib/static.py
```diff
@@ -75,6 +75,13 @@ def serve_file(path, content_type=None, disposition=None, name=None,
         i = path.rfind('.')
         if i != -1:
             ext = path[i:].lower()
+            if ext in mimetypes.encodings_map.keys():
```
webknjaz, line 78: Could you please describe the logic you are adding in more detail? It should be possible to read and understand this code easily in the future.
(add it as a comment in code)

webknjaz (changes requested):
This change should definitely be covered with appropriate testcase

cherrypy/lib/static.py
darander: Sure, I could add a comment with something like:

> Set the content-type based on the real filename extension
> not the compression type. Also set the content-encoding
> for compressed files.

Would that suffice, or should it be reworded?

webknjaz: Well, looks good, but too abstract. What is the `real filename extension`? How is the encoding determined? What will happen if I call the file `test.js.42.gz`, for example? Add tests for corner cases please.

webknjaz: Also, I'm not sure, but probably this behavior must be configurable.
Let's ask @jaraco's opinion.

darander:
I've added a couple of checks, hopefully in the correct way. (Never used tox and pytest before).
The patch will only modify the current behaviour, if the uncompressed file has a known mimetype.

I skipped making the behaviour configurable, as I hope it should be fine with the extra check that was added. But I let @jaraco decide upon that.

darander:
Yeah, I guess there's a better way to solve it.

PR #1560, opened 2017-02-10
[wip] Allow the use of precompile files

### LINKED ISSUE #1558: Wrong 'Content-Type' when retrieving gzip:ed js

darander:
When serving static files and having tools.gzip.on=True, the compression
is done on the fly.

This patch allows the use of precompressed files for the accepted
mimetypes in encoding.py.

Further, add application/javascript and text/css to the accepted mimetypes.

Fixes #1558

There is currently no test added for this.
I've got a half test here, but before I add it, I'd like to ask if it's OK to add `tools.gzip.on = True` to the configuration in `test_static`? Or where should I put a test to verify this functionality?


darander:
Well, disregard this one for moment, as other tests are failing...

cherrypy/lib/encoding.py
```diff
@@ -412,7 +415,13 @@ def gzip(compress_level=5, mime_types=['text/html', 'text/plain'],
                 cherrypy.log('Gzipping', context='TOOLS.GZIP')
             # Return a generator that compresses the page
             response.headers['Content-Encoding'] = 'gzip'
-            response.body = compress(response.body, compress_level)
+            # Check if there exists a pre-compressed file, and use it.
+            file = cherrypy.serving.request.path_abs + '.gz'
+            if os.access(file, os.F_OK):
+                from cherrypy.lib.static import serve_file
```
webknjaz: This should definitely go to the beginning of the module.

cherrypy/lib/encoding.py
```diff
@@ -412,7 +415,13 @@ def gzip(compress_level=5, mime_types=['text/html', 'text/plain'],
                 cherrypy.log('Gzipping', context='TOOLS.GZIP')
             # Return a generator that compresses the page
             response.headers['Content-Encoding'] = 'gzip'
-            response.body = compress(response.body, compress_level)
+            # Check if there exists a pre-compressed file, and use it.
```
webknjaz: I've got concerns about such behavior being unobvious. I think such things must be explicitly configured.

cherrypy/lib/static.py
```diff
@@ -94,6 +94,7 @@ def serve_file(path, content_type=None, disposition=None, name=None,
     #   this way CP won't load the whole file in memory
     content_length = st.st_size
     fileobj = open(path, 'rb')
+    cherrypy.serving.request.path_abs = path
```
webknjaz: Why?

cherrypy/lib/encoding.py
darander: I'll have to check this later. I'm not completely following you, OTOH, I'm not at a computer for the moment.

cherrypy/lib/encoding.py
darander: I can see  your point. I'm likely look into a config option once I can all tests passing...

cherrypy/lib/static.py
darander: Simply the fact that I didn't find the absolute (local)  path anywhere. It's quite likely that I missed out somewhere...

darander:
See my comments inline. 
Add I said in my first comment, I currently don't expect this to be merged, as I jag missed a few test cases that failed... 

I haven't been able to understand the failures yet, and will look into them next week.

jaraco:
I'm concerned that the implicit behavior of matching a file as a "pre-compressed" file would be problemmatic. What if, for example, someone had foo.js and foo.js.gz, but foo.js.gz had malicious code and foo.js didn't. Then one would request foo.js and get malicious code and it wouldn't be obvious why the wrong file was being served. That sort of complication isn't appropriate for a general purpose tool.

I think I still would implement this as a separate tool. How about instead writing a wrapper or specialization of the gzip tool that does what you want. Maybe call it gzip_precomp or similar. Then, with this tool installed, `tools.gzip_precomp.on = True` would enable the feature for the resource and wouldn't affect existing users of tools.gzip... and maybe even others would want a tools.gzip_precomp. It could start out as a local, proprietary plugin, maybe be exposed as a third-party library, or even be incorporated into CherryPy as a standard tool if appropriate. That incremental adoption strategy would allow the code to be implemented and vetted without the burden of a rigorous test suite and review.

Now I haven't looked into the gzip tool enough to know if it has the hooks to do what you seek without having to copy a lot of the functionality. If you want to refactor the gzip tool to be more extensible, but without changing its behavior, I'd be very welcoming to such a change.

darander:
Sure, I'll look at creating another tool, that's likely the best  just as you say. 

I'll look at that next week. The first attempt might have quite some code duplication, I'll set what can be done about that in later revisions.

darander:
I've push a new RFC.

I hope it's a little bit closer to what you would expect,
I'm not really sure where the tests should be placed, for now, I've put them in `test_static.py`.
Also, currently only `tools.staticdir` is supported, not `tools.staticfile`.

cherrypy/test/test_static.py
```diff
@@ -160,6 +162,28 @@ def test_static(self):
         #   we just check the content
         self.assertMatchesBody('^Dummy stylesheet')
 
+    if __name__ == '__main__':
```
webknjaz: ?

darander: Hm, that shouldn't have been part of the commit...
Damn IDE adding stuff.

darander:
Disregard this one, I need to push another version fixing an issue with the erroneous line that was added...

darander:
This one should pass...
Unfortunately, I've had to remove the test where I asked for a file foo.html, and only foo.html.gz exists.
This seems to work in my own applcation, but not in the test case; though, I've got no idea why it looks like it's working in my own application...

darander:
Hm, I've also got another, much smaller patch. In that one, I'm runnign the tool as a before_request-tool, and re-writing the request.path_info, to add '.gz' if such a file exists.

Would something like that be acceptable?

jaraco:
Thanks for working on this. It's coming along nicely.

The fact that it duplicates a lot of code concerns me. Duplicated code is code that will diverge. A bug will get fixed in one copy but not the other, or a feature will be added to one but not the other. Any implementation should first look to extract the common functionality into a helper class or component functions such that the divergence in the implementation matches the essential differences in functionality (in this case loading of the content and compressing it). I'd be happy to accept a refactoring of the existing tool to expose the hooks necessary to customize the behavior to help facilitate your goals.

Moreover, I'm not convinced this functionality will have general purpose use, so probably is best for the custom behavior to be implemented in a third-party library or gist.

webknjaz:
@darander hello, should we expect any additional contribution on this?

PR #1568, opened 2017-02-26
CPWSGIServer.version contains CherryPy version + Cheroot version

mar10:
Update cherrypy/cheroot#5

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v10.1.2-0 / Unreleased
+-------
+
+* CPWSGIServer.version contains CherryPy version + Cheroot version
+
 v10.1.1
 -------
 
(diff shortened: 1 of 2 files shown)


mar10:
This is a proposal related to cherrypy/cheroot#12
As explained there, i currently have problems to get the tests running, and also could need a hint were would be a good place to add a test for this change.

webknjaz (commented):
Could you please increase min version of cheroot in `setup.py` to `v5.2.1`?

mar10:
Ok. I also created an issue for this, that I will reference: #1579 .
Btw.: do you think that bumping the patch number should generally be part of PR's?
PR authors won't know in advance if it get's accepted, in which order and on which target release it may land.

mar10:
ok, sorry, forget my 'btw'... I misunderstood

mar10:
I based this PR on the master branch. Let's close this mess and proceed with #1580 instead.
Sorry for the confusion.

PR #1569, opened 2017-02-26
Skip tests that require writing unicode filenames on incompatible filesystems.

### LINKED ISSUE #1567: StaticTest.test_unicode fails on codeanywhere
```
____________________________________________________________________________________________ StaticTest.test_unicode _____________________________________________________________________________________________
                                                                                                                                                                                                                  
self = <cherrypy.test.test_static.StaticTest testMethod=test_unicode>                                                                                                                                             
                                                                                                                                                                                                                  
    @pytest.mark.xfail(py27_on_windows, reason="#1544")                                                                                                                                                           
    def test_unicode(self):                                                                                                                                                                                       
>       with self.unicode_file():                                                                                                                                                                                 
                                                                                                                                                                                                                  
cherrypy/test/test_static.py:379:                                                                                                                                                                                 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python2.7/contextlib.py:17: in __enter__                                                                                                                                                                 
    return self.gen.next()                                                                                                                                                                                        
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
                                                                                                                                                                                                                  
    @staticmethod                                                                                                                                                                                                 
    @contextlib.contextmanager                                                                                                                                                                                    
    def unicode_file():                                                                                                                                                                                           
        filename = ntou('Слава Україні.html', 'utf-8')                                                                                                                                                            
        filepath = os.path.join(curdir, 'static', filename)                                                                                                                                                       
>       with io.open(filepath, 'w', encoding='utf-8') as strm:                                                                                                                                                    
E       UnicodeEncodeError: 'ascii' codec can't encode characters in position 43-47: ordinal not in range(128)                                                                                                    
                                                                                                                                                                                                                  
cherrypy/test/test_static.py:366: UnicodeEncodeError                                                                                                                                                              
===================================================================== 1 failed, 252 passed, 4 skipped, 2 xfailed, 5 xpassed in 80.13 seconds =====================================================================
ERROR: InvocationError: '/home/cabox/workspace/.tox/python/bin/pytest'                                                                                                                                            
____________________________________________________________________________________________________ summary _____________________________________________________________________________________________________
ERROR:   python: commands failed
```

Running tox on codeanywhere causes this test to fail.
* **I'm submitting a ...**
[x ] bug report
[ ] feature request

jaraco:
Fixes #1567

### DIFF
--- a/cherrypy/test/test_static.py
+++ b/cherrypy/test/test_static.py
@@ -17,6 +17,17 @@ from cherrypy._cpcompat import (
 from cherrypy.test import helper
 
 
+@pytest.fixture
+def unicode_filesystem(tmpdir):
+    filename = tmpdir / ntou('☃', 'utf-8')
+    tmpl = "File system encoding ({encoding}) cannot support unicode filenames"
+    msg = tmpl.format(encoding=sys.getfilesystemencoding())
+    try:
+        io.open(filename, 'w')
+    except UnicodeEncodeError:
+        pytest.skip(reason=msg)
+
+
 curdir = os.path.join(os.getcwd(), os.path.dirname(__file__))
 has_space_filepath = os.path.join(curdir, 'static', 'has space.html')
 bigfile_filepath = os.path.join(curdir, 'static', 'bigfile.log')
@@ -375,7 +386,7 @@ class StaticTest(helper.CPWebCase):
         sys.version_info < (3,)
     )
     @pytest.mark.xfail(py27_on_windows, reason="#1544")
-    def test_unicode(self):
+    def test_unicode(self, unicode_filesystem):
         with self.unicode_file():
             url = ntou('/static/Слава Україні.html', 'utf-8')
             # quote function requires str
--- a/cherrypy/test/test_static.py
+++ b/cherrypy/test/test_static.py
@@ -4,10 +4,13 @@ import io
 import os
 import sys
 import platform
+import tempfile
 
+import six
 from six.moves import urllib
 
 import pytest
+import py.path
 
 import cherrypy
 from cherrypy.lib import static
@@ -23,9 +26,9 @@ def unicode_filesystem(tmpdir):
     tmpl = "File system encoding ({encoding}) cannot support unicode filenames"
     msg = tmpl.format(encoding=sys.getfilesystemencoding())
     try:
-        io.open(filename, 'w')
+        io.open(six.text_type(filename), 'w')
     except UnicodeEncodeError:
-        pytest.skip(reason=msg)
+        pytest.skip(msg)
 
 
 curdir = os.path.join(os.getcwd(), os.path.dirname(__file__))
@@ -386,7 +389,8 @@ class StaticTest(helper.CPWebCase):
         sys.version_info < (3,)
     )
     @pytest.mark.xfail(py27_on_windows, reason="#1544")
-    def test_unicode(self, unicode_filesystem):
+    def test_unicode(self):
+        self.ensure_unicode_filesystem()
         with self.unicode_file():
             url = ntou('/static/Слава Україні.html', 'utf-8')
             # quote function requires str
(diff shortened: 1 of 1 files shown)


cherrypy/test/test_static.py
```diff
@@ -4,10 +4,13 @@
 import os
 import sys
 import platform
+import tempfile
 
+import six
```
webknjaz: Nitpick: I think you could follow the same import syntax as in the line below. It would look nicer.

cherrypy/test/test_static.py
```diff
@@ -17,6 +20,17 @@
 from cherrypy.test import helper
 
 
+@pytest.fixture
+def unicode_filesystem(tmpdir):
+    filename = tmpdir / ntou('☃', 'utf-8')
+    tmpl = "File system encoding ({encoding}) cannot support unicode filenames"
+    msg = tmpl.format(encoding=sys.getfilesystemencoding())
+    try:
+        io.open(six.text_type(filename), 'w')
```
webknjaz: I'm not sure, but doesn't it leave an open file descriptor in memory?
Perhaps `io.open(..).close()` would work better.
Or it could be:
```python
with io.open(..):
    pass
```

cherrypy/test/test_static.py
```diff
@@ -376,6 +390,7 @@ def unicode_file():
     )
     @pytest.mark.xfail(py27_on_windows, reason="#1544")
     def test_unicode(self):
+        self.ensure_unicode_filesystem()
```
webknjaz: This should probably go to the `unicode_file()` method, so in case of reusing it would always be checked automatically.

cherrypy/test/test_static.py
```diff
@@ -385,6 +400,17 @@ def test_unicode(self):
 
             expected = ntou('Героям Слава!', 'utf-8')
             self.assertInBody(expected)
+    
+    def ensure_unicode_filesystem(self):
```
webknjaz: This could be a `@staticmethod`

cherrypy/test/test_static.py
jaraco: Good idea.

cherrypy/test/test_static.py
jaraco: Yeah, I'm slightly tempted not to use io at all and simply test encoding of the filename with `getfilesystemencoding`. As it is, that file handle will never get assigned and will get garbage-collected almost immediately, but you're right that it's not explicitly closed. Perhaps it is worthwhile to close it explicitly. I'll do that.

cherrypy/test/test_static.py
jaraco: Based on this and the previous comment, I made the function a global function and put it in the proximity of `unicode_file`.

cherrypy/test/test_static.py
```diff
@@ -4,10 +4,13 @@
 import os
 import sys
 import platform
+import tempfile
 
+from six import text_type as str
```
webknjaz, line 9: I bet `flake8` checker would tell you to not shadow built-ins

webknjaz (approved):

PR #1578, opened 2017-03-11
Encode redirect response

### LINKED ISSUE #1577: Encode URLs in redirect responses
* **I'm submitting a ...**
[x] bug report
[ ] feature request

* **What is the current behavior?**
The standard response page for 30x errors contains the target URL as `href` as well as clear text.

* **What is the expected behavior?**
Even though browsers *mostly* will not display the result, but redirect instead, you never know:
Any input from unknown sources should be html-escaped.

* **Please tell us about your environment:**
CherryPy 3.8 until 10.1

mar10:
Fix #1577

### DIFF
--- a/cherrypy/_cperror.py
+++ b/cherrypy/_cperror.py
@@ -271,7 +271,7 @@ class HTTPRedirect(CherryPyException):
                 307: 'This resource has moved temporarily to ',
             }[status]
             msg += '<a href=%s>%s</a>.'
-            msgs = [msg % (saxutils.quoteattr(u), u) for u in self.urls]
+            msgs = [msg % (saxutils.quoteattr(u), escape_html(u)) for u in self.urls]
             response.body = ntob('<br />\n'.join(msgs), 'utf-8')
             # Previous code may have set C-L, so we have to reset it
             # (allow finalize to set it).
--- a/cherrypy/test/test_core.py
+++ b/cherrypy/test/test_core.py
@@ -150,6 +150,9 @@ class CoreRequestHandlingTest(helper.CPWebCase):
             def url_with_quote(self):
                 raise cherrypy.HTTPRedirect("/some\"url/that'we/want")
 
+            def url_with_xss(self):
+                raise cherrypy.HTTPRedirect("/some<script>alert(1);</script>url/that'we/want")
+
             def url_with_unicode(self):
                 raise cherrypy.HTTPRedirect(ntou('тест', 'utf-8'))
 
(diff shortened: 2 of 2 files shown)


webknjaz (commented):
Could you please add some testing of this case?

cherrypy/_cperror.py
```diff
@@ -271,7 +271,7 @@ def set_response(self):
                 307: 'This resource has moved temporarily to ',
             }[status]
             msg += '<a href=%s>%s</a>.'
-            msgs = [msg % (saxutils.quoteattr(u), u) for u in self.urls]
+            msgs = [msg % (saxutils.quoteattr(u), escape_html(u)) for u in self.urls]
```
webknjaz, line 274: Why not change `saxutils.quoteattr(u)` as well? Any objections?

mar10: escape_html seems to be enough to prevent XSS, and we use it for clear text inside the `<a>` tag, not for an attribute (as quoteattr would suggest), it's also faster.
(But I guess quoteattr would work as well.)

mar10:
Concerning the tests, it would go something like 
```py
    class RedirectApp(object):
        def index(self, **kwargs):
            raise cherrypy.HTTPRedirect("example.com?redir=<script>alert(1);</script>")
        index.exposed = True
    
    cherrypy.quickstart(RedirectApp())
```

and

```py
    import requests
    res = requests.get("http://127.0.0.1:8080?redir=<abc>", allow_redirects=False)
    assert res.status_code == 303
    assert '<' not in res.content
```

But I don't know where and how to add it.

cherrypy/_cperror.py
webknjaz: I just think that using several different libs for doing the same thing. Moreover, current implementation does conversion of the **exactly same** variable during each loop iteration. It's much more optimal to do this one time.

webknjaz: Oh.. I misunderstood the purpose of both conversions (and that they are different)

webknjaz:
Regarding tests, please check [`cherrypy/test/test_request_obj.py`](https://github.com/cherrypy/cherrypy/blob/master/cherrypy/test/test_request_obj.py), I don't see tests for 3xx status codes, but it should be relatively easy to add them there :)

P.S. Please note `getPage()`, `assertStatus()` and `assertBody()` helpers.

mar10:
Looking at the code I would not consider this change harmful and we already invested quite a bit  of time on this one-liner ;)
But I'll see what I can do (may take a while until I find time though).

webknjaz:
@mar10 I've got a minute and found a place to inject a test. Merged :)

mar10:
Cool, thank you!

Issue #1579, opened 2017-03-11
Version number should mention CherryPy and Cheroot


mar10:
* **I'm submitting a ...**
[ ] bug report
[x] feature request
[ ] question about the decisions made in the repository

* **What is the current behavior?**
Currently the server version is reported like 'CherryPy/10.1'
The version number was taken from the Server object, which has moved to the Cheroot project recently (which has its own version numbers).
cherrypy/cheroot#5 changed this, so the Cheroot server now reports its version number correctly.

I propose that CherryPy reports its version number now like 'CherryPy/10.1 Cheroot/5.1'.

webknjaz:
Thank you for your contribution!

PR #1580, opened 2017-03-11
Report composite version 

mar10:
Update #1579

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v10.1.2-0 / Unreleased
+-------
+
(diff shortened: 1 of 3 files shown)


mar10:
This seems to break because of

> No matching distribution found for cheroot>=5.2.1 (from CherryPy==10.1.2.dev11+ng6421ba1)

Or did I get your request wrong:

> Could you please increase min version of cheroot in setup.py to v5.2.1?

webknjaz:
Sorry, I missed that `5.2.1` isn't yet released. Please use `5.2.0`

webknjaz (approved):

PR #1589, opened 2017-04-14
Remove secret credentials

notWhaleB:
It seems that this credentials shouldn't be in the public access.

* **Other information**:


webknjaz:
[They are encrypted.](https://docs.travis-ci.com/user/encryption-keys/)

PR #1596, opened 2017-05-17
Fix #1595 Fix path normlisation

### LINKED ISSUE #1595: `cherrypy.url` fails to normalize path
This call to `cherrypy.url` fails with `IndexError`:
```
>>> cherrypy.url(qs='../../../../../../etc/passwd')
...
IndexError: pop from empty list
```
The culprit seems in this logic, which strips `newurl` of as many `atoms` as there are `..`: 
https://github.com/cherrypy/cherrypy/blob/master/cherrypy/_helper.py#L261,L271

There are various problems.

- That logic should only applied to the "path" part of `newurl`, not to the full url.

- As a consequence of the point above, `..` in the query string `qs` should not be considered

- To consider: redundant `..` should be ignored, to mimic `os.path.normpath`:

```
>>> os.path.normpath('/etc/../../../usr')
'/usr'
```

lbolla:
* Fix path normalization
* See issue #1595

### DIFF
--- a/cherrypy/_helper.py
+++ b/cherrypy/_helper.py
@@ -223,6 +223,30 @@ def url(path='', qs='', script_name=None, base=None, relative=None):
     if qs:
         qs = '?' + qs
 
+    def normalize_path(path):
+        if './' not in path:
+            return path
+
+        # Normalize the URL by removing ./ and ../
+        atoms = []
+        for atom in path.split('/'):
+            if atom == '.':
+                pass
+            elif atom == '..':
+                # Don't pop from empty list
+                # (i.e. ignore redundant '..')
+                if atoms:
+                    atoms.pop()
+            elif atom:
+                atoms.append(atom)
+
+        newpath = '/'.join(atoms)
+        # Preserve leading '/'
+        if path.startswith('/'):
+            newpath = '/' + newpath
+
+        return newpath
+
     if cherrypy.request.app:
         if not path.startswith('/'):
             # Append/remove trailing slash from path_info as needed
@@ -246,7 +270,7 @@ def url(path='', qs='', script_name=None, base=None, relative=None):
         if base is None:
             base = cherrypy.request.base
 
-        newurl = base + script_name + path + qs
+        newurl = base + script_name + normalize_path(path) + qs
     else:
         # No request.app (we're being called outside a request).
         # We'll have to guess the base from server.* attributes.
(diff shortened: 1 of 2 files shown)


jaraco:
Brilliant! Thanks.

jaraco:
[Released with v10.2.2](http://cherrypy.readthedocs.io/en/latest/history.html#v10-2-2).

webknjaz (commented):
(we could consider using `pathlib` and its packport `pathlib2`)

jaraco:
@webknjaz Does pathlib implement this behavior? If so, then I'm +1. If it only does half the job, then it's hardly worth the complexity.

webknjaz:
@jaraco 
I've found this on the Internet: http://stackoverflow.com/a/41375442/595220, but didn't check it yet.

Also, there's `os.path.normpath`, which looks working well:
```python
$ ipython

In [1]: import os.path

In [2]: os.path.normpath('../')
Out[2]: '..'

In [3]: os.path.normpath('../../')
Out[3]: '../..'

In [4]: os.path.normpath('./../../')
Out[4]: '../..'

In [5]: os.path.normpath('./../../x/y/z')
Out[5]: '../../x/y/z'

In [6]: os.path.normpath('/x/./../../x/y/z')
Out[6]: '/x/y/z'

In [7]: os.path.normpath('/other/../../../../../page1')
Out[7]: '/page1'
```

(tested under Python 2.7)

lbolla:
I experimented with `normpath`, but the problem is that, on Windows, it converts forward to backward slashes. I also opted for the minimal change (above) to avoid breaking existing behaviour, in case `normpath` has other corner cases that are currently untested by CherryPy.

webknjaz:
Fair enough. Will have to check `pathlib` more carefully, it's got OS-specific classes.

jaraco:
> on Windows it converts to backslashes

In that case, you _could_ use `posixpath` for consistent behavior across platforms.

But you're right about corner cases. Even using normpath, I ran found that the test suite fails without these considerations:

```
def normalize_path(path):
    if path == './':
        return ''
    normed = posixpath.normpath(path)
    if path.endswith('/') and not normed.endswith('/'):
        normed += '/'
    return normed
```

And you're right, while that's more concise, it's more complicated because it involves two (competing) implementations, and it still may not be compatible in untested cases.

PR #1599, opened 2017-06-04
Enable travis CI pipelines

### LINKED ISSUE #1594: [TODO] Change Travis CI config to use newly presented build stages feature
Refs:
* https://blog.travis-ci.com/2017-05-11-introducing-build-stages
* https://docs.travis-ci.com/user/build-stages#Examples

*Note: I'm going to move PYPI deployment to separate stage, so it would wait for tests in all envs to complete.*

webknjaz:
CI improvement

Closes #1594

  - deploy is run under Python 3.6 and may succeed while other envs fail
  - linters and less important env tests share the same job queue as important ones

Enable Travis CI use pipelines in order to:
  - fail fast
  - run linters and light-weight env tests against latest Python versions first
  - run deploy stage only if the rest of test jobs completed successfully

* **Other information**:
OS X tests seem unstable currently and it takes lots of time for them to start and finish. Also, to run Python tests under OS X some hacks needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ _base_envs:
 - &linux_base
   os: linux
   dist: trusty
-  sudo: required
+  sudo: false
   addons:
     apt:
       packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,13 @@ _base_envs:
   os: linux
   dist: trusty
   sudo: false
+- &pyenv_base
+  <<: *stage_test
+  python: pypy
+  env:
+  - PYTHON_VERSION=pypy2-5.7.1
+  - &env_pyenv PYENV_ROOT="$HOME/.pyenv"
+  - &env_path PATH="$PYENV_ROOT/bin:$PATH"
   addons:
     apt:
       packages:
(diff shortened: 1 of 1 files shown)


webknjaz:
@jaraco I would like to hear your opinion about the test job *Python 3.3 @ OS X*, which looks failing more often than others.

jaraco:
Yeah, that's the problem with blocking releases on passing all tests - if you have any intermittent failures, they cause problems... and CherryPy has lots of intermittent failures. I'm not sure the test suite is robust enough to accept this sort of change.

I'd be okay with xfailing the tests along with filing a ticket for each marker... but I don't know if/when anyone will find the time to make the tests more robust. If the issue is just on macOS with Python 3.3, I'd be okay with excluding that from the matrix.

webknjaz:
Sometimes the tests fail under other OS X envs, 3.3 just looks more likely to fail. I think, I'll need to investigate failing tests and use xfails in the end.

jaraco:
Do we know what happens if a blocking test fails, causing the process to stop at that stage, but then someone manually re-runs the test that failed? Does Travis then re-evaluate the stage logic and run subsequent stages, or does instead the stage logic only ever rely on the results of the first run?

webknjaz:
No, Travis doesn't re-run the next stage automatically if one restarts failed job of previous stage and it succeeds. One should also hit restart buttons of all jobs in subsequent stages.

Actually, I recall there's some script called `travis_retry`, which just re-runs command in a loop, but I didn' t test it. We could try use it.

jaraco (approved):

Issue #1498, opened 2016-09-09
Restricting File Upload?


EternityForest:
It appears that when someone uploads a file, it gets placed on disk immediately, before any user code is called.

Is this correct? And if so, is there a security issue due to the possibility of someone repeatedly uploading large files and using up disk bandwidth, or even possibly destroying an SD card by exceeding write endurance?

Would it be practical to add some sort of file filter such that any attempted uploads would be filtered through a function, which could decide what to do with the raw stream and either return a file handle, or some higher level object created from the stream, or just reject it entirely?

Thanks for your time!

webknjaz:
File is not being saved to disk unless you do this explicitly. In your request handler you may access the `request.body` readable file-like object and then read this stream, process and write anywhere you want.

Please check these modules for more info regarding your question and request processing implementation:
https://github.com/cherrypy/cherrypy/blob/master/cherrypy/_cprequest.py#L323-L329
https://github.com/cherrypy/cherrypy/blob/master/cherrypy/_cpreqbody.py#L362-L363

EternityForest:
So I was doing some testing with an unrelated bug in some other code and noticed that uploading files causes a file like /tmp/#4460225 to be created and immediately deleted, so something does appear to be written to disk.

Also, documentation says "When a client uploads a file to a CherryPy application, it's placed
on disk immediately."

Is the documentation out of date?  Do I have a config option set that I don't know about?

webknjaz:
Could you please provide a source code snippet with an app example?

EternityForest:
Allright here's a minimal example for linux, I don't have a windows machine handy.

```python
import cherrypy,html
class HelloWorld(object):
    def index(self,file=None):
        return 'File:'+html.escape(repr(file))+"""<form method="POST" enctype="multipart/form-data">
                <input name="file" type="file"></input><input type="submit"></form>"""
    index.exposed = True

cherrypy.quickstart(HelloWorld())
```

Then in bash(I don't know how to test this on windows):
`sudo fatrace|grep python`

Every time a large file is uploaded, I get something like this which appears to be coming from the cherrypy process:

```
python3(9793): W /tmp/#4460211 (deleted)
python3(9793): W /tmp/#4460211 (deleted)
python3(9793): W /tmp/#4460211 (deleted)
python3(9793): W /tmp/#4460211 (deleted)
python3(9793): W /tmp/#4460211 (deleted)
python3(9793): W /tmp/#4460211 (deleted)
[...]

PR #1603, opened 2017-06-23
Add dynamic upload limits, configurable maxrambytes

EternityForest:
This pull request enables user code to dynamically accept or reject file uploads and large HTTP request parts with a callback function, before those files actually get written to disk.

It also allows configuring the threshold that determines if a given item is written to disk as a tempfile or just kept in RAM.

#1498 

The temporary file threshold is set at 1000 bytes and​ cannot be changed in any documented way that I know of.

As far as I can tell, by default anyone can upload a large file which will be placed into a tempfile. As this happens before your application code, there's not much to be done to prevent unauthorized users from creating large tempfiles, using RAM or SSD write cycles.

The request.body.maxbytes option may now be a tuple of (default, function). Up to default bytes will be read, after which function will be called with no arguments and whatever it returns will be used as the upload limit for that request. This allows your code to inspect HTTP headers to decide what to do.

The reason for the use of a default is to avoid calling the function in cases where it is a somewhat expensive operation, such as if it involves a call to a users database.

* **Other information**:
I believe this to be somewhat important for embedded systems, where RAM is limited and /tmp is often kept in a tmpfs to prevent wearing out the flash memory. In these situations it would be easy for an attacker to continually write to tmpfs until it was full, potentially using up all available tempfile space.

In addition, systems that do not use a tmpfs for /tmp could simply run out of disk space on small SD cards, or even wind up with totally destroyed flash on some very cheap cards without wear leveling.


cherrypy/_cprequest.py
```diff
@@ -643,7 +643,7 @@ def respond(self, path_info):
 
                     self.body = _cpreqbody.RequestBody(
                         self.rfile, self.headers, request_params=self.params)
-
+
```
webknjaz: Changing this file is absolutely unnecessary

docs/advanced.rst
```diff
@@ -447,7 +447,8 @@ and wrap your entire CherryPy application with it:
    def secureheaders():
        headers = cherrypy.response.headers
        headers['X-Frame-Options'] = 'DENY'
-       headers['X-XSS-Protection'] = '1; mode=block'
+       headers['        self.assertStatus(413)
```
webknjaz: This looks accidentally changed

webknjaz (changes requested):
There's a bunch of linters failed: https://travis-ci.org/cherrypy/cherrypy/jobs/246086738

You may use `pre-commit` tool to check it locally.
Running the following must help you:
```shell
$ pre-commit run --all-files
```

cherrypy/_cpreqbody.py
```diff
@@ -724,14 +738,22 @@ def default_proc(self):
...
 
         Return fp_out.
         """
-        if fp_out is None:
-            fp_out = self.make_file()
-        self.read_lines_to_boundary(fp_out=fp_out)
-        return fp_out
+        if not fp_out:
+            x = self.read_lines_to_boundary()
+            if isinstance(x,(str,bytes)):
```
webknjaz: I'm not sure whether this block would work under both Python 2 and Python 3. There's [`_cpcompat` module having helper functions](https://github.com/cherrypy/cherrypy/blob/1639205/cherrypy/_cpcompat.py#L27-L91) for dealing with different string/byte representations.

cherrypy/_cpreqbody.py
```diff
@@ -724,14 +738,22 @@ def default_proc(self):
...
+        a BytesIO object, based on maxrambytes.
 
         Return fp_out.
         """
-        if fp_out is None:
-            fp_out = self.make_file()
-        self.read_lines_to_boundary(fp_out=fp_out)
-        return fp_out
+        if not fp_out:
+            x = self.read_lines_to_boundary()
```
webknjaz, line 747: You can do this call once before the loop with `fp_out`.

cherrypy/_cpreqbody.py
```diff
@@ -802,7 +843,13 @@ def read(self, size=None, fp_out=None):
             # Check lengths.
             self.bytes_read += datalen
             if self.maxbytes and self.bytes_read > self.maxbytes:
-                raise cherrypy.HTTPError(413)
+                if self.maxbytes_extend:
+                    print("extending maxbytes")
```
webknjaz: This statement is not needed. If you want to log smth, use logging

cherrypy/_cpreqbody.py
```diff
@@ -802,7 +843,13 @@ def read(self, size=None, fp_out=None):
...
+                    self.maxbytes = self.maxbytes_extend()
+                    #We only want to call maxbytes extend once per request I think
+                    self.maxbytes_extend =None
+                if self.maxbytes and self.bytes_read > self.maxbytes:
```
webknjaz: Do `bytes_read` and `maxbytes` mutate since previous check? If not, this conditional doesn't make any sense, because it's always `True` at this point.

cherrypy/_cpreqbody.py
EternityForest: Seems to work on both in the interactive shell, bytes appears to be an alias of string in python. There's probably a better way, but this seemed like a decent way to express "Is this value object and not a file like object?"

webknjaz: `str` is alias of unicode type in Python 3 and bytes in Python 2. I'd go for try/except following [EAFP principle](https://docs.python.org/3/glossary.html).

cherrypy/_cpreqbody.py
EternityForest: Uh oh. That was a debug print statement that I was using(So I could do a search for print statements and delete it easier than searching through the log calls).

Looks like I forgot the last step.

cherrypy/_cpreqbody.py
webknjaz: You could also use `isinstance(x, file)`

EternityForest: So just check if the BytesIO constructor fails due to getting a file-like object? I like the explicitness of checking for the specific type, but it does kind of make it look like x might be an actual unicode string if you don't know the python2 history.

Either way there should probably be a comment there I'd say.

webknjaz: Two options are:
1) `isinstance(x, file)`
2) try/`except TypeError`

EternityForest:
What's the policy for amending commits here? Amend, squash, add a new commit?

webknjaz:
@EternityForest just maintain a clean history. Add commits if needed. You may want to do a little history cleanup in the end of review. Anyway GitHub has 3 ways of merging and if we want to squash, we'll just hit appropriate button in PR :)

cherrypy/_cpreqbody.py
```diff
@@ -582,6 +581,7 @@ def process(self):
             proc(self)
 
     def default_proc(self):
+
```
webknjaz, line 584: Why did you add this line?

cherrypy/_cpreqbody.py
```diff
@@ -724,14 +736,22 @@ def default_proc(self):
...
-            fp_out = self.make_file()
-        self.read_lines_to_boundary(fp_out=fp_out)
-        return fp_out
+        if not fp_out:
+            x = self.read_lines_to_boundary()
+            try:
+                return io.BytesIO(x)
+            except TypeError:
+                return x
+        else:
```
webknjaz, line 752: I'd remove `else`-block to reduce code nesting

cherrypy/_cpreqbody.py
webknjaz: I still think this function should be called once before the `if`-block

webknjaz: Also, please use meaningful variable name instead of `x`.

cherrypy/_cpreqbody.py
webknjaz: Try doing as little unrelated changes as possible

cherrypy/_cpreqbody.py
```diff
@@ -724,14 +736,22 @@ def default_proc(self):
                 self.file = result
 
     def read_into_file(self, fp_out=None):
-        """Read the request body into fp_out (or make_file() if None).
+        """Read the request body into fp_out.
+
+        if fp_out not given, use either make_file or
```
webknjaz, line 741: Please start sentence with capital letter

cherrypy/_cpreqbody.py
```diff
@@ -214,6 +215,9 @@ def process_multipart(entity):
     # Read all parts
     while True:
         part = entity.part_class.from_fp(entity.fp, ib)
+        # Inherit maxrambytes from the parent entity.
```
webknjaz, line 218: Please move this to `Part.from_fp` classmethod

cherrypy/_cpreqbody.py
```diff
@@ -745,7 +765,13 @@ def __init__(self, fp, length, maxbytes, bufsize=DEFAULT_BUFFER_SIZE,
         # Wrap our fp in a buffer so peek() works
         self.fp = fp
         self.length = length
-        self.maxbytes = maxbytes
+        if isinstance(maxbytes, tuple):
```
webknjaz, line 768: This should work for any iterable, I think. Not just tuples.

cherrypy/_cpreqbody.py
```diff
@@ -832,8 +876,12 @@ def read(self, size=None, fp_out=None):
             # Check lengths.
             self.bytes_read += datalen
             if self.maxbytes and self.bytes_read > self.maxbytes:
-                raise cherrypy.HTTPError(413)
-
+                if self.maxbytes_extend:
```
webknjaz, line 879: This block of code duplicates code above. It must be moved to separate private/protected method

webknjaz (changes requested):
Please also add docstrings to functions you added

cherrypy/_cpreqbody.py
webknjaz, line 768: After thinking this through, I recall that there's some places in config, where we allow setting callable instead of scalar as a value. We should probably allow setting `maxbytes` to callable and test it like:
```python
try:
    self.maxbytes = maxbytes()
except TypeError:
    self.maxbytes = maxbytes
```

webknjaz: @jaraco any additional thoughts?

webknjaz:
Closing this, since there's three solutions provided within #1498 + this one is incomplete and I don't have time to work on it myself.

PR #1604, opened 2017-06-28
Fix missing import

mar10:
`text_or_bytes` is never defined, so a NameError is always raised. 
So it is always set to `str`, which causes a problem in functions like this, where a `unicode` is not recognized as string:
```py
    def update(self, config):
        """Update self from a dict, file or filename."""
        if isinstance(config, text_or_bytes):
            # Filename
            config = Parser().dict_from_file(config)
        elif hasattr(config, 'read'):
            # Open file object
            config = Parser().dict_from_file(config)
        else:
            config = config.copy()
        self._apply(config)
```

### DIFF
--- a/cherrypy/lib/reprconf.py
+++ b/cherrypy/lib/reprconf.py
@@ -18,6 +18,8 @@ by adding a named handler to Config.namespaces. The name can be any string,
 and the handler must be either a callable or a context manager.
 """
 
+from cherrypy._cpcompat import text_or_bytes
+
 try:
     # Python 3.0+
     from configparser import ConfigParser


webknjaz (approved):
Could you please also remove try/except block? (Line 27)

jaraco:
I think this should be `six.string_types`. Much of the compat module is obviated by `six`.

webknjaz:
@jaraco are you sure? it appears you were replacing `basestring` with `text_or_bytes = six.text_type, six.binary_type` in this commit d3c9807

jaraco:
You're right. I'm not sure. In most code, `string_types` is the right thing, but in CherryPy, often bytes are used even on Python 3, so I may well be mistaken.

webknjaz:
I'm accepting this as is. We can always change this if we find out better way.

PR #1605, opened 2017-06-28
Windows causes timeout during port check

corylutton:
During startup port checking always results in a timeout on Windows. This just skips the check if the system is a windows system.

Currently it checks until timeout and adds time to startup unnecessarily.

### DIFF
--- a/cherrypy/process/servers.py
+++ b/cherrypy/process/servers.py
@@ -250,7 +250,7 @@ class ServerAdapter(object):
         if not os.environ.get('LISTEN_PID', None):
             # Wait for port to be occupied if not running via socket-activation
             # (for socket-activation the port will be managed by systemd )
-            if isinstance(self.bind_addr, tuple):
+            if isinstance(self.bind_addr, tuple) and os.name != 'nt':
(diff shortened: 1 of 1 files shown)


jaraco:
Hmm. Amputating a feature because it's broken on a platform should be the last resort. I've [added portend to appveyor](https://ci.appveyor.com/project/jaraco/portend/) to test for and capture failures on the underlying behavior, which has resulted in [this failing build](https://ci.appveyor.com/project/jaraco/portend/build/job/en9er8k79wjwh4a7). Perhaps [this recent removal](https://github.com/jaraco/portend/commit/3bc7936f077487f2145c2f59a3309b0242607f23) is implicated in the failure. Regardless, CherryPy seeks to support Windows fully in this regard, so the issue should be corrected upstream in portend.

jaraco:
@corylutton: [portend 2.1.1](https://pypi.org/project/portend/) is now released with what I think will fix the issue. Can you upgrade portend in your environment and confirm?

corylutton:
@jaraco I updated and it seems to work correctly, thanks for your effort in resolving.

PR #1611, opened 2017-07-12
Expose HTTPRedirect default status

jaraco:
While adapting some internal code, I had a situation where I wanted to handle the response directly rather than raising an HTTPRedirect, but I still wanted to use the same response code. After inspecting the logic for how HTTPRedirect calculates the appropriate response code for a redirect, I realized I wanted to re-use this logic rather than copy/paste it.

This change extracts that logic into the `HTTPRedirect.default_status` class property, such that it can be used by a third party application even if HTTPRedirect itself isn't used.

I'm a little sad this change adds a new dependency for the classproperty, but since it's not stdlib, that's the best option.


cherrypy/_cperror.py
```diff
@@ -228,22 +228,24 @@ def __init__(self, urls, status=None, encoding=None):
...
-        if status is None:
-            if request.protocol >= (1, 1):
-                status = 303
-            else:
-                status = 302
-        else:
-            status = int(status)
-            if status < 300 or status > 399:
-                raise ValueError('status must be between 300 and 399.')
+        status = int(status) if status is not None else self.default_status
```
webknjaz: Why not assign `self.status` right here?

cherrypy/_cperror.py
```diff
@@ -228,22 +228,24 @@ def __init__(self, urls, status=None, encoding=None):
...
+        if not 300 <= status <= 399:
+            raise ValueError('status must be between 300 and 399.')
 
         self.status = status
         CherryPyException.__init__(self, abs_urls, status)
```
webknjaz: Can we use `super()`?

cherrypy/_cperror.py
```diff
@@ -123,6 +123,9 @@ class Root:
 import six
 from six.moves import urllib
 
+from jaraco.classes.properties import classproperty
```
webknjaz, line 128: I'd move it before `six` preserving alphabetic order

webknjaz (approved):

Safihre:
Another dependency.. And also a non standard language concept of class property. 
Is there really no better and more general solution?

cherrypy/_cperror.py
jaraco: I've been using a convention of creating four sections for imports: stdlib, compatibility (stdlib functionality but from third-party modules), third-party modules, and finally intrapackage imports. Perhaps my convention diverges slightly from the more common convention of putting all third-party imports (even six) in the same group.

cherrypy/_cperror.py
jaraco: Yes, I think that would be preferable.

cherrypy/_cperror.py
jaraco: Yes, perhaps, but that's outside the scope of this PR.

cherrypy/_cperror.py
webknjaz: No, it's okay. It's good that you explained your way of structuring imports. I think makes sense, we should probably document this approach in contribution docs to make everyone follow the same code style.

cherrypy/_cperror.py
```diff
@@ -228,22 +228,24 @@ def __init__(self, urls, status=None, encoding=None):
...
-                raise ValueError('status must be between 300 and 399.')
+        status = int(status) if status is not None else self.default_status
+        if not 300 <= status <= 399:
+            raise ValueError('status must be between 300 and 399.')
 
         self.status = status
         CherryPyException.__init__(self, abs_urls, status)
 
+    @classproperty
+    def default_status(cls):
```
jaraco: codacy fail - this is a classmethod, so the convention is `cls`.

jaraco: codacy fail - it can't be a static method because it's a class property.

jaraco:
> Another dependency.. And also a non standard language concept of class property.
Is there really no better and more general solution?

It is a non-standard concept; i.e. it's not in the stdlib. I personally think it should be as it's a fairly basic concept in OO, even if only occasionally used.

Yes, the additional dependency makes me a little sad, but not as sad as a copy/paste of the implementation (and subsequent maintenance burden). That is, if we assume a classproperty is desired.

I believe a classproperty is desired. It would be trivial to make that `default_status` a static method, such that the usage is `HTTPRedirect.default_status()` or `HTTPRedirect.get_default_status()`, but that feels a little too verbose to me. Of course, such a method would have the advantage of communicating to the user that the value is calculated (and might change based on context).

@Safihre : Classproperty details aside, what interface do you think is best for the end user who wishes to resolve the default status for a redirect?

Safihre:
>  It would be trivial to make that default_status a static method, such that the usage is HTTPRedirect.default_status() or HTTPRedirect.get_default_status(), but that feels a little too verbose to me. Of course, such a method would have the advantage of communicating to the user that the value is calculated (and might change based on context).

What is so wrong with that? 
`get_default_status()` seems pretty reasonable and understandable if an external third party application wants to get this info: it's a getter/setter-style.

webknjaz:
> it's a getter/setter-style

It's normally applied to instances, not classes. It's pythonic way to use `@property` decorator, but it's not applicable to properties at the class level. I support @jaraco's intention to bring a bit more elegance into the codebase.

PR #1612, opened 2017-07-12
Added note to docs about sessions when more than one app is on a single domain.

mmulqueen:

Updates documentation on sessions.

N/A - No bug report.

Currently documentation does not mention the existence of the config property tools.sessions.name. This is a useful tool if two separate applications are in the same cookie context.

Documention now mentions the config property and how it can help in the above case.

* **Other information**:

See also my answer on StackOverflow: https://stackoverflow.com/questions/2597024/cherrypy-sessions-for-same-domain-different-port/45065890#45065890


webknjaz (approved):

webknjaz:
@mmulqueen thank you for your contribution!

PR #1620, opened 2017-08-15
Fix gzip and caching tools integration (#1190)

### LINKED ISSUE #1190: Static files served by 'tools.staticdir' cause decompression errors in browsers when both 'tools.gzip' and 'tools.caching' are on, but only when served from the cache.
Originally reported by: **Michiel Overtoom (Bitbucket: [MichielOvertoom](http://bitbucket.org/MichielOvertoom), GitHub: @motoom)**

---

I think I found a bug when both 'tools.gzip' and 'tools.caching' are on, for static files served by 'tools.staticdir'. Static content is placed uncompressed in the cache, but incorrectly marked as compressed, and the
result is that only the first serving is send correctly compressed to the client. Subsequent requests cause errors in the browser.

By the way, no exceptions are thrown by CherryPy.

I have attached a small test program which demonstrates this problem. Reloading the static file in the browser should produce the problem.

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1190

hexaclock:

Fixes an issue where an initial page load (when having both gzip and caching enabled) may cause a content decode error. Subsequent page loads work correctly.

#1190

https://github.com/cherrypy/cherrypy/issues/1190

* **Other information**:

### DIFF
--- a/cherrypy/lib/caching.py
+++ b/cherrypy/lib/caching.py
@@ -411,7 +411,9 @@ def tee_output():
                              body, response.time), len(body))
 
     response = cherrypy.serving.response
-    tee(response.body)
+    tee_gen = tee(response.body)
+    if tee_gen:
+        response.body = tee_gen
 
 
 def expires(secs=0, force=False, debug=False):
--- a/cherrypy/lib/caching.py
+++ b/cherrypy/lib/caching.py
@@ -346,6 +346,16 @@ def get(invalid_methods=('POST', 'PUT', 'DELETE'), debug=False, **kwargs):
         if debug:
             cherrypy.log('Reading response from cache', 'TOOLS.CACHING')
         s, h, b, create_time = cache_data
+        if h.get('Content-Encoding') == 'gzip' and b[0:2] != '\x1f\x8b':
+            # In some cases, a request will be cached, have its Content-Encoding header 
+            # set to 'gzip', but not actually gzipped. We check this scenario here,
+            # invalidate any cache entries, and force encoding.gzip to run.
+            cherrypy._cache.delete()
+            response.headers.pop('Content-Length', None)
+            response.headers.pop('Content-Encoding', None)
+            request.cached = False
+            request.cacheable = True
+            return False
         age = int(response.time - create_time)
         if (age > max_age):
             if debug:
@@ -411,10 +421,7 @@ def tee_output():
                              body, response.time), len(body))
 
     response = cherrypy.serving.response
-    tee_gen = tee(response.body)
-    if tee_gen:
-        response.body = tee_gen
-
+    response.body = tee(response.body)
 
 def expires(secs=0, force=False, debug=False):
     """Tool for influencing cache mechanisms using the 'Expires' header.
--- a/cherrypy/lib/caching.py
+++ b/cherrypy/lib/caching.py
@@ -344,7 +344,7 @@ def get(invalid_methods=('POST', 'PUT', 'DELETE'), debug=False, **kwargs):
                 return False
 
         if debug:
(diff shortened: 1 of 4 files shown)


cherrypy/lib/caching.py
```diff
@@ -411,7 +411,9 @@ def tee(body):
                              body, response.time), len(body))
 
     response = cherrypy.serving.response
-    response.body = tee(response.body)
+    tee_gen = tee(response.body)
+    if tee_gen:
```
webknjaz: could you please explain your solution in more detail (and maybe comment/docstring in code)?

webknjaz:
@hexaclock thanks for your contribution!
Any ideas about how to test this? If not, I'll need more time to analyse this change and think about the consequences ;)

hexaclock:
It forces the CachingTool to now run after the encoding tool. In turn, this ensures the `tee_output` attached at before_finalize runs after the content is already gzipped as per `_d.gzip = Tool('before_finalize', encoding.gzip, priority=80)`

webknjaz:
Sounds fair, now it's important to check whether this ordering change does not interfere with other tools. I'll also think about testing as suggested in #1190.

hexaclock:
@webknjaz any luck? I may have some time this weekend to take a stab at writing a test for this based on @motoom's example in #1190.

webknjaz:
@hexaclock sorry, I didn't have time for this yet.

hexaclock:
@webknjaz 
I tried my test against the old _cptools.py and my edits - fails on the old and works on the new, as expected. Modeled after @motoom's paste in #1190.

cherrypy/test/test_caching.py
```diff
@@ -272,6 +286,40 @@ def testExpiresTool(self):
...
+        
+        self.getPage(idx_uri, method='GET', headers=headers)
+        idx_resp_headers = dict(self.headers)
+        idx_gz_content_len = idx_resp_headers['Content-Length']
+        
+        self.getPage(jpg_uri, method='GET', headers=headers)
+        jpg_resp_headers = dict(self.headers)
+        jpg_gz_content_len = jpg_resp_headers['Content-Length']
+        
+        for _ in range(3):
```
webknjaz: Is if required to have 3 queries? Wouldn't it be enough to have 2?

cherrypy/test/test_caching.py
```diff
@@ -129,9 +129,23 @@ def wrongtype(self):
...
+            'tools.gzip.mime_types': ['text/*', 'image/*'],
+            'tools.caching.on': True,
+            'tools.staticdir.on': True,
+            'tools.staticdir.dir': 'static',
+            'tools.staticdir.root': curdir
+            
+        })
+        class GzipStaticCache(object):
+            @cherrypy.expose
+            def dummy(self):
```
webknjaz: I believe, this handler not needed.

cherrypy/test/test_caching.py
hexaclock: Ah yes, you're right. Updated in latest commit.

cherrypy/test/test_caching.py
hexaclock: Removed.

hexaclock:
@webknjaz Assuming this builds ok, are we good to go?

hexaclock:
Seems to have failed due to what seems to be an environment problem: https://pastebin.com/vbhq7kXn

webknjaz:
@hexaclock I've restarted failed jobs

webknjaz:
Overall, looks good, but I still feel a need to check other tools priorities. I don't want this to affect other stuff.

webknjaz:
Okay, I've added some test refactoring and it should be ok now to merge it after tests complete.

PR #1638, opened 2017-10-10
File in sub-dir of staticdir in long-path notation would not load

Safihre:
If a `staticdir` is supplied in Windows long-path notation, serving a file from a subdirectory would fail because `os.stat` does not like the mixing of forward and backwards slashes.
Tested on both Python 2.7.14 and 3.6.2.

In the included test the old behavior would be to do:
```python
os.stat('\\?\C:\cherrypy\static/index.html')
```
Which fails due to the mixed slashes, resulting in a 404.

### DIFF
--- a/cherrypy/lib/static.py
+++ b/cherrypy/lib/static.py
@@ -314,6 +314,13 @@ def staticdir(section, dir, root='', match='', content_types=None, index='',
     branch = request.path_info[len(section) + 1:]
     branch = urllib.parse.unquote(branch.lstrip(r'\/'))
 
+    # On Windows requesting a file in sub-dir of the staticdir results
+    # in mixing of delimiter styles, eg: C:\static\js/script.js
+    # Python normally converts this but not when the staticdir is
+    # supplied in long-path notation, eg: \\?\C:\static\js/script.js
+    if os.name == 'nt':
+        branch = branch.replace('/', '\\')
+
     # If branch is "", filename will end in a slash
     filename = os.path.join(dir, branch)
     if debug:
--- a/cherrypy/test/test_static.py
+++ b/cherrypy/test/test_static.py
@@ -111,6 +111,10 @@ class StaticTest(helper.CPWebCase):
                 'tools.staticdir.dir': 'static',
                 'tools.staticdir.root': curdir,
             },
+            '/static-long': {
+                'tools.staticdir.on': True,
+                'tools.staticdir.dir': r'\\?\%s' % curdir,
+            },
             '/style.css': {
                 'tools.staticfile.on': True,
                 'tools.staticfile.filename': os.path.join(curdir, 'style.css'),
@@ -184,6 +188,14 @@ class StaticTest(helper.CPWebCase):
         #   into \r\n on Windows when extracting the CherryPy tarball so
         #   we just check the content
         self.assertMatchesBody('^Dummy stylesheet')
+
+    @pytest.mark.skipif(platform.system() != 'Windows', reason='Windows only')
+    def test_static_longpath(self):
+        """Test serving of a file in subdir of a Windows long-path staticdir."""
+        self.getPage('/static-long/static/index.html')
+        self.assertStatus('200 OK')
+        self.assertHeader('Content-Type', 'text/html')
+        self.assertBody('Hello, world\r\n')
 
     def test_fallthrough(self):
         # Test that NotFound will then try dynamic handlers (see [878]).


cherrypy/lib/static.py
```diff
@@ -314,6 +314,10 @@ def staticdir(section, dir, root='', match='', content_types=None, index='',
     branch = request.path_info[len(section) + 1:]
     branch = urllib.parse.unquote(branch.lstrip(r'\/'))
 
+    # On Windows, correct the path for long-path support
+    if os.name == 'nt':
+        branch = branch.replace('/', '\\')
```
webknjaz: I think it's better to start relying on `pathlib`

cherrypy/test/test_static.py
```diff
@@ -132,6 +132,14 @@ def dynamic(self):
                 'error_page.404': error_page_404,
             }
         }
+
+        # Windows specific long-path support
+        if(platform.system() == 'Windows'):
```
webknjaz: Don't use parens.

webknjaz: Plus this doesn't need to be conditional.

cherrypy/test/test_static.py
```diff
@@ -132,6 +132,14 @@ def dynamic(self):
...
+            rootconf['/static-long'] = {
+                'tools.staticdir.on': True,
+                'tools.staticdir.dir': '\\\\?\\%s' % curdir,
```
webknjaz: Why don't you use `r'string'` notation here?

webknjaz (changes requested):
Could you please deal with reduced coverage?

cherrypy/lib/static.py
Safihre: Pathlib is not available in Python 2.7 or anything below 3.4 so I don't think it applies right now.

Safihre:
What can I do about the reduced coverage? Have you looked at the report? 
It says reduced coverage because it's platform specific and probably codecov is not testing the Windows specific stuff.

cherrypy/lib/static.py
webknjaz: https://pypi.python.org/pypi/pathlib/

Safihre: You want to introduce a whole new dependency just for a 2 line change?

webknjaz:
@Safihre It looks like codecov's CI detection script has broke: 
https://ci.appveyor.com/project/jaraco/cherrypy/build/1.0.330/job/wljtenhosjpuqhjd#L884

Safihre:
That explains.
Probably because it's building a PR and thus doesn't include secret-vars and thus the key is missing.

webknjaz:
@Safihre it's not related to PR. codecov works w/o tokens for public builds in public CIs

webknjaz:
Refs:
* https://github.com/codecov/codecov-python/issues/114 (PR: https://github.com/codecov/codecov-python/pull/115)
* https://github.com/appveyor/ci/issues/1853

cherrypy/lib/static.py
webknjaz: Well, this just looks redundant to me, that's why I'm looking into using this package, which handles path processing way better than such simplistic things.

And I don't truly understand what is this `\\?` notation and why one would like to combine different delimiter styles. I just want this change, which I cannot check, to not reduce reliability.

Safihre: Agreed, Windows is weird. Only using this notation we can reliably get all paths (including those with unicode) to work on Windows. The mixing of delimiter styles is inherent to Cherrypy running on Windows, in that sense it's actually surprising that it worked so far. Windows users will supply paths in Windows style.
Unfortunately I have no experience with `pathlib` and combined with my lack of understanding of cherrypy internals, me implementing it would probably not work out well.
Simple isn't necessarily bad ;)

webknjaz: Okay, let's postpone pathlib integration question, but I'll ask you to document the problem this change solves in code via comment.

cherrypy/lib/static.py
```diff
@@ -314,6 +314,13 @@ def staticdir(section, dir, root='', match='', content_types=None, index='',
     branch = request.path_info[len(section) + 1:]
     branch = urllib.parse.unquote(branch.lstrip(r'\/'))
 
+    # On Windows requesting a file in sub-dir of the staticdir results
+    # in mixing of delimiter styles, eg: C:\static\js/script.js
+    # Python normally converts this but not when the staticdir is
+    # supplied in long-path notation, eg: \\?\C:\static\js/script.js
+    if os.name == 'nt':
+        branch = branch.replace('/', '\\')
```
webknjaz, line 322: Please use `r'\'`

cherrypy/test/test_static.py
```diff
@@ -185,6 +189,13 @@ def test_static(self):
         #   we just check the content
         self.assertMatchesBody('^Dummy stylesheet')
 
+    @pytest.mark.skipif(platform.system() != 'Windows', reason='Windows only')
+    def test_static_longpath(self):
```
webknjaz: Please add docstring to this test

cherrypy/lib/static.py
Safihre: That's not possible:
```
E       branch = branch.replace('/', r'\')
E                                        ^
E   SyntaxError: EOL while scanning string literal
```
EDIT: not possible in general with only 1 char, not just as function parameter.

webknjaz:
@Safihre plz also rebase your branch in order for it to pick up my workaround for codecov's bug

cherrypy/lib/static.py
webknjaz: Oh.. I see

cherrypy/test/test_static.py
```diff
@@ -185,6 +189,14 @@ def test_static(self):
...
+        """Test serving of a file in subdir of a Windows long-path staticdir"""
```
webknjaz: Please add a period at the end of docstring. This is required by PEP 257.

webknjaz: (this is the last nitpick and I'll merge PR after fixing it)

webknjaz (approved):

webknjaz:
@Safihre thank you!

Issue #1001, opened 2010-04-12
Adding support for client certificate verification in SSLAdapter (patch included)


bb-migration:
Originally reported by: **Anonymous**

---

#### Enhancement
Adding support for client SSL certificate verification to wsgiserver's SSLAdapter

#### Reason
SSL support is critical for modern webservers to provide secure services to users.
However, there are times when applications running behind the webservers
need to determine which clients are actually communicating them. While HTTP
basic_auth can provide authentication, SSL provides another means to verify
client identity: client certification verification.

Similar to the server providing its SSL certificate, when client verification is in use, 
clients must provide a certificate signed by a CA that the server recognizes in order
for the client to be allowed to connect.

#### Changes
This patch adds another optional keyword argument to the SSLAdapter **`__init__()`** called
''client_CA''. ''client_CA'' is a string that contains a path to a CA certificate. When client_CA
is present, the SSLAdapter knows to perform client verification using this CA. When 
absent, SSLAdapter behaves as before, ie with no client verification. 

#### Bugs/Issues
While verification is performed correctly for both the ssl_pyopenssl SSLAdapter and the 
ssl_builtin SSLAdapter, the different implementations provide varying levels of support 
for SSL client environment variables that are traditionally provided by Apache's mod_ssl.
See this page for details [Mod_SSL Environment Variables](http://www.modssl.org/docs/2.8/ssl_reference.html#ToC25).

ssl_pyopenssl currently provides '''no''' client environment variables due to the fact that the
SSL handshake and thus the access to the client's certificate occurs at first data transfer - 
well after the environment variables are set by the SSLAdapter `wrap()` function.

ssl_builtin provides minimal environment variables. The major limiting factor is that
python's builtin ssl routines only expose a small amount of information about the 
certificates, and then only for the client certificate. This problem will be difficult to fix
if ssl_builtin must depend solely on python's ssl.

#### Patch
See attached Diff 

Reported by nmitchell@anl.gov

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1001

ghost:
_Original comment by_ **Anonymous**:

---

I ran into a problem with this patch in the tick() method of the `HTTPServer` class in `wsgiserver/__init__.py`

[...]

PR #1644, opened 2017-10-14
Require client certificates

dpeschman:
Feature

#1001 - partial implementation

Using the built-in SSL adapter, client certificates are neither required nor validated with a CA. Client certificate fields (subject, issuer) are not passed downstream via WSGI environment.

New field `ssl_verify_mode` on Server can be one of none/optional/required to specify whether client certificates are required. These values map to Python's verify_mode values [1]. The default is "none" for backwards compatibility.

Subject and issuer of the client certificate are passed downstream via WSGI environment. For instance, certificate subject is passed down as SSL_CLIENT_S_DN_* fields.

* **Other information**:
This change depends on changes to Cheroot [2].

Changes to the base test classes were required to wipe out configuration changes made by the new tests.  Without them, tests run after the new tests would try to connect over HTTPS when they didn't want to, and similar issues.

This patch was inspired by, and much of the code derived from, the patch in #1001. That change includes additional features, such as hostname verification and parallel support for pyOpenSSL, which are omitted here.

[1] https://docs.python.org/3.5/library/ssl.html?highlight=ssl#constants
[2] https://github.com/cherrypy/cheroot/pull/52


cherrypy/test/test_client_cert.py
```diff
@@ -0,0 +1,118 @@
...
+    def test_connect(self):
+        """If server_should_reject is false, verify the index can be fetched.
+        Otherwise, verify that a URLError is raised when connecting.
+        """
+        context = ssl.create_default_context(cafile=self.ca_chain)
+        context.load_cert_chain(self.client_cert, keyfile=CLIENT_KEY)
+        context.check_hostname = False
+
+        if self.server_should_reject:
+            self.assertRaises(
```
webknjaz: We are trying to move away from pure `unittest`. Could you please re-write this using pytest?

cherrypy/test/test_client_cert.py
```diff
@@ -0,0 +1,140 @@
...
+CLIENT_CERT = join(SSL_DIR, 'client.cert')
+CLIENT_IP_CERT = join(SSL_DIR, 'client_ip.cert')
+CLIENT_WILD_CERT = join(SSL_DIR, 'client_wildcard.cert')
+CLIENT_WRONG_CA = join(SSL_DIR, 'client_wrong_ca.cert')
+CLIENT_WRONG_HOST = join(SSL_DIR, 'client_wrong_host.cert')
+SERVER_KEY = join(SSL_DIR, 'server.key')
+SERVER_CERT = join(SSL_DIR, 'server.cert')
+
+
+def subdict(d, key_predicate):
```
webknjaz: plz choose a better name

cherrypy/test/test_client_cert.py
```diff
@@ -0,0 +1,140 @@
...
+
+    server_should_reject = False
+    server_host = 'localhost'
+    client_host = 'localhost'
+    ca_chain = CA_CERT
+    client_cert = CLIENT_CERT
+    verify_mode = None  # None, 'none', 'optional', 'required'
+
+    @classmethod
+    def setup_server(cls):
```
webknjaz, line 44: How do you run this method exactly?

cherrypy/test/test_client_cert.py
```diff
@@ -0,0 +1,140 @@
...
+        assert wsgi_env['SSL_CLIENT_I_DN_O'] == 'CherryPy'
+        assert wsgi_env['SSL_CLIENT_I_DN_ST'] == 'XX'
+
+
+BAD_CLIENT_CERTS = [CLIENT_WRONG_CA]
+GOOD_CLIENT_CERTS = [CLIENT_CERT, CLIENT_IP_CERT, CLIENT_WILD_CERT]
+SERVER_HOSTS = ['localhost', '127.0.0.1', '0.0.0.0']
+CLIENT_HOSTS = ['localhost']
+
+TESTS = [
```
webknjaz, line 105: Do you think you could rewrite this using `@pytest.parametrize`?

cherrypy/test/helper.py
```diff
@@ -226,23 +226,24 @@ def _setup_server(cls, supervisor, conf):
...
+            baseconf['server.ssl_certificate'] = None
+            baseconf['server.ssl_private_key'] = None
 
         # helper must be imported lazily so the coverage tool
         # can run against module-level statements within cherrypy.
         # Also, we have to do "from cherrypy.test import helper",
         # exactly like each test module does, because a relative import
         # would stick a second instance of webtest in sys.modules,
         # and we wouldn't be able to globally override the port anymore.
-        if supervisor.scheme == 'https':
```
webknjaz, line 236: why did you drop this?

webknjaz:
@dpeschman please rebase your branch on top of master

cherrypy/test/test_client_cert.py
dpeschman: I'll give it a shot.

cherrypy/test/test_client_cert.py
dpeschman: It's run by CPWebCase.setup_class; this class, ClientCertTests, is multi-inherited with CPWebCase into the derived test classes at line 137.

cherrypy/test/helper.py
dpeschman: This would change HTTP_CON in the webtest.WebCase base class to HTTPS; all subsequent tests would use HTTPS and fail.  webtest.WebCase already configures its connection type properly on its own.

cherrypy/test/test_client_cert.py
dpeschman: I tried changing ClientCertTests to inherit from CPWebCase, and added a parameterized test function to it, but pytest parameterize does not work on unittest.TestCase test methods.  My other ideas to do this are bad, requiring code duplication from CPWebCase.  Do you have any suggestions?

dpeschman:
I don't know what I did wrong, but this PR now includes all the commits in the world.  Let me know when is a good time, and I'll make a new PR with just my changes.

webknjaz:
This pull request fixes 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/cherrypy/cherrypy/rev/pr-fe3d701792b46fb4fc77d75ac5f718c055fc26c2)

**fixed alerts:**

* 1 for __init__ method calls overridden method

---

PR #1645, opened 2017-10-17
Added uuidv4 and now() in RFC3339 UTC Format

spagno:
feature

There isn't the possibility to configure acces_log_format with uuidv4 and timestamp in rfc3339 format

You can use "%(i)s" to have a unique request_id in the log
You can use "%(z)s" to have a UTC rfc3339 timestamp in the log

* **Other information**:
first PR ever, if something is wrong please don't be mean :dagger:

### DIFF
--- a/cherrypy/_cplogging.py
+++ b/cherrypy/_cplogging.py
@@ -262,7 +262,7 @@ class LogManager(object):
                  'f': dict.get(inheaders, 'Referer', ''),
                  'a': dict.get(inheaders, 'User-Agent', ''),
                  'o': dict.get(inheaders, 'Host', '-'),
-                 'i': str(response.uuid),
+                 'i': request.unique_id,
                  'z': LazyRfc3339UtcTime(),
                  }
         if six.PY3:
--- a/cherrypy/_cplogging.py
+++ b/cherrypy/_cplogging.py
@@ -314,6 +314,7 @@ class LogManager(object):
         month = monthnames[now.month - 1].capitalize()
         return ('[%02d/%s/%04d:%02d:%02d:%02d]' %
                 (now.day, month, now.year, now.hour, now.minute, now.second))
+                
     @staticmethod
     def time_z(self):
         """Return now() in RFC3339 UTC Format."""
--- a/cherrypy/_cplogging.py
+++ b/cherrypy/_cplogging.py
@@ -138,6 +138,7 @@ class NullHandler(logging.Handler):
     def createLock(self):
         self.lock = None
 
+
 class GetUuid(object):
     def __init__(self):
         self._uuid = None
@@ -147,6 +148,7 @@ class GetUuid(object):
         if self._uuid is None:
             self._uuid = uuid.uuid4()
         return self._uuid
+
 
 class LogManager(object):
 
--- a/cherrypy/_cplogging.py
+++ b/cherrypy/_cplogging.py
@@ -113,7 +113,6 @@ import os
 import sys
 
 import six
-import uuid
 
 import cherrypy
 from cherrypy import _cperror
@@ -252,7 +251,6 @@ class LogManager(object):
             status = response.output_status.split(ntob(' '), 1)[0]
             if six.PY3:
                 status = status.decode('ISO-8859-1')
-
         atoms = {'h': remote.name or remote.ip,
                  'l': '-',
                  'u': getattr(request, 'login', None) or '-',
@@ -263,7 +261,7 @@ class LogManager(object):
                  'f': dict.get(inheaders, 'Referer', ''),
                  'a': dict.get(inheaders, 'User-Agent', ''),
                  'o': dict.get(inheaders, 'Host', '-'),
-                 'i': str(uuid.uuid4()),
+                 'i': response.uuid,
                  'z': self.time_z(),
                  }
         if six.PY3:
--- a/cherrypy/_cplogging.py
+++ b/cherrypy/_cplogging.py
@@ -111,10 +111,12 @@ import datetime
 import logging
 import os
 import sys
+
+import six
 import uuid
 
 import cherrypy
-import six
+
 from cherrypy import _cperror
 from cherrypy._cpcompat import ntob
 
@@ -215,8 +217,7 @@ class LogManager(object):
         if traceback:
             exc_info = _cperror._exc_info()
 
-        self.error_log.log(severity, ' '.join(
-            (self.time(), context, msg)), exc_info=exc_info)
+        self.error_log.log(severity, ' '.join((self.time(), context, msg)), exc_info=exc_info)
 
     def __call__(self, *args, **kwargs):
         """An alias for ``error``."""
--- a/cherrypy/_cplogging.py
+++ b/cherrypy/_cplogging.py
@@ -317,7 +317,7 @@ class LogManager(object):
     def time_z(self):
         """Return now() in RFC3339 UTC Format."""
         now = datetime.datetime.now()
-        return now.isoformat("T") + "Z"
+        return now.isoformat('T') + 'Z'
 
     def _get_builtin_handler(self, log, key):
         for h in log.handlers:
--- a/cherrypy/_cplogging.py
+++ b/cherrypy/_cplogging.py
@@ -314,7 +314,7 @@ class LogManager(object):
         month = monthnames[now.month - 1].capitalize()
         return ('[%02d/%s/%04d:%02d:%02d:%02d]' %
                 (now.day, month, now.year, now.hour, now.minute, now.second))
-
+    @staticmethod
     def time_z(self):
         """Return now() in RFC3339 UTC Format."""
         now = datetime.datetime.now()
--- a/cherrypy/_cplogging.py
+++ b/cherrypy/_cplogging.py
@@ -119,6 +119,7 @@ import cherrypy
 from cherrypy import _cperror
 from cherrypy._cpcompat import ntob
 
+
 # Silence the no-handlers "warning" (stderr write!) in stdlib logging
 logging.Logger.manager.emittedNoHandlerWarning = 1
 logfmt = logging.Formatter('%(message)s')
--- a/cherrypy/_cplogging.py
+++ b/cherrypy/_cplogging.py
@@ -113,6 +113,7 @@ import os
 import sys
 
 import six
+import uuid
 
 import cherrypy
 from cherrypy import _cperror
@@ -136,6 +137,39 @@ class NullHandler(logging.Handler):
 
     def createLock(self):
         self.lock = None
+
+
+class CachedUuid(object):
+    def __init__(self, wrapped):
+        self.wrapped = wrapped
+        try:
+            self.__doc__ = wrapped.__doc__
+        except:  # pragma: no cover
+            pass
+
+    # original sets the attributes on the instance
+    # def __get__(self, inst, objtype=None):
+    #    if inst is None:
+    #        return self
+    #    val = self.wrapped(inst)
+    #    setattr(inst, self.wrapped.__name__, val)
+    #    return val
+
+    # ignore the instance, and just set them on the class
+    # if called on a class, inst is None and objtype is the class
+    # if called on an instance, inst is the instance, and objtype
+    # the class
+    def __get__(self, inst, objtype=None):
+        # ask the value from the wrapped object, giving it
+        # our class
+        val = self.wrapped(objtype)
+
+        # and set the attribute directly to the class, thereby
+        # avoiding the descriptor to be called multiple times
+        setattr(objtype, self.wrapped.__name__, val)
+
+        # and return the calculated value
+        return val
 
 
 class LogManager(object):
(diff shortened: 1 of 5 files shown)


webknjaz:
Hi @spagno,

What did you try to do here? What went wrong? Do you need any assistance with this?

webknjaz:
If you're concerned about broken linters, you just need to fix this:
https://travis-ci.org/cherrypy/cherrypy/jobs/289121270#L715

And then pushing new commit to the branch should fix that. You'll need to reopen this PR in such case.

spagno:
thank you very much, I'm trying to understand why I broke that

spagno:
Understood, I'll reopen the PR

webknjaz:
Oh, just follow the link and you'll see:
```
Fix double quoted strings................................................Failed
hookid: double-quote-string-fixer
Files were modified by this hook. Additional output:
Fixing strings in cherrypy/_cplogging.py
```

Which means that you used double quotes for strings, while the convention is to use single quotes.

You also can `pip install pre-commit` locally + do `pre-commit run --all-files` manually if needed :)

webknjaz:
@spagno may I expect any tests for this additionally?

spagno:
@webknjaz yes, I'm working on it. Just don't know if test all parameters (creating a new testing def) or just the new ones

webknjaz:
@spagno both cases are fine.
I'd be thankful for testing just your changes, but if you're able to cover more — that would be fantastic :)

spagno:
@webknjaz well I'll try!

spagno:
seems like python in windows manages log differently from linux; can't run appropriate tests on windows, I have to delete the tests (for now) and reimplement them

cherrypy/_cplogging.py
```diff
@@ -111,14 +111,13 @@
 import logging
 import os
 import sys
-
-import six
+import uuid
 
 import cherrypy
+import six
```
webknjaz: Don't move it here, as it's a third-party. It must remain between stdlib imports and the project imports.

cherrypy/_cplogging.py
```diff
@@ -216,7 +215,8 @@ def error(self, msg='', context='', severity=logging.INFO,
         if traceback:
             exc_info = _cperror._exc_info()
 
-        self.error_log.log(severity, ' '.join((self.time(), context, msg)), exc_info=exc_info)
+        self.error_log.log(severity, ' '.join(
```
webknjaz: I think, it's nicer to split lines between arguments of `log()` method, not inside argument expressions.
Also please not that we follow 120-chars line restrictions (there's no 80-char terminals nowadays), so you don't really have to split it.

cherrypy/_cplogging.py
```diff
@@ -262,6 +262,8 @@ def access(self):
                  'f': dict.get(inheaders, 'Referer', ''),
                  'a': dict.get(inheaders, 'User-Agent', ''),
                  'o': dict.get(inheaders, 'Host', '-'),
+                 'i': str(uuid.uuid4()),
```
webknjaz: Why exactly do you need a new uuid for each access call shouldn't it be per-request?

cherrypy/_cplogging.py
```diff
@@ -262,6 +262,8 @@ def access(self):
...
+                 'z': self.time_z(),
```
webknjaz: What about exposing these atoms to error log?

cherrypy/test/logtest.py
```diff
@@ -4,7 +4,6 @@
 import time
 
 import six
-
```
webknjaz: no, there must be separation between stdlib and project imports

webknjaz (changes requested):

webknjaz:
@spagno please don't remove tests w/o review as it's harder for me to find them in history and nearly impossible to comment in a nice way.

I've looked at the failed test and it turns out that your `assertValidUUIDv4 ` helper has this `data` var:
```python
data = [b'127.0.0.1 - - [18/Oct/2017:01:46:50] "GET /as_yield HTTP/1.1" 200 7 "" ""\r\n', b"b'test suite marker: '1508291210....RER" "USERAGENT" HOST\r\n', b"b'test suite marker: '1508291210.6715672\n", b'21afb35b-518c-4c20-a745-c6c071ba82c8\r\n']
```

And then at the first iteration of the loop you throw an error, while the UUID is the 3rd element.

*If what I see in log is correct*, you could improve it via smth like this:
```python
...
data = [chunk.decode('utf-8').rstrip('\n').rstrip('\r') for chunk in data]
for log_chunk in data:
    try:
        uuid_log = data[-1]
        uuid_obj = UUID(uuid_log, version=4)
    except (TypeError, ValueError):
        msg = '%r not a valid UUIDv4' % uuid_log
        self._handleLogError(msg, data, marker, lines)
    else:
        if str(uuid_obj) != uuid_log:
            msg = '%r not a valid UUIDv4' % uuid_log
            self._handleLogError(msg, data, marker, lines)
```

cherrypy/_cplogging.py
spagno: sorry, it was the visualstudio code import sorter, I'll not use it anymore

cherrypy/_cplogging.py
spagno: because if you want to set different logformat with cherrypy._cplogging.LogManager.access_log_format(format)
if you want to have unique uuid for each request I thougth that could be the right place to set the uuid.
Use case:
I'm developing a simple semaphore dashboard for our CI; I want that the dashboard logs would be compliant with our best practices: in my case the configuration would be:
cherrypy._cplogging.LogManager.access_log_format('{"request_id":"%(i)s", "user-agent":"%(a)s", "timestamp": "%(z)s", "referrer":"%(f)s", "status":%(s)s, "remote_ip":"%(h)s"}')

webknjaz: I mean that it would call `uuid.uuid4()` every time `access()` is called if I remember correctly, meaning that there can be several different UUIDs generated per request if you'd call `access()` more then once. Right?

webknjaz: I've explored prior art and found out that @OpenStack uses `%(uuid)s` format string for this, so I'd like to follow the same convention.
Do you know any project using `%(i)s` for this?

spagno: Nope, uuid is generally custom (for example nginx) and in some cases it isn't supported and you have to do workaround for it (in haproxy you have to use a lua script for that). I decided to use "i" because there other atoms are single char, to mantain a sort of "convention"

webknjaz: oh.. okay, makes sense. But I'm still not sure about multiple calls

spagno: every call should have an unique uuid to track every single request; If there's a better place for doing that, it's even better!

webknjaz: Imagine that you can manually invoke `access()` several times in your handler. What would be the result?

spagno: maybe I didn't understand well the code but isn't it already invoked every time there's an access log?

webknjaz: It is invoked once per request, but who stops users from doing this again in their code?

cherrypy/_cplogging.py
```diff
@@ -312,6 +314,11 @@ def time(self):
         return ('[%02d/%s/%04d:%02d:%02d:%02d]' %
                 (now.day, month, now.year, now.hour, now.minute, now.second))
 
+    def time_z(self):
```
webknjaz: @spagno plz turn this into `@staticmethod`

cherrypy/_cplogging.py
```diff
@@ -113,12 +113,13 @@
 import sys
 
 import six
+import uuid
 
 import cherrypy
+
```
webknjaz: remove this line

cherrypy/_cplogging.py
```diff
@@ -113,12 +113,13 @@
...
 from cherrypy import _cperror
 from cherrypy._cpcompat import ntob
 
-
```
webknjaz: restore this line

cherrypy/_cplogging.py
```diff
@@ -113,12 +113,12 @@
 import sys
 
 import six
+import uuid
 
 import cherrypy
 from cherrypy import _cperror
 from cherrypy._cpcompat import ntob
 
-
```
webknjaz: restore this line

cherrypy/test/test_logging.py
```diff
@@ -127,6 +127,44 @@ def testCustomLogFormat(self):
 
         cherrypy._cplogging.LogManager.access_log_format = original_logformat
 
+    def testUUIDv4ParameterLogFormat(self):
+        '''Test a customized access_log_format string,
```
webknjaz: This docstring is not PEP257 compliant

cherrypy/test/test_logging.py
```diff
@@ -127,6 +127,44 @@ def testCustomLogFormat(self):
...
+        )
+
+        self.markLog()
+        self.getPage('/as_string')
+        self.assertValidUUIDv4()
+
+        cherrypy._cplogging.LogManager.access_log_format = original_logformat
+
+    def testTimezLogFormat(self):
+        '''Test a customized access_log_format string,
```
webknjaz: Use double quotes for docstrings and follow PEP257

spagno: sorry, just copied the original testCustomLogFormat (which has ''' )
I also fixed that

cherrypy/test/test_logging.py
```diff
@@ -107,8 +107,8 @@ def testNormalYield(self):
                            % self.prefix())
 
     def testCustomLogFormat(self):
-        '''Test a customized access_log_format string,
-           which is a feature of _cplogging.LogManager.access() '''
+        """Test a customized access_log_format string,
```
webknjaz: https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

webknjaz:
A few comments are still unanswered

cherrypy/_cplogging.py
spagno: I got your point but I don't have idea how to solve this: maybe i should wrap the creation of the uuid like the time() and time_z() methods. Could be reasonable?

cherrypy/_cplogging.py
spagno: you are right, I'll check how to do it

spagno: I read the code and I'm in a little in trouble: in the error_log the atoms aren't used at all, is that right? If it's right I think that adding this support could be out of scope of this PR and I can work on that with a different PR if there's the interest about that

webknjaz: well, supporting these atoms we allow users use them if they want to change that error_log format template. Okay, you can work on this separately.

cherrypy/_cplogging.py
webknjaz: I'd go for some lazy property: on the first invocation it should save the result to `cherrypy.serving.request` (which is thread-local) and if the value is saved there just return it. I think, you should put a lazy object here, so that it would be called only if rendered into log with `__str__()`.
Also I'm thinking of that not everyone might want to have that in their logs, so maybe in makes sense to only invoke this if they've added `%(i)s` into their format string.

webknjaz: (applies to both `%(i)s` and `%(z)s`)

spagno: Agree, for that reason I added only the atoms without touching the default access_log_format which is in standard Apache format

webknjaz: But they are still not lazy, meaning we waste CPU time calculating them.

spagno: got it, I'll fix it, thanks for the tips and sorry for the poor start quality: like I said it's my first public PR and I'm a sysadmin not a developer :D

webknjaz: No worries @spagno, I'm not judging. I'm here to help and walk you through the process :) You're doing great!

cherrypy/_cprequest.py
```diff
@@ -911,6 +912,9 @@ def collapse_body(self):
         self.body = newbody
         return newbody
 
+    def __str__(self):
```
webknjaz: Oops.. You got me wrong. This magic method here will cause `str(response)` to return the UUID, breaking representation of the response itself. I meant you need to assign some lazy object as request's attribute from within `_cplogging` module.

cherrypy/test/test_logging.py
```diff
@@ -127,6 +125,40 @@ def testCustomLogFormat(self):
...
+        self.getPage('/as_string')
+        self.assertValidUUIDv4()
+
+        cherrypy._cplogging.LogManager.access_log_format = original_logformat
+
+    def testTimezLogFormat(self):
+        """Test a customized access_log_format string, which is a feature of _cplogging.LogManager.access()."""
+        original_logformat = cherrypy._cplogging.LogManager.access_log_format
+        cherrypy._cplogging.LogManager.access_log_format = (
+            '{h} {l} {u} {z} "{r}" {s} {b} "{f}" "{a}" {o}' if six.PY3
```
webknjaz: should have `{i}`

spagno:
They are 3 different tests:
1. test the custom apache log
2. test only the uuid
3. test the modified apache log with the timestamp with timezone instead of
the apache's one

we can collapse 2 and 3 adding also {i} testing all atoms togheter , do you
prefer that?

regards

2017-10-20 11:36 GMT+02:00 Sviatoslav Sydorenko <notifications@github.com>:

webknjaz:
@spagno I've pushed improvements and everything is likely to work well for now. Still, I'd like to better test time_z (maybe use spy object for that).

Issue #1646, opened 2017-10-18
HTTPError doesn't support http.client constants.


coady:
* **I'm submitting a ...**
[x] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Do you want to request a *feature* or report a *bug*?**
*bug*

* **What is the current behavior?**
`http.client` constants aren't accepted as valid input to `HTTPError`.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem**.
```
In []: cherrypy.HTTPError(http.client.NOT_FOUND)
...
Out []: HTTPError: (500, "Illegal response status from server ('HTTPStatus.NOT_FOUND' is non-numeric).")
```

* **What is the expected behavior?**
The same as with any `int`.

* **What is the motivation / use case for changing the behavior?**
It's effectively a Python 3 regression, since Python 2 `httplib` constants work, and an `IntEnum` is a valid `int`.

* **Please tell us about your environment:**

- Cheroot version: X.X.X
- CherryPy version: 11.X.X
- Python version: 3.6.X
- OS: XXX
- Browser: [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)

webknjaz:
Hi @coady,

Thanks for letting us know. While I cannot promise you an instant fix, I'll gladly accept any contributions (e.g. tests, patch, docs) helping to improve this behavior.

PR #1647, opened 2017-10-19
Fix for http.client status codes.

coady:
Bug fix

#1646

#1646

* **Other information**:

### DIFF
--- a/cherrypy/test/test_httputil.py
+++ b/cherrypy/test/test_httputil.py
@@ -2,31 +2,50 @@
 import pytest
 
 from cherrypy.lib import httputil
+from six.moves import http_client
 
 
-class TestUtility(object):
-    @pytest.mark.parametrize(
-        'script_name,path_info,expected_url',
-        [
-            ('/sn/', '/pi/', '/sn/pi/'),
-            ('/sn/', '/pi', '/sn/pi'),
-            ('/sn/', '/', '/sn/'),
-            ('/sn/', '', '/sn/'),
-            ('/sn', '/pi/', '/sn/pi/'),
-            ('/sn', '/pi', '/sn/pi'),
-            ('/sn', '/', '/sn/'),
-            ('/sn', '', '/sn'),
-            ('/', '/pi/', '/pi/'),
-            ('/', '/pi', '/pi'),
-            ('/', '/', '/'),
-            ('/', '', '/'),
-            ('', '/pi/', '/pi/'),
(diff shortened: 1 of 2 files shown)


cherrypy/test/test_httputil.py
```diff
@@ -30,3 +31,17 @@ def test_urljoin(self, script_name, path_info, expected_url):
         """Test all slash+atom combinations for SCRIPT_NAME and PATH_INFO."""
         actual_url = httputil.urljoin(script_name, path_info)
         assert actual_url == expected_url
+
+
+def test_valid_status():
```
webknjaz: Would you please move this to class or remove the class above so that all tests would be in the same style?

cherrypy/test/test_httputil.py
```diff
@@ -30,3 +31,17 @@ def test_urljoin(self, script_name, path_info, expected_url):
...
+    code, reason, message = httputil.valid_status(200)
+    assert code == 200 and reason == 'OK'
```
webknjaz: Could you please rewrite this using `@pytest.parametrize`, so that the test body would be simpler like one above

cherrypy/test/test_httputil.py
```diff
@@ -30,3 +31,17 @@ def test_urljoin(self, script_name, path_info, expected_url):
...
+
+    code, reason, message = httputil.valid_status('500')
+    assert code == 500 and reason == 'Internal Server Error'
+
+    code, reason, message = httputil.valid_status(http_client.NOT_FOUND)
+    assert code == 404 and reason == 'Not Found'
+
+    with pytest.raises(ValueError):
```
webknjaz: Could you please test more branches, so that coverage would not go down?

coady: The changed code has 100% coverage.  Overall coverage % can decrease when (covered) code is removed.

webknjaz: Well, that is weird :)

cherrypy/test/test_httputil.py
```diff
@@ -2,31 +2,50 @@
 import pytest
 
 from cherrypy.lib import httputil
+from six.moves import http_client
```
webknjaz: Please note that the imports order must be the following: stdlib, then third-party and finally our project imports

webknjaz:
@coady I've added bits of refactoring to your branch. Thank you for your contribution!

PR #1649, opened 2017-10-21
Upgrade pre-commit hooks

asottile:
I also made the pep257 exclude regex a little more readable :)


webknjaz:
@asottile the point of one line regexp was that I could easily cut off that line with `sed`, see `tox.ini`.

It's used in PEP 257 marked job, which basically exists because not all of the files are compliant with it.

.pre-commit-config.yaml
```diff
@@ -1,21 +1,18 @@
 -   repo: git://github.com/pre-commit/pre-commit-hooks
-    sha: 5bf6c09
+    sha: v1.1.1
     hooks:
     -   id: trailing-whitespace
         exclude: cherrypy/test/static/index.html
     -   id: flake8
-        args:
```
webknjaz: I prefer multi line sequences, because they produce better diffs and are easily readable. (I also often synchronize contents via `vimdiff`)

.pre-commit-config.yaml
```diff
@@ -37,14 +34,29 @@
     sha: f464d46
     hooks:
     -   id: pep257
-        exclude: tests/dist|cherrypy/(_cp|__init__|__main__|daemon|_helper|(lib/(__init__|auth|auth_basic|auth_digest|caching|covercp|cpstats|cptools|encoding|gctools|httpauth|httputil|jsontools|lockfile|locking|profiler|reprconf|sessions)|process/(plugins|servers|win32)|scaffold/__init__).py|test|tutorial)
+        exclude: |
```
webknjaz: This is tightly coupled with the hack in `tox.ini` removing `exclude:` from here in order to list all files not yet converted.

asottile:
cool, I backed out the formatting changes.  Also used the `--config` feature to eliminate the `sed` hack, let me know what you think :)

webknjaz (approved):

webknjaz:
Wow! It looks much nicer now! Didn't know about `--config` option.
Thank you very much!

Issue #1625, opened 2017-08-18
Response timeout is harmful - remove it


jaraco:
* **I'm submitting a ...**
[X ] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **What is the current behavior?**

As described in [this answer](https://stackoverflow.com/a/26304235/70170), when CherryPy raises a TimeoutError, it's doing so because the request took longer than the timeout to complete, but it was still allowed to complete.

The [docs say](http://docs.cherrypy.org/en/latest/advanced.html#response-timeouts):

> response.timeout: the number of seconds to allow responses to run

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem**.

Consider this script:

```
__requires__ = ['cherrypy']

import time
import os

import cherrypy


class Server:
	@cherrypy.expose
	def index(self):
		time.sleep(5)
		return 'done!'

	@classmethod
	def run(cls):
		config = {
			'global': {
				'server.socket_port': int(os.environ.get('PORT', 8080)),
				'engine.timeout_monitor.frequency': 1,
				'response.timeout': 2,
			},
		}
		cherrypy.quickstart(cls(), config=config)


__name__ == '__main__' and Server.run()
```

Run it with [rwt](https://pypi.org/project/rwt):

```
$ PORT=8081 rwt -- test-timeout.py
Collecting cherrypy
  Using cached CherryPy-11.0.0-py2.py3-none-any.whl
Collecting six (from cherrypy)
  Using cached six-1.10.0-py2.py3-none-any.whl
Collecting cheroot>=5.2.0 (from cherrypy)
  Using cached cheroot-5.8.3-py2.py3-none-any.whl
Collecting portend>=2.1.1 (from cherrypy)
  Using cached portend-2.1.2-py2.py3-none-any.whl
Collecting tempora>=1.8 (from portend>=2.1.1->cherrypy)
  Using cached tempora-1.9-py2.py3-none-any.whl
Collecting pytz (from tempora>=1.8->portend>=2.1.1->cherrypy)
  Using cached pytz-2017.2-py2.py3-none-any.whl
Installing collected packages: six, cheroot, pytz, tempora, portend, cherrypy
Successfully installed cheroot-5.8.3 cherrypy-11.0.0 portend-2.1.2 pytz-2017.2 six-1.10.0 tempora-1.9
[18/Aug/2017:13:22:38] ENGINE Listening for SIGTERM.
[...]

PR #1651, opened 2017-10-26
Remove request timeout support

jaraco:

Feature removal.

#1625

A timeout monitor will periodically update a flag indicating if a request has exceeded its declared timeout. If that flag was set while the request was processed, it will raise an error rather than returning the result.

Instead of monitoring and setting flags, it's left to the implementer to monitor the time elapsed and act accordingly.

* **Other information**:


jaraco:
I'm waffling on whether TimeoutError should be removed. On one hand, it may be nice to supply it in case applications wish to implement their own timeout logic and raise (and catch) the same exceptions. On the other hand, I'm struggling to think of how an application may have ever used this exception.

The more I think about it, the more inclined I am to remove it... because of course we can restore it readily if it turns out someone needs it.

PR #1656, opened 2017-10-29
Use single subscribe method for decorator or direct invocation

jaraco:
This approach presents an alternative approach to afb5b33fe3, in which only one subscribe method is presented and the decorator functionality is exposed when no callback is supplied.

Thoughts?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,8 +1,9 @@
 v11.2.0
 -------
 
-* Added ``cherrypy.engine.subscribed``, providing syntactic sugar
-  for decorating a callback for ``cherrypy.engine.subscribe``.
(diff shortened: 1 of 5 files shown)


webknjaz:
```python
E   TypeError: subscribe() got multiple values for keyword argument 'channel'
```

jaraco:
Ugh. Did I really do that?

I tested the code, then reformatted it, then committed and pushed and made a PR. ...but I added a regression in the reformatting, now corrected.

webknjaz (approved):

webknjaz:
@jaraco try re-running https://ci.appveyor.com/project/jaraco/cherrypy/build/1.0.459/job/cls4x77usuklwess and feel free to merge this PR

PR #1672, opened 2017-12-17
Factor classproperty out (referring to #1671)

benjaoming:

This is a variant of #1671, perhaps less complicated because it removes the use of `classproperty`. I'd be in favor of this because: 1 less dependency, less code, less engineered


benjaoming:
Nevermind, I get the idea that this should be evaluated lazily now! So needs a bit of engineering :)

PR #1677, opened 2017-12-20
Add VS Code editor and virtualenv to .gitignore

Sonophoto:

Dev-ops Update

Modifies .gitignore in project repo to ignore configuration files for:

   Visual Studio Code Editor: https://code.visualstudio.com/
   Virtualenv: https://pypi.python.org/pypi/virtualenv

N/A

These configuration files are present in calls to `git status` and `git add . --all`

These files are ignored by `git status` and `git add . --all`

* **Other information**:

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,9 +27,14 @@ sphinx/source/_build
 # Py.test
 /.cache/
 
-# Eclipse, PyDev
+# Eclipse, PyDev, VS Code
 .project
 .pydevproject
+.vscode
+
+# virtualenv
+VENV
+venv
 
 # test results in junit format for Appveyor
 /junit-test-results.xml


.gitignore
```diff
@@ -27,9 +27,14 @@ sphinx/source/_build
...
 /.cache/
 
-# Eclipse, PyDev
+# Eclipse, PyDev, VS Code
 .project
 .pydevproject
+.vscode
+
+# virtualenv
+VENV
```
webknjaz, line 36: Modern env managers allow keeping env folders outside of the repo. I suggest you using those, like pyenv, for example. Who knows whether we decide having these folder names in repo in future?

/cc: @jaraco

Sonophoto: Ah, Ok, Well sorry to bother you with my outdated tools.

webknjaz: No worries. I could still accept vscode part. I didn't, because it wasn't a separate PR.

jaraco:
In my opinion, .gitignore is abused to exclude everybody's personal local files. That's not the appropriate place for these exclusions. Exclusions that are broadly excluded (*.pyc) or excluded for a particular developer's environment (.vscode) should go in the developer's global .gitignore file, rather than copied into every .gitignore of every project that developer touches. Here's what my .gitignore_global file looks like:

```
$ cat ~/.gitignore_global 
*~
.DS_Store
.cache
.eggs
__pycache__
build
dist
*.egg-info
.tox
.xprocess
*.orig
*.pyc
node_modules
.mypy_cache
```

Then, individual projects should include a .gitignore that ignores files that are specifically generated by that project or expected to be there uniquely for that project but aren't already broadly ignored. That way, most projects have no .gitignore or a short .gitignore that ignores files distinct to that project.

Otherwise, what you end up with is a .gitignore in each project that's full of ignores that are irrelevant to most environments and makes it difficult to discern cruft from intentional entries.

That said, cherrypy already has a .gitignore laden with cruft, so what's two more?

Sonophoto:
Cruft is bad. I agree 100%

I think your opinion is 100% rational. I did not know about the global .gitignore config and now I have one set up. Thank you very much for the tip.

.gitignore
Sonophoto: I've switched to using the global .gitignore. I was not aware of that feature. 

We use virtualenv because it is built-in to python. We use built-in libraries (batteries included) in all cases without a very very good reason to use an external library.

webknjaz:
@jaraco Personally, I'd prefer it to be project-specific and reflect common practices i.e. be in sync with some upstream templates like https://www.gitignore.io/api/osx%2Cweb%2Cgit%2Cpydev%2Clinux%2Cpython%2Cwindows%2Cpycharm%2Ball

Personally I don't enjoy maintaining my global gitignore, because it may start conflicting with projects at some point + project-specific gitignore is like a docs for tooling that's expected to be used.

.gitignore
webknjaz: @Sonophoto JFYI those tools all use a canonical virtualenv under their hoods. After all, you don't have to use smth external if you don't want to: `venv` argument is just a folder path. So you could have a bunch of virtualenvs existing in some `~/.py-venvs/` folder and use that with a pure virtualenv. This would allow you to not keep envs in project dirs.

webknjaz: Oh, and using this approach you're not restricted to use some conventional `venv/` dir for environment anymore to match someone else's ignoring rules. You could use your own naming conventions: I, for example, include project name and python interpreter version into env name even with `pyenv` tool (which btw is able to automatically activate env once I enter project folder).

PR #1683, opened 2018-01-03
Support unicode in login/password during HTTP Auth

### LINKED ISSUE #1680: CherryPy will allways decode basic authentication information with ISO-8859-1
I have submitted the observed behavior on [stackoverflow](https://stackoverflow.com/q/48009727/406423).

```python
import cherrypy

class SimpleWebpage(object):
    @cherrypy.expose
    def index(self):
        return "<html><head></head><body>Authenticated</body></html>"

def hexcode(s):
    return ' '.join(hex(ord(x))[2:] for x in s)

def test_hexcode(realm, username, password):
    username_hexcode = hexcode(username)
    password_hexcode = hexcode(password)
    print(f"realm: {realm!r}, username: {username!r}-{username_hexcode!r}, "
          f"password: {password!r}-{password_hexcode!r}")
    return False

cherrypy.tree.mount(SimpleWebpage(), '/',
                    {'/': {'tools.auth_basic.checkpassword': test_hexcode,
                           'tools.auth_basic.on': True,
                           'tools.auth_basic.realm': 'MY_REALM',}})

cherrypy.config.update({'tools.sessions.on': True,})
cherrypy.server.socket_host = '0.0.0.0'
cherrypy.engine.autoreload.unsubscribe()
cherrypy.engine.start()
cherrypy.engine.block()
```

Issuing 
`curl -u 'curl:€öäü' -i -X GET http://10.25.5.17:8080/` 
will print 
`realm: 'MY_REALM', username: 'curl'-'63 75 72 6c', password: 'â\x82¬Ã¶Ã¤Ã¼'-'e2 82 ac c3 b6 c3 a4 c3 bc'` 
on the python console.

[RFC-7617](https://tools.ietf.org/html/rfc7617) explains how encoding should be handled concerning basic authentication.

Currently CherryPy doesn't offer a way to indicate what encoding scheme it wants and how it decodes it. Right now browsers will use different charsets:

On Redhat 6
- curl 7.19.7 (x86_64-redhat-linux-gnu) sends UTF-8

On Windows 10 Version 1709
- Firefox 57.0.2 (32-Bit) sends (probably) ISO-8859-1
- curl 7.56.1 (i686-pc-cygwin) sends UTF-8
- Google-Chrome 63.0.3239.108 (64-Bit) sends UTF-8
- Internet Explorer 11.786.15063.0 sends (probably)  ISO-8859-1 
- Edge 40.15063.674.0 sends (probably) ISO-8859-1

A possible solution would be if  CherryPy would send `WWW-Authenticate: Basic realm="foo", charset="UTF-8"` and decode the authentication string as utf-8 instead of ISO-8851-1. But I'm sure that some other things might need to be considered. And I haven't tested if Firefox will switch to UTF-8 with this header.

webknjaz:
Bugfix

#1680

creds are decoded using latin-1

Creds are decoded according to RFC7617, supporting UTF-8.

* **Other information**:
This should close #1680 once polished.


webknjaz:
This pull request introduces 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/cherrypy/cherrypy/rev/pr-d0d2143f52b19343c7824e00e361597379a3065f)

**new alerts:**

* 1 for Illegal raise

---

PJaros:
I've tested this PR successfully. Decoding the €-sign doesn't work with 8859-1 which is no surprise as this was introduced with 8859-15. I think it's still an interesting test-case ;)

I've the detailed findings and some test-code in this gist: https://gist.github.com/PJaros/b624e1b600d197c227b132f3c34ba3db

webknjaz:
@PJaros You are not right regarding removal of 4-5-6. I took this from the RFC: the input should be normalized using NFC.
And the reason is that Unicode allows to produce single visual symbol with several different sequences of characters. For example, `ä` could typed as a one Unicode character or a sequence of `a` and `:`, this would lead to comparison operation claiming that strings are different, while they're not (visually, for end-user). NFC will convert everything into the same type of sequences to avoid ambiguity.

PJaros:
@webknjaz Ah, thanks for the explanation concerning the steps 4, 5 and 6. I have yet much to learn about Unicode and UTF-8.

webknjaz:
Yeah, unicode is hard :)

webknjaz:
This pull request introduces 1 alert and fixes 1 - [view on lgtm.com](https://lgtm.com/projects/g/cherrypy/cherrypy/rev/pr-0814e7a508922588a39fdf62a5007bb73b340374)

**new alerts:**

* 1 for Illegal raise

**fixed alerts:**

* 1 for Potentially uninitialized local variable

---

webknjaz:
@PJaros feel free to review changes here.

PR #1684, opened 2018-01-04
Attempt to decode Name/Password from Basic Auth as UTF-8 

PJaros:
and cope with the fact that curl (tested on 7.19.7 and 7.56.1) and google-chrome 63.0.3239.108 will send 'UTF-8' encoded strings within the base64 encoded basic auth string.

Attempting to send charset in the 'www-authenticate'-header was ignored by Firefox 57.0.2 (32-Bit), curl 7.56.1, Google-Chrome 63.0.3239.108, Internet Explorer 11.786.15063.0 and Edge 40.15063.674.0. So RFC-7617 is beeing ignored as of 2018-01-01 from those browers.

Bug fix

https://github.com/cherrypy/cherrypy/issues/1680

Name/Password will allways be decoded with ISO-8859-1

This will try to decode name and password as UTF-8 if possible.

* **Other information**:

### DIFF
--- a/cherrypy/lib/auth_basic.py
+++ b/cherrypy/lib/auth_basic.py
@@ -44,6 +44,12 @@ def checkpassword_dict(user_password_dict):
     return checkpassword
 
 
+def decode_utf8(s):
+    """Convert string to bytes and try to interpret it as utf-8 encoded string."""
+    s_bytes = bytes([ord(c) for c in s])
+    return s_bytes.decode('utf-8')
+
+
 def basic_auth(realm, checkpassword, debug=False):
     """A CherryPy tool which hooks at before_handler to perform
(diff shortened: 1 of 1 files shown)


webknjaz:
I've done some testing and it turned out that Firefox, while trying to use `latin1`, corrupts the data during encoding in some characters. I've tested with `curl:€öäü` (login:password) and the data I get after decoding is different at the position of `€`, all other chars are represented as expected:
```python
ipdb> base64_decode(params)[5].encode()  # corrupted
b'\xc2\xac'
ipdb> '€'.encode()  # valid
b'\xe2\x82\xac
```
So during encoding inside of Firefox the first two bytes of `€` are just replaced with some garbage.

webknjaz:
Closing in favor of #1683

Issue #1688, opened 2018-01-13
[TODO: remove in v15] Duplicate HTTP Auth Basic and HTTP Auth Digest tools have to be merged


webknjaz:
While working on #1680 I realized that there's functional duplication in tools we provide by default.
There's [`auth_basic`, `auth_digest`](https://github.com/cherrypy/cherrypy/blame/880f946/cherrypy/_cptools.py#L538-L539) and also [`basic_auth`, `digest_auth`](https://github.com/cherrypy/cherrypy/blame/880f946/cherrypy/_cptools.py#L529-L530).
They also have duplicate tests spread across 3-5 test modules, which has to be handled as well.

#913 (0704e6ca8c3e5a6014ecc88e3265854faf9fe764) and as mentioned in #914:
> Note that with this new auth_digest and the auth_basic in ticket #913, the files lib/auth.py and lib/httpauth.py can be completely deprecated.

* **I'm submitting a ...**
- [ ] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository
- [X] [WAT](http://i0.kym-cdn.com/photos/images/newsfeed/000/173/580/Wat.jpg?1315930588)

* **Do you want to request a *feature* or report a *bug*?**
Deprecated tool removal.

* **What is the current behavior?**
There's ambiguity making end-users confused about which tool to use.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem**.
See src

* **What is the expected behavior?**
Just one well-documented and well-tested implementation for each of Basic/Digest HTTP Auth methods.

* **What is the motivation / use case for changing the behavior?**
[There should be one-- and preferably only one --obvious way to do it. (Zen of Python)](https://zen-of-python.info/there-should-be-one-and-preferably-only-one-obvious-way-to-do-it.html#13)

* **Please tell us about your environment:**

- Cheroot version: —
- CherryPy version: 3.2.0‒13.1.0+
- Python version: —
- OS: —
- Browser: —

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)
—

webknjaz:
@jaraco we've forgotten to remove this thing. Maybe we need some label for deprecations to check before doing major releases?

webknjaz:
The only thing I worry about is that new tool runs on `before_handler`, while the old one was used to be running on `on_start_resource`.

[...]

PR #1689, opened 2018-01-13
Deprecate old auth tools

jaraco:
This change is the first of two phases for #1688.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,13 @@
+v13.2.0
+-------
+
+* #1688: Officially deprecated ``basic_auth`` and ``digest_auth``
+  tools and the ``httpauth`` module, triggering DeprecationWarnings
+  if they're used. Applications should instead adapt to use the
+  more recent ``auth_basic`` and ``auth_digest`` tools.
+  This deprecated functionality will be removed in a subsequent
+  release soon.
+
 v13.1.0
 -------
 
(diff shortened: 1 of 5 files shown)


webknjaz (changes requested):
Plz don't remove tests as this will conflict with #1683. We need to move tests to valid implementation.

jaraco:
> conflict with #1683

Perhaps #1683 should disregard the deprecated modules?

> We need to move tests to valid implementation.

Is that true? Or can we simply ignore them? If someone wishes to port/implement those tests for the new, preferred tools, I welcome that, but I'd consider that outside the scope of this effort.

webknjaz:
This pull request fixes 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/cherrypy/cherrypy/rev/pr-09499e69661ded5f51b4edc04f2d73c5d8c29f80)

**fixed alerts:**

* 1 for Potentially uninitialized local variable

---

jaraco:
I've removed the commit that removed the tests.

webknjaz:
Actually, there's `DeprecationTool` in _cptools. And it looks like a right place for adding deprecation

jaraco:
Hmm. I honestly don't understand how DeprecatedTool is meant to work. In the two places it's used, it's not wrapping another tool, but merely acting as a stand-in for another tool that was removed. I think we should remove the DeprecatedTool also, because its name betrays its purpose (it's not deprecating if it's replacing) and because there's no useful example of using it to deprecate a tool.

Issue #1690, opened 2018-01-19
AntiStampede Cache leaves orphaned threading.Event object on 304 Not Modified response


mrvanes:
AntiStampede Cache leaves orphaned threading.Event object on 304 Not Modified response and results in a 30 second timeout on subsequent request.

We are not able to reliably reproduce this problem, we only know that it happens and have the logs to scrutinize the TOOLS.CACHING code path followed when hitting the bug.

We've narrowed down the bug to
- static files
- an 'If-Modified-Since' request provoking a 304 Not modified response, leaving an orphaned threading.Event, causing
- a 30 seconds timeout on subsequent (normal) request

The logs show that the 304 provoking request follows the AntistampeCache.wait path, setting the threading.Event while returning a None to MemoryCache.get() variant object, causing it to return None to get()'s cache_data, causing it to follow the 'request is not cached' path of get(), returning False, but MemoryCache.put() is never called, thus leaving the threading.Event object orphaned.

The subsequent request for the same static file also follows the AntistampeCache.wait path. Encountering the orphaned threading.Event object causing it to fruitlessly wait 30 seconds after which it resolves the problem by diligently populating the cache object with the (eventually) responded static file.

After this, normal operation is resumed.

This happens mostly once a day, probably after cache has expired, after one of the clients was the last to requests an 'If-Modified-Since' static file response. But until now, we were not able to come up with a clean reproducible test case.

CherryPy is part of the Pyff daemon we deployed.
https://github.com/leifj/pyFF/issues/116

- CherryPy version: 11.1.0
- Python version: 2.7 due to Pyff dependancy
- OS: Debian GNU/Linux 9
- Browser: Chrome 63

The logging showing the problem, which was produced by us inserted extra debugging lines looks as follows. I understand that interpreting these logs without knowledge of the location of these statements is awkward. Nevertheless it clearly shows the timeout after the offending 304'd static file request.
```
Jan 19 08:43:26 proxy2 pyffd[4100]: [19/Jan/2018:08:43:26] TOOLS.CACHING get https:/***/static/bootstrap/css/bootstrap.min.css
Jan 19 08:43:26 proxy2 pyffd[4100]: [19/Jan/2018:08:43:26] TOOLS.CACHING Wait result for key: (), type(value) <type 'NoneType'>
[...]

PR #1692, opened 2018-02-02
Fix static file cached 304 response

mrvanes:
Bugfix

#1690 

#1690

No timeouts

### DIFF
--- a/cherrypy/_cprequest.py
+++ b/cherrypy/_cprequest.py
@@ -903,6 +903,11 @@ class Response(object):
             # and 304 (not modified) responses MUST NOT
             # include a message-body."
             dict.pop(headers, 'Content-Length', None)
+            # self.body is a generator object that must be yielded
+            # for _caching.tee_output() to finish it's job and
+            # remove orphaned threading._Event object from cache
+            for i in self.body:
+                pass
             self.body = b''
         else:
             # Responses which are not streamed should have a Content-Length,
--- a/cherrypy/_cprequest.py
+++ b/cherrypy/_cprequest.py
@@ -2,8 +2,11 @@ import sys
 import time
 
 import uuid
+
 import six
 from six.moves.http_cookies import SimpleCookie, CookieError
+
+from more_itertools.recipes import consume
 
 import cherrypy
 from cherrypy._cpcompat import text_or_bytes, ntob
@@ -879,6 +882,14 @@ class Response(object):
         self.body = newbody
         return newbody
 
+    def _flush_body(self):
+        """
+        Discard self.body but consume any generator such that
+        any finalization can occur, such as is required by
+        caching.tee_output().
+        """
+        consume(iter(self.body))
+
     def finalize(self):
         """Transform headers (and cookies) into self.header_list. (Core)"""
         try:
@@ -903,11 +914,7 @@ class Response(object):
             # and 304 (not modified) responses MUST NOT
             # include a message-body."
             dict.pop(headers, 'Content-Length', None)
-            # self.body is a generator object that must be yielded
-            # for _caching.tee_output() to finish it's job and
-            # remove orphaned threading._Event object from cache
-            for i in self.body:
-                pass
+            self._flush_body()
             self.body = b''
         else:
             # Responses which are not streamed should have a Content-Length,
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v17.2.0
+-------
+
+* :issue:`1690`: Prevent orphaned Event object in cached
+  304 response.
+
 v17.1.0
 -------
 
--- a/cherrypy/lib/caching.py
+++ b/cherrypy/lib/caching.py
@@ -405,10 +405,16 @@ def tee_output():
             output.append(chunk)
             yield chunk
 
-        # save the cache data
-        body = b''.join(output)
-        cherrypy._cache.put((response.status, response.headers or {},
-                             body, response.time), len(body))
+        # save the cache data, but only if the body isn't empty
+        # e.g. an uncached 304 static file response will trigger this
+        # if we don't delete the cache, an empty 304 response will
+        # be cached.
+        body = ntob('').join(output)
+        if not body:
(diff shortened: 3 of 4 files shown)


webknjaz:
Thanks! I'll try to review this ASAP. But if not, it's okay to ping me.

cherrypy/_cprequest.py
```diff
@@ -929,6 +929,11 @@ def finalize(self):
             # and 304 (not modified) responses MUST NOT
             # include a message-body."
             dict.pop(headers, 'Content-Length', None)
+            # self.body is a generator object that must be yielded
+            # for _caching.tee_output() to finish it's job and
+            # remove orphaned threading._Event object from cache
+            for i in self.body:
```
webknjaz: This could be replaced by `self.collapse_body()`, but I'm still trying to figure out a fix on a different level.

jaraco (changes requested):
This change seems to have elicited a new failure in the Windows AppVeyor tests. I've restarted the tests for this PR twice and they fail consistently. We'll have to devise a fix for that failed test before we can accept this.

cherrypy/_cprequest.py
```diff
@@ -929,6 +939,7 @@ def finalize(self):
             # and 304 (not modified) responses MUST NOT
             # include a message-body."
             dict.pop(headers, 'Content-Length', None)
+            self._flush_body()
```
webknjaz, line 917: @jaraco do you fully agree with such approach? I'd prefer not putting the body there over consuming it here.

jaraco (changes requested):
I'm still uneasy with this PR. I don't fully understand its implications and as much as I've tried to document the changes, I don't know what to write to explain the behavior.

Furthermore, this sort of change really needs a test to capture the desired behavior. The associated ticket describes a scenario which should trigger the failure, so what needs to be done is codify that scenario in a test. I'm out of time to work on this today, but I welcome others to implement the test.

webknjaz:
@jaraco it feels like you got into my head and wrote down every single thought I had regarding this PR :)
I've been trying to go through the code and trace everything which is happening during such flow, but only succeed with a few parts of it so far. I need more time to understand it in full. It seems to me that it patches the side-effect, but should fix the root cause of the issue.
And of course regression test is 100% mandatory for such a complicated case.

mrvanes:
It took me more than a month to understand the time-outs we experienced (and that included working my way into the dungeons of cherrypy). I'm glad it takes you guys a while as well ;)

The issue I linked fully describes the bug and condition, and it's hard to explain it better. It does help to exactly follow the code as I explain, but it also helps to fully grok the CachingTool.

The problem arises around a race condition that is caused by expiring a previous normally cached static file. As the MemoryCache.expire_cache daemon thread only deletes the variant and leaves the uri in the store dictionary, MemoryCache.get() will now follow variant = uricache.wait() codepath on a next request, and AntiStampedeCache will guard the cache by inserting a threading._Event object in the cache (because it didn't find a result). Now, if this next request happens to be an 'if-modified-since' request, the static.serve_fileobj() path will except (caused by cptools.validate_since()) before the body is filled with the fileObject and set headers to 304 not-modified. Response.finalize() on it's turn used to skip the tee_output generator on this and leave the threading._Event behind, holding off the next requests untill the threading._Event timed out. Causing our observed time-outs.

However, only consuming tee_output on this 304 case would fill the cache with an empty body and repsonse headers 304. This is nonsense for someone that would request the same uri/variant without an if-modified-since header, because that client would not have the file in cache, but would be served with an empty body, containing a header (not-modified) it never asked for!

I do doubt the sanity of the coarse empty body case cache deletion in tee_output. It certainly helps to solve the observed case and doesn't harm others. The only problem is, it interferes with the caching of potentially expensive empty body results, but I have no clue how to correctly address that and well, we had to move on with our project ;)

By the way: This bug also raises my question why first requests to non-existing uri's in caching.store won't receive the AntiStampedeCache guard treatment. Of all circumstances, a freshly restarted server that has empty (memory)cache should be protected against a request stampede? Of course, I see why it happens, but I think it could use some attention as well?

mrvanes:
So, did my second attempt at explaining the race help?

webknjaz:
I didn't have time to check this attentively just yet.

mrvanes:
I suddenly realised why I had so much trouble understanding the CachingTool in the first place and suggest another fix for that: the _caching.get() function isn't a getter, so it's name is confusing. It loads the server.response if cache is available and returns true or false on success/failure to do that. I think it should be renamed load().

mrvanes:
Still no time to review the PR?

c00kiemon5ter:
Hi everyone,

I think this is fine as it is now. It is a proper fix, even if it is not the best possible solution. As this does not affect any public API or method signature, I think this should be merged. If you think there is room for improvement, lets open an issue and link it to this PR with a reference to the merge commit and work on it when time allows - but until that is done we can have the fix.

What do you think?

webknjaz:
@c00kiemon5ter Thanks for your opinion!

The main blocker for me as a maintainer is being able to understand the implementation and what are the consequences for other parts of code.

I'm really sorry that I didn't find time to dig into it, but to me it feels irresponsible to blindly merge changes. I hope you understand this.

It's still in my long backlog. As a community member, you can try helping @jaraco and me to verify things and comment on this more, so that it will make it easier for us to do reviews and would take much less time.
Normally we have some limited amount of FOSS time available and we cannot always commit to spending a lot of it, but if it starts looking easy to go through the chances that we'll look closer and maybe reach "merge stage" are much higher. Thanks you in advance for understanding!

P.S. @mrvanes could you please rebase this PR on top of current master?

webknjaz:
// Side note:
> def get(invalid_methods=('POST', 'PUT', 'DELETE'), debug=False, **kwargs):

it looks like the list should also include `PATCH`

webknjaz:
// Also side note:

https://tools.ietf.org/html/rfc7231#section-4.3.7
> Responses to the OPTIONS method are not cacheable.

jaraco (approved):

Issue #1694, opened 2018-03-01
handling defined filename encoding in Content-Disposition


pawciobiel:
* **I'm submitting a ...**
- [x] bug report

* **Do you want to request a *feature* or report a *bug*?**
I'd like to report a bug because this appears to me as undocumented and undesired behaviour,
but I understand that this could also be feature request.

* **What is the current behavior?**

When POSTing a file containing non ASCII characters in it's name, using  Content-Type: multipart/form-data
and Content-Disposition with defined encoding: 

`form-data; name="myFile"; filename*=utf-8''upload_test_file_%C5%82%C3%B3%C4%85%C3%A4.txt`

when accessing upload in POST handler like `myFile.filename`
I get *file content* instead of it's name.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem**.

```
import requests
fn = '/tmp/test_\xc5\x82\xc3\xb3\xc4\x85\xc3\xa4.po'
with open(fn, 'rb') as fh: r = requests.post(url, files={'myFile': fh})
```

* **What is the expected behavior?**

I think that the best way would be to handle encoding and return name as unicode object
but if this isn't possible, perhaps return HTTP 400 Bad Request or at least throw NotImplementedError which would be returned as HTTP 500?

```
def POST(self, myFile):
    fname = myFile.filename  # should contain python unicode
```

* **What is the motivation / use case for changing the behavior?**

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

https://tools.ietf.org/html/rfc5987

* **Please tell us about your environment:**

- Cheroot version: 6.0.0
- CherryPy version: 14.0.0
- Python version: 2.7
- OS: linux
- Browser: [requests ]

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)

https://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http

webknjaz:
Fixed by #1695

PR #1695, opened 2018-03-01
Fix _cpreqbody.Entity to handle Content-Disposition filename* with encoding

pawciobiel:
* Fix _cpreqbody.Entity to handle Content-Disposition filename* with encoding
* Change _cpcompat.unquote_qs to work with unicode on py2

fix for #1694

File uploads with
`Content-Disposition: form-data; name="myFile"; filename*=utf-8''upload_test_file_%C5%82%C3%B3%C4%85%C3%A4.txt`
should work now.


cherrypy/test/test_http.py
```diff
@@ -160,10 +180,13 @@ def test_post_filename_with_special_characters(self):
...
         was reported as a bug in:
            https://github.com/cherrypy/cherrypy/issues/1146/
-           https://github.com/cherrypy/cherrypy/issues/1397'''
+           https://github.com/cherrypy/cherrypy/issues/1397
+           https://github.com/cherrypy/cherrypy/issues/1694'''
         # We'll upload a bunch of files with differing names.
         fnames = ['boop.csv', 'foo, bar.csv', 'bar, xxxx.csv', 'file"name.csv',
-                  'file;name.csv', 'file; name.csv']
+                  'file;name.csv', 'file; name.csv',
+                  ('test_łóąä.txt', 'utf-8')
```
webknjaz: m/b `u'test_łóąä.txt'`?

webknjaz (dismissed):
Linters must not fail https://travis-ci.org/cherrypy/cherrypy/jobs/349840614#L482

cherrypy/test/test_http.py
pawciobiel: So I wasn't sure but I think it may be better to pass str not unicode object here
and this is why I put "# coding: utf-8" at the top of the file.

pawciobiel:
@webknjaz I've fixed pylint stuff. Could you give it another chance please?

cherrypy/_cpreqbody.py
```diff
@@ -470,6 +470,10 @@ def __init__(self, fp, headers, params=None, parts=None):
                     self.filename.endswith('"')
                 ):
                     self.filename = self.filename[1:-1]
+            elif 'filename*' in disp.params:
+                # @see https://tools.ietf.org/html/rfc5987
+                fn_encoding, _, filename = disp.params['filename*'].split("'")
```
webknjaz: Could you please keep a language part?

cherrypy/_cpreqbody.py
```diff
@@ -470,6 +470,10 @@ def __init__(self, fp, headers, params=None, parts=None):
...
+                self.filename = unquote_qs(filename, fn_encoding)
```
webknjaz: I'm not sure whether we'd want to have only one filename field. I think it's appropriate to save all of them.

We could have some `@property` returning just one thing though.

cherrypy/_cpcompat.py
```diff
@@ -100,11 +100,12 @@ def base64_decode(n, encoding='ISO-8859-1'):
...
-    return (
-        urllib.parse.unquote(atom_spc, encoding=encoding, errors=errors)
-        if six.PY3 else
-        urllib.parse.unquote(atom_spc).decode(encoding, errors)
-    )
+    if six.PY3:
+        return urllib.parse.unquote(atom_spc, encoding=encoding, errors=errors)
+    else:
+        if isinstance(atom_spc, six.text_type):
+            atom_spc = str(atom_spc)
```
webknjaz: Is this covered by tests?

cherrypy/test/test_http.py
```diff
@@ -20,13 +21,32 @@ def encode_multipart_formdata(files):
...
     BOUNDARY = '________ThIs_Is_tHe_bouNdaRY_$'
     L = []
     for key, filename, value in files:
         L.append('--' + BOUNDARY)
-        L.append('Content-Disposition: form-data; name="%s"; filename="%s"' %
-                 (key, filename))
+
+        filename_encoding = None
+        if isinstance(filename, (list, tuple)):
+            filename, filename_encoding = filename
```
webknjaz: There's too much logic in here. Tests should be simple and keep testing old cases.

cherrypy/test/test_http.py
```diff
@@ -160,10 +180,13 @@ def test_post_filename_with_special_characters(self):
...
            https://github.com/cherrypy/cherrypy/issues/1146/
-           https://github.com/cherrypy/cherrypy/issues/1397'''
+           https://github.com/cherrypy/cherrypy/issues/1397
+           https://github.com/cherrypy/cherrypy/issues/1694'''
         # We'll upload a bunch of files with differing names.
-        fnames = ['boop.csv', 'foo, bar.csv', 'bar, xxxx.csv', 'file"name.csv',
-                  'file;name.csv', 'file; name.csv']
+        fnames = [
+            'boop.csv', 'foo, bar.csv', 'bar, xxxx.csv', 'file"name.csv',
+            'file;name.csv', 'file; name.csv', ('test_łóąä.txt', 'utf-8')
```
webknjaz: this difference in supplied types is redundant.

webknjaz:
@jaraco could you take a look at this, please?

jaraco:
I apologize I haven't had time to review this yet. I still plan to do so.

cherrypy/_cpreqbody.py
```diff
@@ -470,6 +470,10 @@ def __init__(self, fp, headers, params=None, parts=None):
                     self.filename.endswith('"')
                 ):
                     self.filename = self.filename[1:-1]
+            elif 'filename*' in disp.params:
```
jaraco: RFC says filename* should be preferred to filename if both are specified, so this should be `if` and not `elif`.

cherrypy/_cpreqbody.py
jaraco: Hmm. That seems out of scope for this change. This change only seeks to add support for non-ascii filenames. Supporting multiple filenames can be done independently.

jaraco: Is `unquote_qs` the right operation here? I'm skeptical because it translates `+` to ` `, which I don't see in the spec. That's also a bug with the existing _cp_compat; it's doing too much already. Let's fix that first.

cherrypy/test/test_http.py
```diff
@@ -20,13 +21,32 @@ def encode_multipart_formdata(files):
...
     for key, filename, value in files:
         L.append('--' + BOUNDARY)
-        L.append('Content-Disposition: form-data; name="%s"; filename="%s"' %
-                 (key, filename))
+
+        filename_encoding = None
+        if isinstance(filename, (list, tuple)):
+            filename, filename_encoding = filename
+        if filename_encoding:
+            if six.PY3:
```
jaraco: If at all possible, tests shouldn't switch Python versions. It should be possible to write universal code. Switching on Python versions adds complexity and potential pitfalls.

cherrypy/test/test_http.py
```diff
@@ -160,10 +180,13 @@ def test_post_filename_with_special_characters(self):
         '''Testing that we can handle filenames with special characters. This
         was reported as a bug in:
            https://github.com/cherrypy/cherrypy/issues/1146/
-           https://github.com/cherrypy/cherrypy/issues/1397'''
+           https://github.com/cherrypy/cherrypy/issues/1397
+           https://github.com/cherrypy/cherrypy/issues/1694'''
```
jaraco: It would have been nice if the preceding commit had formatted without the trailing end symbol, but they didn't so now we get a messy diff. Let's not repeat that mistake and instead terminate the multiline string on another line.

cherrypy/test/test_http.py
jaraco: I don't understand the comment. The only addition is the one new filename with non-ascii characters. I don't see how that's redundant to all the other names with ascii characters.

jaraco (changes requested):

jaraco:
This is a great PR. I have just a few nitpicks... and one thing I wish to correct before this PR is accepted (in unquote_qs). I'll continue to work it from the PR branch.

cherrypy/_cpcompat.py
```diff
@@ -100,11 +100,12 @@ def base64_decode(n, encoding='ISO-8859-1'):
...
     atom_spc = atom.replace('+', ' ')
-    return (
-        urllib.parse.unquote(atom_spc, encoding=encoding, errors=errors)
-        if six.PY3 else
-        urllib.parse.unquote(atom_spc).decode(encoding, errors)
-    )
+    if six.PY3:
+        return urllib.parse.unquote(atom_spc, encoding=encoding, errors=errors)
+    else:
+        if isinstance(atom_spc, six.text_type):
```
jaraco: I'm hoping this change is unnecessary, because it complicates the behavior.

jaraco:
I've updated the _cpcompat.py with the changes I wanted to see (conflicting with this PR); tomorrow, I hope to revisit this and implement the required changes herein to address the conflict and other concerns I raised above.

jaraco (approved):

cherrypy/test/test_http.py
```diff
@@ -11,22 +12,46 @@
...
+    filename can be a string or a tuple ('filename string', 'encoding')
     """
     BOUNDARY = '________ThIs_Is_tHe_bouNdaRY_$'
     L = []
     for key, filename, value in files:
         L.append('--' + BOUNDARY)
-        L.append('Content-Disposition: form-data; name="%s"; filename="%s"' %
-                 (key, filename))
+
+        fn_key, encoded = encode_filename(filename)
```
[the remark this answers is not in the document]
jaraco: Codacy is mistaken here.

cherrypy/_cpcompat.py
```diff
@@ -105,9 +105,13 @@ def _unquote_compat(string, encoding='utf-8', errors='replace'):
...
 
 
+def _quote_compat(string, encoding='utf-8', errors='replace'):
+    return urllib.parse.quote(string.encode(encoding, errors))
+
+
 unquote_plus = urllib.parse.unquote_plus if six.PY3 else _unquote_plus_compat
 unquote = urllib.parse.unquote if six.PY3 else _unquote_compat
-
+quote = urllib.parse.quote if six.PY3 else _quote_compat
```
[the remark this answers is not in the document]
jaraco: Codacy is wrong here.

cherrypy/_cpreqbody.py
pawciobiel: So I can't remember exactly but I think the way I tested it on some older version was I POSTed a file with spaces and some UTF8 chars in it's name using `requests` and saw that `requests` is using quote.

cherrypy/test/test_http.py
webknjaz: I don't remember why I wrote it back then :man_shrugging:

cherrypy/_cpreqbody.py
webknjaz: Not if it breaks access to the old field

webknjaz: `requests` is often not a good example of following RFCs, unfortunately.
As for `+` vs `%20` for ` `, I guess it depends on how client decides to encode that. I think I saw something regarding this in query params spec in some rfc.

cherrypy/test/test_http.py
```diff
@@ -11,22 +12,53 @@
...
+        tmpl = \
```
webknjaz, line 59: please use braces instead of escaping EOL

Issue #1391, opened 2015-12-01
File session locking race


ghost:
Originally reported by: **Anonymous**

---

We encountered a race condition in the FileSession class, one that causes our application to discard session data prematurely when a non-trivial load is generated against the application.

The issue is that on UNIX locks are associated with file objects and not pathnames, so this race which is clearly marked as a race:

```python
        try:
            # Open lockfile for writing without truncation:
            self.fp = open(path, 'r+')
        except IOError:
            # If the file doesn't exist, IOError is raised; Use a+ instead.
            # Note that there may be a race here. Multiple processes
            # could fail on the r+ open and open the file a+, but only
            # one will get the the lock and write a pid.
            self.fp = open(path, 'a+')
```

Can generate actually two different file objects, one that is overwritten by the other, but which still has an allocated space. These are marked as (deleted) in the lsof output.

With these two handles in place in two separate threads, each will successfully acquire a lock but in reality the critical section will run in parallel.

I quick workaround is to avoid removing the file altogether, so only the initial connection will have the race:

```diff
--- a/cherrypy/lib/sessions.py    2015-11-30 14:00:08.011323375 +0100
+++ b/cherrypy/lib/sessions.py    2015-11-30 12:41:37.864640838 +0100
@@ -554,7 +554,7 @@

     def release_lock(self, path=None):
         """Release the lock on the currently-loaded session data."""
-        self.lock.remove()
+        #self.lock.remove()
         self.lock.release()
         self.locked = False
```

This solves the race because the lock file will never be removed and thus only the first branch will execute in LockFile constructor, causing all racing threads to open the already existing files. A small race in the initial request would still happen, if the lock file does not yet exist.

I see various alternatives to resolve this problem:
- actually place the lock on the session file instead of a separate lockfile
- always create the lockfile and never remove it
- after locking recheck the pid in the file and see if we lost the race (after locking, we could read it after reopening the file by pathname, that would return the file object that won the race), in that case we could release our stale fd and relock the other one

But I do have a couple of issues how this lock works anyway:
- we are writing a pid, but in a threaded application it is always the same (ever since NPTL took over linuxthreads, so 7-8 years at least on Linux)
- I don't see too much advantage in writing the pid in the first place, it generates I/O on every request (ok I know it is probably cached, but anyway), since it's always the same it doesn't give the user any clues what is blocking the session
[...]

PR #1705, opened 2018-04-19
Fix for concurrent file session access

### LINKED ISSUE #1193: sessions lock race condition deleting expired session
Originally reported by: **deecodameeko (Bitbucket: [deecodameeko](http://bitbucket.org/deecodameeko), GitHub: @deecodameeko)**

---

I found an issue last night with sessions where it tries to release a lock on a session that isn't locked.

Here's a dump of the logs:

```
#!python
File "/root/.virtualenvs/web_api/lib/python2.7/site-packages/cherrypy/_cprequest.py", line 102, in run
    hook()
  File "/root/.virtualenvs/web_api/lib/python2.7/site-packages/cherrypy/_cprequest.py", line 62, in __call__
    return self.callback(**self.kwargs)
  File "/root/.virtualenvs/web_api/lib/python2.7/site-packages/cherrypy/lib/sessions.py", line 706, in close
    sess.release_lock()
  File "/root/.virtualenvs/web_api/lib/python2.7/site-packages/cherrypy/lib/sessions.py", line 479, in release_lock
    os.unlink(path + self.LOCK_SUFFIX)
OSError: [Errno 2] No such file or directory: '/san/in/api-sessions/session-b93639d62bec1e4a766e244d1a22d0bf7ebdc32c.lock'

```

It seems it doesn't check if the file exists before unlinking it. Perhaps it should? I'm not sure what could cause this issue in the first place. 

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1193

BlackZork:
BugFix

fixes #1193

First you need two cherrypy processes (Tom and Jerry) with sessions stored on shared filesystem. Application has two exposed handlers: `ShortAction` and `LongAction`.

Here is an description what happens:

1. Tom started to service `ShortAction` and it aquires lock on session.
2. Jerry started to service `LongAction` and tries to aquire lock on session.
3. `fcntl.flock(self.fp.fileno(), flags)` fails for Jerry and `UnixLockFile` raises `LockError`
4. Jerry catches it in line 79 and tries to seek and close `fp` in line 82.
5. Tom in meanwhile finished servicing `ShortAction` and deleted lock file.
6. Jerry calls `self.fp.close` and `FileNotFoundError` is raised.

I catch `FileNotFoundError` and throw `LockError` instead. Jerry then tries to aquire lock on session again.

* **Other information**:
Tested and debugged on Python 3.6, Cherrypy 14.1, Linux.

### DIFF
--- a/cherrypy/lib/lockfile.py
+++ b/cherrypy/lib/lockfile.py
@@ -77,8 +77,11 @@ class SystemLockFile(object):
         try:
             self._lock_file()
         except Exception:
-            self.fp.seek(1)
-            self.fp.close()
+            try:
+                self.fp.seek(1)
+                self.fp.close()
+            except IOError:
+                pass
             del self.fp
             raise
 


jaraco:
It looks to me like the lockfile functionality was borrowed from the [zc.lockfile project](https://github.com/zopefoundation/zc.lockfile/blob/master/src/zc/lockfile/__init__.py), and the two implementations seem to be slightly different.

I struggle to think what's the right thing to do here. I notice that zc.lockfile doesn't suppress an IOError but it doesn't `seek(1)` either.

So I'm a little uncomfortable accepting this change. Since there are no tests nor comments describing the caught exception, it could easily be optimized away. Moreover, I suspect the zc.lockfile implementation is robust enough. Could we use it (by synchronizing with their implementation)? Does that solve the issue, or does the issue also exist in that library?

BlackZork:
I do not have answer to any of your questions - I have never used zc.lockfile module.

I should mention in my description that it occurs when sessions are stored on glusterfs filesystem. This PR fixes one of problems that I was able to identify and describe (so at least it is possible to create unit test for it). It is also reported in #1391.

webknjaz:
+1 for just depending on `zc.lockfile` and letting its developers maintain it. @BlackZork could you please check that replacing vendored implementation with external dependency fixes your case?

Also, it would be easier for us to validate things if there was some example reproducer (tiny cherrypy app) with instruction demonstrating the issue (this could also be a docker image). Please also try adding tests for validating your case, so that we could prevent future regressions.

webknjaz:
Also, let's ping @bazsi, @deecodameeko, and @cyraxjoe as they were active in the mentioned issues.

bazsi:
My bug is #1391 does not seem to be fixed by this patch and I would similarly be reluctant to accept it. 

In my case, the situation is quite clearly described in my ticket, basically file based locking is not actually locking the session object at all. I did and still consider this a critical issue, however I am not working on the codebase any more, where this was an issue, thus I am unable to help.

jaraco:
@BlackZork Can you take a look at #1729 and see if it addresses the error in your environment?

BlackZork:
No, on glusterfs this problem still exists.

webknjaz:
@BlackZork could you please change your patch according to the new circumstances? It might go to `cherrypy.lib.sessions` or `zc.lockfile` upstream directly.

Issue #1370, opened 2015-06-23
Fails with wrong headers


bb-migration:
Originally reported by: **Daniel Frischhut (Bitbucket: [daniel_frischhut](http://bitbucket.org/daniel_frischhut), GitHub: Unknown)**

---

We got a request with the following header-fields:

```json
  'CONNECTION': 'close'
  'CACHE-CONTROL': 'no-cache'
  'ACCEPT': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'
  'ACCEPT-CHARSET': 'ISO-8859-1,utf-8;q=0.7,*;q=0.7)'
  'USER-AGENT': 'Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.95 Safari/537.36'
  'ACCEPT-LANGUAGE': 'en-us'
  'ACCEPT-ENCODING': 'gzip,deflate'
```

and cherrypy faild with follwing Traceback:

```python-traceback
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/cherrypy/_cprequest.py", line 670, in respond
    response.body = self.handler()
  File "/usr/local/lib/python2.7/dist-packages/cherrypy/lib/encoding.py", line 260, in __call__
    ct.params[\'charset\'] = self.find_acceptable_charset()
  File "/usr/local/lib/python2.7/dist-packages/cherrypy/lib/encoding.py", line 146, in find_acceptable_charset
    encs = request.headers.elements(\'Accept-Charset\')
  File "/usr/local/lib/python2.7/dist-packages/cherrypy/lib/httputil.py", line 456, in elements
    return header_elements(key, value)
  File "/usr/local/lib/python2.7/dist-packages/cherrypy/lib/httputil.py", line 237, in header_elements
    return list(reversed(sorted(result)))
  File "/usr/local/lib/python2.7/dist-packages/cherrypy/lib/httputil.py", line 217, in __lt__
    if self.qvalue == other.qvalue:
  File "/usr/local/lib/python2.7/dist-packages/cherrypy/lib/httputil.py", line 207, in qvalue
    return float(val)
ValueError: invalid literal for float(): 0.7)
```

i edited the httputil.py in line 207 and surrounded the return float(val) with a try - except and returned the default value.

---
- Bitbucket: https://bitbucket.org/cherrypy/cherrypy/issue/1370

[...]

PR #1707, opened 2018-04-20
Fix 1370

seils:
Bug fix.

#1370 

Client receives HTTP 500 error; ValueError exception is thrown in cherrypy.

Client receives HTTP 400 Bad Request.

* **Other information**:
Includes test case.

### DIFF
--- a/cherrypy/lib/httputil.py
+++ b/cherrypy/lib/httputil.py
@@ -18,6 +18,7 @@ import six
 from six.moves import range, builtins
 from six.moves.BaseHTTPServer import BaseHTTPRequestHandler
 
+import cherrypy
 from cherrypy._cpcompat import ntob, ntou
 from cherrypy._cpcompat import text_or_bytes
 from cherrypy._cpcompat import unquote_qs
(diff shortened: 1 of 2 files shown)


webknjaz (changes requested):
This patch is too specific. It should catch different headers malformed in different ways. I'd suggest wrapping for-loop in header_elements with this try/except instead to fix this on a broader level.

webknjaz:
And I'd probably want a test in https://github.com/cherrypy/cherrypy/blob/master/cherrypy/test/test_httputil.py as well

cherrypy/lib/httputil.py
```diff
@@ -202,7 +203,12 @@ def qvalue(self):
         val = self.params.get('q', '1')
         if isinstance(val, HeaderElement):
             val = val.value
-        return float(val)
+        try:
+            return float(val)
+        # Fail client requests with invalid quality value
+        # https://github.com/cherrypy/cherrypy/issues/1370
+        except ValueError:
+            raise cherrypy.HTTPError(400)
```
webknjaz, line 211: I think, I'd like to see 'Malformed HTTP header' as a reason argument of the exception.

seils: Thanks for the quick feedback, @webknjaz.  I'll start working on a more general purpose solution.

Issue #1377, opened 2015-07-19
disabling wsgi interface fails under python3


ghost:
Originally reported by: **Tim Miller (Bitbucket: [lashni](http://bitbucket.org/lashni), GitHub: Unknown)**

---

http://docs.cherrypy.org/en/latest/advanced.html#no-need-for-the-wsgi-interface

Error is encountered when the example code from the link above is run under a python 3.4.3 virtualenv on archlinux, traceback triggers when the server is accessed. Error doesn't occur with python 2.7.10.

Happens with both current head and CherryPy 3.8.0.

```
[19/Jul/2015:17:41:01] ENGINE Listening for SIGUSR1.
[19/Jul/2015:17:41:01] ENGINE Listening for SIGTERM.
[19/Jul/2015:17:41:01] ENGINE Listening for SIGHUP.
[19/Jul/2015:17:41:01] ENGINE Bus STARTING
CherryPy Checker:
The Application mounted at '' has an empty config.

[19/Jul/2015:17:41:01] ENGINE Started monitor thread '_TimeoutMonitor'.
[19/Jul/2015:17:41:01] ENGINE Started monitor thread 'Autoreloader'.
[19/Jul/2015:17:41:01] ENGINE Serving on http://127.0.0.1:8080
[19/Jul/2015:17:41:01] ENGINE Bus STARTED
[19/Jul/2015:17:41:05] NATIVE_ADAPTER Traceback (most recent call last):
  File "/home/lashni/dev/serve/lib/python3.4/site-packages/cherrypy/_cpnative_server.py", line 27, in respond
    sn = cherrypy.tree.script_name(req.uri or "/")
  File "/home/lashni/dev/serve/lib/python3.4/site-packages/cherrypy/_cptree.py", line 257, in script_name
    path = path[:path.rfind("/")]
TypeError: 'str' does not support the buffer interface

ValueError('invalid literal for int() with base 10: "b\'5"',)
Traceback (most recent call last):
  File "/home/lashni/dev/serve/lib/python3.4/site-packages/cherrypy/_cpnative_server.py", line 27, in respond
    sn = cherrypy.tree.script_name(req.uri or "/")
  File "/home/lashni/dev/serve/lib/python3.4/site-packages/cherrypy/_cptree.py", line 257, in script_name
    path = path[:path.rfind("/")]
TypeError: 'str' does not support the buffer interface

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #1712, opened 2018-05-23
Fix native HTTP server compatibility

webknjaz:
  - [X] bug fix
  - [x] feature
  - [x] docs update
  - [x] refactoring
  - [x] other

#1377

Various exceptions related to str/bytes py2/3 compat are being raised.

Will be working

* **Other information**:
Still WIP, don't merge.
I hope to get some help from @jaraco.

* **Checklist**:

### DIFF
--- a/cherrypy/_cpnative_server.py
+++ b/cherrypy/_cpnative_server.py
@@ -95,7 +95,7 @@ class NativeGateway(cheroot.server.Gateway):
         req = self.req
 
         # Set response status
-        req.status = str(status or '500 Server Error')
+        req.status = status or b'500 Server Error'
 
         # Set response headers
         for header, value in headers:
--- a/cherrypy/test/test_native.py
+++ b/cherrypy/test/test_native.py
@@ -1,0 +1,29 @@
+"""Docstring."""
+
+import pytest
+from requests_toolbelt import sessions
+
+import cherrypy._cpnative_server
+
+
+@pytest.fixture
+def cp_native_server(request):
+    """A native server."""
+    class Root(object):
+        @cherrypy.expose
+        def index(self):
+            return 'Hello World!'
+
+    cls = cherrypy._cpnative_server.CPHTTPServer
+    cherrypy.server.httpserver = cls(cherrypy.server)
+
+    cherrypy.tree.mount(Root(), '/')
+    cherrypy.engine.start()
+    request.addfinalizer(cherrypy.engine.stop)
+    url = 'http://localhost:{cherrypy.server.socket_port}'.format(**globals())
+    return sessions.BaseUrlSession(url)
+
+
+def test_basic_request(cp_native_server):
+    """A request to a native server should succeed."""
+    cp_native_server.get('/')
--- a/cherrypy/test/test_native.py
+++ b/cherrypy/test/test_native.py
@@ -6,6 +6,12 @@ from requests_toolbelt import sessions
 import cherrypy._cpnative_server
 
 
+pytestmark = pytest.mark.skipif(
+    'sys.platform == "win32"',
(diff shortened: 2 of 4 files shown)


cherrypy/_cpnative_server.py
```diff
@@ -7,6 +7,7 @@
 import cheroot.server
 
 import cherrypy
+from cherrypy._cpcompat import tonative
```
jaraco: I consider `tonative` to be deprecated. I'd like to avoid using it if possible and instead use either text or bytes explicitly on all Pythons.

jaraco (commented):
When I rebase your changes onto the bugfix/1377-add-tests branch, I notice that b936d6e0 fixes the failing test, suggesting that the second commit, 4c6b8ff7843933fe06579f00b62d1f581b5accb2, isn't necessary to fix the bug. Is there an additional bug that isn't caught by those tests, or can this second commit be omitted?

jaraco (approved):
I've stripped out the commit that seems to be unnecessary, including the one that fixes the reported issue. I believe this PR is ready to go.

webknjaz:
@jaraco did it cause AppVeyor failures?

cherrypy/_cpnative_server.py
```diff
@@ -95,7 +95,7 @@ def send_response(self, status, headers, body):
         req = self.req
 
         # Set response status
-        req.status = str(status or '500 Server Error')
+        req.status = status or b'500 Server Error'
```
webknjaz, line 98: @jaraco there's ` or ''` over the place defaulting to bytes/unicode all over the place. Shall we try `from __future__ import unicode_strings`?

cherrypy/test/test_native.py
```diff
@@ -0,0 +1,29 @@
+"""Docstring."""
```
webknjaz: @jaraco I guess we could explain what's going on in this test suite here.

cherrypy/_cpnative_server.py
jaraco: I'd very much like to do that. It was a feature that wasn't available in the early implementations (Python 2.3 compatibility and all), and even some people still advise against its use. It's my opinion, however, that using `from __future__ import unicode_literals` is the best way to achieve compatibility. Let's do it as time permits.

jaraco:
I do see appveyor failures; the failures seem unrelated, though I can't tell for sure. Is the startup/teardown of the native server causing conflicts with other tests?

webknjaz:
I have no idea about the conflicts, but there's clearly some race conditions involved.

cherrypy/_cpnative_server.py
webknjaz: My concern is that from code it's not always obvious what kind of data the variable is expected to hold. Should we introduce some convention of prefixing `b_` or suffixing `_b` to improve the situation?

cherrypy/test/test_native.py
```diff
@@ -0,0 +1,35 @@
...
+    cherrypy.server.httpserver = cls(cherrypy.server)
+
+    cherrypy.tree.mount(Root(), '/')
+    cherrypy.engine.start()
+    request.addfinalizer(cherrypy.engine.stop)
+    url = 'http://localhost:{cherrypy.server.socket_port}'.format(**globals())
+    return sessions.BaseUrlSession(url)
+
+
+def test_basic_request(cp_native_server):
```
webknjaz, line 33: @jaraco it looks like this test doesn't catch the bug for some reason.

Ref:
* https://github.com/cherrypy/cherrypy/issues/1377#issuecomment-496252972
* https://github.com/cherrypy/cherrypy/pull/1784#issuecomment-504704035

PR #1715, opened 2018-06-11
cpstats: make 'Response Status' serializable

sanitec:
Fixes an issue where it wasn't possible to fetch /cpstats/data when
running the cpstats tool with Python3.

Could potentially also solve the problem by supplying a default
function for 'json.dumps', but this solution seems cleaner.

### DIFF
--- a/cherrypy/lib/cpstats.py
+++ b/cherrypy/lib/cpstats.py
@@ -196,7 +196,7 @@ import time
 import six
 
 import cherrypy
-from cherrypy._cpcompat import json
+from cherrypy._cpcompat import json, tonative
 
 # ------------------------------- Statistics -------------------------------- #
 
@@ -308,6 +308,14 @@ def _get_threading_ident():
     return threading._get_ident()
 
 
+def _get_response_status(resp):
+    try:
+        response_status = resp.output_status
+    except AttributeError:
+        response_status = resp.status
+    return tonative(response_status, 'utf-8')
+
+
 class StatsTool(cherrypy.Tool):
 
     """Record various information about the current request."""
@@ -366,8 +374,7 @@ class StatsTool(cherrypy.Tool):
             w['Bytes Written'] = cl
             appstats['Total Bytes Written'] += cl
 
-        w['Response Status'] = getattr(
-            resp, 'output_status', None) or resp.status
+        w['Response Status'] = _get_response_status(resp)
 
         w['End Time'] = time.time()
         p = w['End Time'] - w['Start Time']
--- a/cherrypy/test/test_tools.py
+++ b/cherrypy/test/test_tools.py
@@ -8,6 +8,8 @@ import types
 import unittest
 import operator
 
+import pytest
+import requests
 import six
 from six.moves import range, map
 from six.moves.http_client import IncompleteRead
@@ -15,6 +17,7 @@ from six.moves.http_client import IncompleteRead
 import cherrypy
 from cherrypy import tools
 from cherrypy._cpcompat import ntou
+from cherrypy.lib import cpstats
 from cherrypy.test import helper, _test_decorators
 
 
@@ -466,3 +469,21 @@ class TestHooks:
         by_priority = operator.attrgetter('priority')
         priorities = list(map(by_priority, hooks))
         assert priorities == [48, 49, 50]
+
+
+@pytest.fixture
+def cpstats_server():
+    cherrypy.tree.mount(
+        cpstats.StatsPage(), '/', {'/': {'tools.cpstats.on': True}})
+    cherrypy.engine.start()
+    cherrypy.engine.wait(cherrypy.engine.states.STARTED)
+    yield 'http://localhost:8080'
+    cherrypy.engine.exit()
+    cherrypy.engine.block()
+    cherrypy.server.httpserver = None
+
+
+def test_get_cpstats_data(cpstats_server):
+    resp = requests.get(cpstats_server + '/data')
+    assert resp.status_code == 200
+    assert 'application/json' in resp.headers['Content-Type']


cherrypy/lib/cpstats.py
```diff
@@ -308,6 +308,13 @@ def _get_threading_ident():
     return threading._get_ident()
 
 
+def _get_response_status(resp):
+    response_status = getattr(resp, 'output_status', None) or resp.status
```
webknjaz: it's probably `response_status = getattr(resp, 'output_status', resp.status)`, though I'd prefer try/except approach

cherrypy/lib/cpstats.py
```diff
@@ -308,6 +308,13 @@ def _get_threading_ident():
...
+    if sys.version_info > (3,):
```
webknjaz: we use `six.PY3` for this.

webknjaz: But it's not needed, because it can be replaced by the function, which encapsulates this check better.

cherrypy/lib/cpstats.py
```diff
@@ -308,6 +308,13 @@ def _get_threading_ident():
...
+        return str(response_status, 'utf-8')
```
webknjaz:
```python
from cherrypy._cpcompat import tonative
...
return tonative(response_status, 'utf-8')
```

webknjaz:
@sanitec would you mind rebasing and adding a test for this?

sanitec:
Sure, where would it be natural to put such a test, and what should it test? I couldn't find any testcases relevant for cpstats.

webknjaz:
Since it's a tool, I'd go for https://github.com/cherrypy/cherrypy/blob/master/cherrypy/test/test_tools.py

Honestly, we're trying to get rid of `unittest` in favor or `pytest`, but still stuck with it. If you want, you can try to avoid inheriting `unittest.TestCase`, but it's not mandatory for now.

cherrypy/lib/cpstats.py
```diff
@@ -308,6 +308,14 @@ def _get_threading_ident():
     return threading._get_ident()
 
 
+def _get_response_status(resp):
+    try:
```
jaraco, line 312: I prefer the original syntax:

```
response_status = getattr(resp, 'output_status', None) or resp.status
```

It achieves the same thing in one line instead of four, but more importantly, it causes `response_status` to be assigned in exactly one place.

Since this PR isn't about addressing the code style, better to focus on changing just the one thing.

cherrypy/lib/cpstats.py
```diff
@@ -308,6 +308,14 @@ def _get_threading_ident():
...
+        response_status = resp.output_status
+    except AttributeError:
+        response_status = resp.status
+    return tonative(response_status, 'utf-8')
```
jaraco, line 316: `tonative` is deprecated. Let's not use it here. Let's be more precise about what should be expected from response_status and what's needed by `w['Response Status']` and convert to that form. Could this be `return response_status.decode('utf-8')` instead? What reason do we have to believe the value would have been encoded as utf-8?

jaraco (changes requested):
I agree this change is the right one - rather than to have a workaround at the encoding layer. I think the solution could be more precise about what's expected/allowed for the inputs/outputs.

I know a lot of cherrypy uses `tonative` and other functions in the compat module, but those functions encourage a continued bimorphism for strings (bytes on Python 2 and unicode on Python 3). Instead, the code should move toward a unified approach (always unicode or bytes regardless of Python). Let's honor that intention in this change and not perpetuate the status quo. Thanks.

jaraco:
Another thing - could you provide an example server/script that someone like myself could use to replicate the issue?

cherrypy/lib/cpstats.py
webknjaz: Let's do `response_status = getattr(resp, 'output_status', resp.status)` then.

cherrypy/lib/cpstats.py
webknjaz: *unicode is hard*

webknjaz:
> Let's honor that intention in this change and not perpetuate the status quo.

I totally agree with this, *but* the problem is that `from __future__ import unicode_strings` would require other changes in the module and I would like to keep a single logical change per PR principle here. Such changes should go to a separate refactoring PR.

sanitec:
@jaraco to replicate:

Run (in a python3 environment):
```python3
import cherrypy
from cherrypy.lib import cpstats


if __name__ == '__main__':
    cherrypy.quickstart(
        cpstats.StatsPage(),
        config={
            '/': {'tools.cpstats.on': True}
        })
```
And then:
`curl http://localhost:8080/data`

sanitec:
Hmm, seems like it doesn't fix the problem entirely. When running that example with the fix I get:
```python
  File "/Users/sfosteru/src/cherrypy/cherrypy/_cprequest.py", line 627, in respond
    self._do_respond(path_info)
  File "/Users/sfosteru/src/cherrypy/cherrypy/_cprequest.py", line 686, in _do_respond
    response.body = self.handler()
  File "/Users/sfosteru/src/cherrypy/cherrypy/_cprequest.py", line 803, in __set__
    raise ValueError(self.unicode_err)
ValueError: Page handlers MUST return bytes. Use tools.encode if you wish to return unicode.
```

webknjaz:
@sanitec could you please also add a test for this case under `cherrypy/test/`?

cherrypy/test/test_tools.py
```diff
@@ -466,3 +469,21 @@ def test_priorities(self):
         by_priority = operator.attrgetter('priority')
         priorities = list(map(by_priority, hooks))
         assert priorities == [48, 49, 50]
+
+
+@pytest.fixture
```
webknjaz, line 474: :+1:

sanitec:
@webknjaz I'm not really sure where to go from here.

cherrypy/lib/cpstats.py
jaraco: If `resp.output_status` can be `None` (or other Falsey value), then you want `getattr(resp, 'output_status') or resp.status`.

cherrypy/test/test_tools.py
```diff
@@ -8,13 +8,16 @@
 import unittest
 import operator
 
+import pytest
+import requests
```
webknjaz, line 12: We probably don't want to introduce this dependency right now.

jaraco: More important than the dependency is the pattern we use to test requests against the server. These tests should follow the same pattern as other tests in the suite. I believe the way these sorts of tests are run elsewhere is the server is mounted but not listening on any port and the requests are invoked through the WebTest object. I believe we've done some work to decouple that webtest concept from unittest.

jaraco:
> the problem is that `from __future__ import unicode_strings` would require other changes in the module

I didn't think I was suggesting converting unicode literals for the module - only always returning unicode for that one function.

webknjaz:
Oh, makes sense then.

cherrypy/lib/cpstats.py
jaraco: I guess let's try webknjaz' suggestion and see if that works.

cherrypy/lib/cpstats.py
jaraco: I don't have any reason to think that tonative is required here at all, so I'm going to try removing it.

cherrypy/test/test_tools.py
jaraco: I'm going to remove these tests and defer testing this stats tool until someone with more familiarity with cherrypy is interested in writing such tests. Feel free to write an issue to capture it.

Issue #1716, opened 2018-06-12
14.2.0's decoding auth_digest parameters breaks authentication if "uri" contains encoded slashes 


teoric:
* **I'm submitting a ...**
- [x] bug report

* **What is the current behavior?**

Starting with 14.2, parameters are decoded in `cherrypy/lib/auth_digest.py`:`HttpDigestAuthorization`'s constructor.

One of the parameters is `uri`. If `uri`contains e.g. a slash, I get an authentication loop.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**

- start up application
- enter an URL, e.g. `testhost:8080/test_page?from=%2F`
- enjoy the loop

* **What is the expected behavior?**

- serving the page

* **Please tell us about your environment:**

- Cheroot version: 6.3.1
- CherryPy version: 14.2.0 or 15.0.0
- Python version: 3.6 and 3.5
- OS: Mac OSX and Linux
- Browser: at_least([Chrome  | Firefox])

(PR to follow.)

webknjaz:
cc @jaraco

jaraco:
I don't yet understand this bug nor why the PR addresses it.

Why is URI special? How is it that the redirect loop is occurring? How is it that not unquoting the uri parameter prevents that?

teoric:
I did not dig too deeply in the internals, but decoding the uri with [`unquote_to_bytes`](https://docs.python.org/3/library/urllib.parse.html) will undo hexadecimal escaping.  As far as I can see, URI is the only parameter that has hex escapes (at least in my setup).  Now assume that uri contains a slash in a parameter (`/`); by unquoting the slash is is now part of the uri. (from `https://testhost:8080/test_page?from=%2F` to `https://testhost:8080/test_page?from=/`)  I suppose this confuses some part of the program (routing?), which may rely on having only literal slashes that separate URI components, and only escaped slashes in the URI components?

Another question is: Why is decoding for URI needed at this stage now? If I understand correctly, these lines should take care of parameters in non-ASCII; but [will the URI ever be non-ASCII?](https://stackoverflow.com/questions/2504564/can-http-uris-have-non-ascii-characters)?  If not, leaving it alone is the better path anyway.

Sorry for the guessing, but I can tell you that this was the only way I found to get authentication working again after dda336459308b7d6f1a6c11cf7129e8779a0fae2.

jaraco:
[...]

PR #1717, opened 2018-06-12
Fix URI decoding for auth_digest

teoric:
decoding URIs in auth header breaks e.g. if it contains slashes

  - [x] bug fix

#1716

Authentication loop, see #1716

Authentication, as before 14.2

* **Other information**:

* **Checklist**:

  - [x] I think the code is trivial

### DIFF
--- a/cherrypy/lib/auth_digest.py
+++ b/cherrypy/lib/auth_digest.py
@@ -24,7 +24,6 @@ import time
 import functools
 from hashlib import md5
 
-from six.moves.urllib.parse import unquote_to_bytes
 from six.moves.urllib.request import parse_http_list, parse_keqv_list
 
 import cherrypy
@@ -142,15 +141,12 @@ def H(s):
     return md5_hex(s)
 
 
-def _try_decode_map_values(param_map, charset):
+def _try_decode_header(header, charset):
     global FALLBACK_CHARSET
 
     for enc in (charset, FALLBACK_CHARSET):
         try:
-            return {
-                k: tonative(v, enc)
-                for k, v in param_map.items()
-            }
+            return tonative(ntob(tonative(header, 'latin1'), 'latin1'), enc)
         except ValueError as ve:
             last_err = ve
     else:
@@ -183,18 +179,13 @@ class HttpDigestAuthorization(object):
         if not self.matches(auth_header):
             raise ValueError('Authorization scheme is not "Digest"')
 
-        scheme, params = auth_header.split(' ', 1)
+        self.auth_header = _try_decode_header(auth_header, accept_charset)
 
-        self.auth_header = auth_header
+        scheme, params = self.auth_header.split(' ', 1)
 
         # make a dict of the params
         items = parse_http_list(params)
         paramsd = parse_keqv_list(items)
-        paramsd = {
-            k: unquote_to_bytes(v) if k != "uri" else v
-            for k, v in paramsd.items()
-        }
-        paramsd = _try_decode_map_values(paramsd, accept_charset)
 
         self.realm = paramsd.get('realm')
         self.username = paramsd.get('username')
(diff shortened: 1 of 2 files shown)


jaraco:
I did some digging into the changes for 14.2, but according to the PR and the changelog, that would affect basic auth, not digest auth. How is that change implicated?

teoric:
Please consider dda336459308b7d6f1a6c11cf7129e8779a0fae2 (_Add support for UTF-8 in **auth_digest** tool_), which introduced the change I am talking about.  I suppose the Changelog is just not up complete.  In the PR #1683, dda336459308b7d6f1a6c11cf7129e8779a0fae2 and three other changes to auth_digest [are listed](https://github.com/cherrypy/cherrypy/pull/1683#commits-pushed-5bc6fe5) among the changes on 22nd April.

jaraco:
Thanks for the clarification. Now that I understand it a _bit_ better, I'm pretty sure this change isn't the right one. I'm not sure what the right change is, though. Let's continue the discussion in the ticket.

webknjaz:
+1 I also felt like this is a weird place for trying to fix such issue.

teoric:
Well, if it fixes a regression, would it not be sensible to accept a hacky fix first and search for the one true way later?

webknjaz:
@teoric,

I'm really grateful that you spend efforts trying to figure this out and help us. We are short on maintainers and are really lucky to have such people like you contributing to the project and making it better!

This change, however, doesn't fix a regression, it monkey-patches what's related/unique to *your* concrete application, not CherryPy. And it patches the wrong place meaning that something in someone else's workflow gets broken. It's dangerous/harmful to throw in patches like this (even with no tests covering this case), this will likely cause another regression, which would be much harder to identify and eliminte.

*As maintainers,* @jaraco and I are responsible for the changes we accept and we do our best to avoid mistakes, which might be hard to fix in future. We have to weight every decision we make to ensure we don't introduce sudden regressions, don't break pre-decided design considerations, keep the code base in a good shape.

I hope that this issue gets fixed soon and as a quick workaround for your app I may suggest you to copy-paste this tool into your app, do whatever temporary patches necessary and use it instead of a built-in tool and then switch back to native tool once the fix is released.

Have a nice day and again thank you for your contributions!

webknjaz:
I've added changes, which should completely eliminate the issue and take into account the format in which browsers send unicode digest auth

teoric:
Thank you very much for unmonkeying and fixing this!

webknjaz:
Could you please check this across browsers, at least FireFox, because I don't have one on the current machine.
Also, I'm going to wait until @jaraco reviews this again to double-check changes.

teoric:
Works for me on Firefox and Chrome; thank you!

webknjaz:
Great! Let's wait till @jaraco validates it as well then :)

cherrypy/lib/auth_digest.py
```diff
@@ -142,15 +141,12 @@ def H(s):
...
+def _try_decode_header(header, charset):
     global FALLBACK_CHARSET
 
     for enc in (charset, FALLBACK_CHARSET):
         try:
-            return {
-                k: tonative(v, enc)
-                for k, v in param_map.items()
-            }
+            return tonative(ntob(tonative(header, 'latin1'), 'latin1'), enc)
```
jaraco, line 149: This code is clearly not optimal or intuitive... and int introduces two additional deprecated calls. I want to clean this up.

jaraco (approved):
Although I'm not happy with the decode routine, since this addresses the issue, I'm happy to accept it to address the regression.

cherrypy/lib/auth_digest.py
webknjaz: Yeah, I tried that. But it's a can of worms, there's too many moving parts. Let's do refactoring separately from the bug fix, please.

PR #1718, opened 2018-06-16
Feature/auth refactor

jaraco:
  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [X] refactoring
  - [ ] other

`auth_basic` uses deprecated `ntob` and `base64_decode`, causing unnecessary steps of encoding/decoding.

Use the natively-available `bytes` objects directly.

* **Other information**:

* **Checklist**:

### DIFF
--- a/cherrypy/_cpcompat.py
+++ b/cherrypy/_cpcompat.py
@@ -21,7 +21,6 @@ and their .encode/.decode methods as needed.
 import re
 import sys
 import threading
-import base64
 
 import six
 from six.moves import urllib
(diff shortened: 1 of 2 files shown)


webknjaz:
@jaraco I feel like this needs additional unit tests

jaraco:
> I feel like this needs additional unit tests.

I'm unsure for what - this change doesn't introduce or remove functionality. So any tests that were already exercising this functionality are still present. Are you thinking those tests are insufficient to protect from regression around this change? I'm pretty sure when I was implementing these changes, I verified that the code was covered by tests. We certainly don't need tests to capture "basic_auth doesn't call ntob unnecessarily". Are you thinking the private extracted function demands additional unit tests?

webknjaz:
It's just a feeling after I saw an issue created a few days ago, which claims that there's a regression related to auth.

I think that private function could be tested, yes. Because while it's nice and elegant an current point of time, it might catch regression if someone modifies that in future.

jaraco:
I've several times now started writing tests for this change, but I've run into a couple of roadblocks. The first was that neither test_auth_basic nor auth_basic support unicode literals, which makes it harder to write robust tests, so I set out to create new module. I even started creating a new test suite just to test this function.

But then as I started writing tests, I realized that these tests are enforcing implementation details - internal interfaces that should be free to change.

I'm now convinced this change doesn't need a test unless the tests for the plugin itself need to be enhanced to improve coverage, but that's outside of the scope for this change, so I'm going to proceed with it as-is.

Thanks for the review in any case.

PR #1729, opened 2018-08-15
Replace cherrypy.lib.lockfile with zc.lockfile.

jaraco:
I hope this change addresses #1193. Presumably supersedes #1705.


cherrypy/lib/sessions.py
```diff
@@ -564,8 +571,9 @@ def acquire_lock(self, path=None):
 
     def release_lock(self, path=None):
```
webknjaz, line 572: Having a path arg here looks weird to me: it's not used anywhere.

CHANGES.rst
```diff
@@ -1,3 +1,9 @@
+v17.3.0
+-------
+
+* :issue:`1193`: Rely on zc.lockfile for session concurrency
```
webknjaz:
```
via :pr:`1729`
```

it helps a lot when I'm trying to verify things

jaraco: I'll do it this time, but I'm really disinclined to do this extra work in general and here's why: Usually, the user wishes to know the "why" of a change (the issue) and is less interested in the "how", so the issue number is usually more relevant. Moreover, the PR number is often not known at the time the PR is submitted, so requires a two phase process - commit changes, create PR, commit more changes. I want to avoid adding extra steps. As long as the changelog gives a link to a relevant issue (preferably the most relevant issue), it's a small matter from there to see the other activity that's linked, including PRs. I don't want to recommend or ascent to manual processes that replicate information that's readily available through automatic processes. I will sometimes link to a PR when there's no issue present or if that PR is somehow not readily reachable from the issue, but my preference is to link to one issue and let an investigator trace that to the PR with one extra click.

PR #1735, opened 2018-09-01
Drop support for Python 2

jaraco:
I'm ready to drop support for Python 2.

Systems with moderately recent pip versions (9.x+ and maybe earlier) will not see these releases due to the 'python_version' declaration in the metadata.

Older systems still depending on an unpinned cherrypy will simply need to pin to `CherryPy<18`.

I'm mainly creating this PR so I can ensure the Travis tests are passing (they don't seem to run against the branch).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,8 +24,6 @@ params = dict(
         'Framework :: CherryPy',
         'License :: OSI Approved :: BSD License',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
@@ -105,7 +103,7 @@ params = dict(
     setup_requires=[
         'setuptools_scm',
     ],
-    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
+    python_requires='>=3.4',
 )
 
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,16 +15,16 @@ jobs:
         ' >> $BASH_ENV
 
     - run: |-
-        for py_ver in 2.7.14 3.6.4 3.5.4 3.4.7 pypy3.5-5.10.0
+        for py_ver in 3.6.4 3.5.4 3.4.7 pypy3.5-5.10.0
         do
           pyenv install "$py_ver" &
         done
         wait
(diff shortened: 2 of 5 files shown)


jaraco:
Oh, does this project not use Travis-CI anymore? Or did I break the Travis tests with this branch?

webknjaz:
@jaraco Travis CI is used as a major CI, but I've connected a few more trying to compare them and maybe make use of missing features. You might need to scroll the list of checks, successful ones are in the bottom of the list.

Circle may be failing, that's fine

setup.py
```diff
@@ -105,7 +103,7 @@
     setup_requires=[
         'setuptools_scm',
     ],
-    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
+    python_requires='>=3.4',
```
webknjaz, line 106: Maybe start at 3.5?

jaraco: Sure. I'll do that as a separate commit, unrelated to this PR, but in the same release.

webknjaz:
@jaraco you forgot to remove Jython (it's 2.7 compatible, Jython3 never got finished)

jaraco:
Indeed. I think the way it's described is still correct. CherryPy 18 supports Jython and IronPython on Python 3.x (although those don't currently exist) and CherryPy 17 supports Jython and IronPython at their 2.7 versions.

jaraco:
I've released 18.0 with these changes. We can iterate on that.

Also, I've created the [maint/17.x branch](https://github.com/cherrypy/cherrypy/tree/maint/17.x). We should try to target that branch for bugfixes that we wish to release on 17.x.

webknjaz:
@jaraco maybe rename it to just `17`?

webknjaz:
or `python2-17`?

webknjaz:
Some automated tools for backporting would understand it better.

jaraco:
In my other projects, I use `17.x-maintenance`. I'm okay with `17`, although that's potentially confusing because it doesn't start with an alpha character. I don't like `python2-` because that conflates the version with one of the things it supports that the next version doesn't. To be more correct, you'd want `python2-python3.4-17`. Feel free to call it what you like.

webknjaz:
`python2-python3.4-17` sounds good, or maybe `lts-17`?

jaraco:
I like `lts-17` better.

webknjaz:
Great! Let's stick to `lts-17`.

P.S. FWIW I feel like just removing metadata isn't enough for cutting the new release. It should actually break things. Otherwise 18.0.0 vs 18.1.0 would have something actually breaking, not 17 vs 18.

webknjaz:
It's also worth changing the banner in docs.

PR #1736, opened 2018-09-04
Don't convert bytes headers to str

tobiashenkel:
Don't convert bytes header values using str() as str(b'foo') becomes
"b'foo'". Instead just leave bytes as is like it was prior to v18.0.0.

  - [x] bug fix

### DIFF
--- a/cherrypy/lib/httputil.py
+++ b/cherrypy/lib/httputil.py
@@ -513,7 +513,7 @@ class HeaderMap(CaseInsensitiveDict):
         transmitting on the wire for HTTP.
         """
         for k, v in header_items:
-            if not isinstance(v, str):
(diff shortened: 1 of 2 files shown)


webknjaz:
Fair enough. Can we have a test for that, please?

webknjaz:
Also, AppVeyor tests failed.

tobiashenkel:
Sure, I added a test case that fails without the fix.

tobiashenkel:
The AppVeyor test seems to fail due to some unrelated installation problem.

webknjaz (approved):
Thank you!

jaraco:
This fix is going out as 18.0.1.

PR #1743, opened 2018-09-10
Rely on pytest-services 2.0 for Windows compatibility

jaraco:


jaraco:
This is intended as a workaround for https://github.com/pytest-dev/pytest-services/issues/23

jaraco:
Test failures look to be spurious failures.

PR #1755, opened 2018-11-04
Don't convert bytes headers to str

h3llrais3r:
Backport https://github.com/cherrypy/cherrypy/pull/1736
Also fixes https://github.com/Lawouach/WebSocket-for-Python/pull/252

  - [x] bug fix

### DIFF
--- a/cherrypy/lib/httputil.py
+++ b/cherrypy/lib/httputil.py
@@ -518,7 +518,7 @@ class HeaderMap(CaseInsensitiveDict):
         transmitting on the wire for HTTP.
         """
         for k, v in header_items:
-            if not isinstance(v, six.string_types):
+            if not isinstance(v, six.string_types) and not isinstance(v, six.binary_type):
                 v = six.text_type(v)
 
             yield tuple(map(cls.encode_header_item, (k, v)))
--- a/cherrypy/lib/httputil.py
+++ b/cherrypy/lib/httputil.py
@@ -518,7 +518,8 @@ class HeaderMap(CaseInsensitiveDict):
         transmitting on the wire for HTTP.
         """
         for k, v in header_items:
-            if not isinstance(v, six.string_types) and not isinstance(v, six.binary_type):
+            if not isinstance(v, six.string_types) and \
+                    not isinstance(v, six.binary_type):
                 v = six.text_type(v)
 
             yield tuple(map(cls.encode_header_item, (k, v)))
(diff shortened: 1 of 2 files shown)


webknjaz:
Hey @h3llrais3r could you please fix tests?

h3llrais3r:
@webknjaz Can you check again?
I've fixed the travis build (flake8 error) but still errors for pydocstyle (not related to my changes) and tests against latest Python nightly.
As far as I can see, the AppVeyor tests are working fine for Python 2.7, Python 3.5 and Python 3.6. Python 3.5 is not working, but that's not related to my changes...

cherrypy/lib/httputil.py
```diff
@@ -518,7 +518,8 @@ def encode_header_items(cls, header_items):
         transmitting on the wire for HTTP.
         """
         for k, v in header_items:
-            if not isinstance(v, six.string_types):
+            if not isinstance(v, six.string_types) and \
```
webknjaz, line 521: Please use brackets instead. Escaping EOL looks ugly and we don't use it.

webknjaz: And this probably only needs `binary_type` check.

cherrypy/test/test_encoding.py
```diff
@@ -50,6 +50,8 @@ def cookies_and_headers(self):
                 cherrypy.response.cookie['candy']['domain'] = 'cherrypy.org'
                 cherrypy.response.headers[
                     'Some-Header'] = 'My d\xc3\xb6g has fleas'
```
webknjaz, line 52: Maybe add u-prefix here?

cherrypy/lib/httputil.py
h3llrais3r: It is actually used in the code. See https://github.com/cherrypy/cherrypy/blob/maint/17.x/cherrypy/_cpdispatch.py#L345-L347.
I based myself on existing examples, that's why I did it like this.
Where do you want brackets exactly?
I only added the check for `binary_type`, I did not touch the existing check for `string_types`.

cherrypy/test/test_encoding.py
h3llrais3r: No u-prefix is used at other places in this test.
This is also not something I've added. I only added the byte header.
So do you still want me to add it or can we leave it like this?

cherrypy/lib/httputil.py
h3llrais3r: So basically, I applied the same fix (and same test) as in the original fix.

cherrypy/test/test_encoding.py
jaraco: Right - definitely doesn't make sense to be making editorial changes in unaffected code.

Issue #1758, opened 2018-11-21
wspbus.py wait() calls publish during the wait cycle


ian-otto:
**I'm submitting a ...**
- [X] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

**Do you want to request a *feature* or report a *bug*?**

I'm reporting a bug.

**What is the current behavior?**

CherryPy server started through `wspbus` consumes 100% of a core during idle. I resolved this by moving the publish line from within the while loop to after the while loop, in which the server now idles while not serving requests.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**

I noticed it when installing and running [wok](https://github.com/kimchi-project/wok), which uses cherrypy as its backend. Just install this software, start the daemon, then visit the page once and watch the CPU usage for the main process shoot to 100%. I imagine a minimal version of this using a basic "quickstart"-like python script would also exhibit the same behavior.

**What is the expected behavior?**

The server utilizes minimal CPU usage when not serving requests.

**What is the motivation / use case for changing the behavior?**

The service should not consume CPU when it isn't doing anything.

**Please tell us about your environment:**

- Cheroot version: 6.5.2
- CherryPy version: 11.2.0
- Python version: 2.7.15rc1
- OS: Ubuntu 18.04
- Browser: [all]

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)

webknjaz:
Sounds like a bug to me. Feel free to send a PR :)

webknjaz:
Fixed by #1759

PR #1759, opened 2018-11-21
publish should be called when the state is changed

ian-otto:
publish should not be called during every wait cycle, only after the wait has found the correct state

  - [X] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

#1758

#1758

N/A

**Other information**:

**Checklist**:

  - [X] I have [squashed related commits together][2] after the changes have been approved
  - [X] Documentation reflects the changes

### DIFF
--- a/cherrypy/process/wspbus.py
+++ b/cherrypy/process/wspbus.py
@@ -375,7 +375,7 @@ class Bus(object):
 
         while self.state not in states:
             time.sleep(interval)
-            self.publish(channel)
+        self.publish(channel)
 
     def _do_execv(self):
         """Re-execute the current process.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v18.1.0
+-------
+
+* :issue:`1758` via :pr:`1759`: In the bus, when awaiting a
+  state change, only publish after the state has changed.
+
 v18.0.1
 -------
 


webknjaz (commented):
Currently, I'm not sure about the impact of this change. Let's see what @jaraco thinks.

cherrypy/process/wspbus.py
```diff
@@ -375,7 +375,7 @@ def wait(self, state, interval=0.1, channel=None):
 
         while self.state not in states:
             time.sleep(interval)
-            self.publish(channel)
+        self.publish(channel)
```
webknjaz, line 378: Remark: I've checked magicbus and it also has publish inside of the loop.
https://github.com/cherrypy/magicbus/blob/master/magicbus/base.py#L335

cc @aminusfu

ian-otto: Is it possible this was just an error that was propagated to other libraries? Is there a commit somewhere where that was specifically added?

webknjaz: Yea.. It's a huge wholesale decade old commit: https://github.com/cherrypy/cherrypy/commit/f5d026cefc83463a0bebb5459ae81f3cb9b5135c#diff-be33a4f55d59dfc70fc6452482f3a7a4R291. Sure it got into magicbus after cherrypy.

But it's hard to guess what was the reason for putting it like this. That's why I'm tagging guys who might know better.

cc @jaraco

jaraco:
Sorry to let this linger so long. I've reviewed this superficially twice but I continue to feel like there's no clear answer. I agree the code itself seems wrong. What I'm struggling with is determining what this code is intended to do and if it still accomplishes the intention in this new form. I'm trying to devise a test or some other action to improve my confidence in the change.

jaraco:
I removed the `publish` call altogether:

```diff
diff --git a/cherrypy/process/wspbus.py b/cherrypy/process/wspbus.py
index 8f762ef0..4af4c752 100644
--- a/cherrypy/process/wspbus.py
+++ b/cherrypy/process/wspbus.py
@@ -375,7 +375,6 @@ class Bus(object):

         while self.state not in states:
             time.sleep(interval)
-            self.publish(channel)

     def _do_execv(self):
         """Re-execute the current process.
```

Then ran the CherryPy tests, and they all passed. That suggests to me that this code isn't covered by the tests and isn't essential to CherryPy in its typical use.

Given our inability to test this functionality, I'm inclined to rely on the evidence described in the ticket for why this change produces a preferred behavior and accept it.

jaraco (commented):
I reviewed the code in a simple comment. Leaving this comment to close the requested review.

jaraco:
This change is being released as v18.1.0.

webknjaz:
Thanks, I also was confused by all of this and wasn't confident about accepting it myself. The decision to remove it seems fair.

jaraco:
We need to revert this change. As discovered in cherrypy/cheroot#99 and [declared by the docs](http://docs.cherrypy.org/en/latest/extend.html#built-in-channels), the 'main' event is supposed to fire periodically. After this change, the 'main' event no longer fires periodically.

webknjaz:
Right, even integration with other things like http://docs.cherrypy.org/en/latest/deploy.html#tornado requires that.

@jaraco can we just have `if channel == 'main': self.publish()` there?

PR #1769, opened 2019-02-14
Reduced log level for non-error events in win32.py

robinchow:
  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [x] other

N/A

In win32.py's ConsoleCtrlHandler, if start/stop is called multiple times, this is logged at the ERROR (40) level.

If start/stop is called multiple times, this is logged at the INFO (20) level.

**Other information**:
Since the code handles the situation well, and there are no side effects to multiple calls, we shouldn't pollute the logs with error level entries.

Also, because ConsoleCtrlHandler is extremely proactive at removing itself (when it captures Ctrl+C or equivalent), even before it attempts to stop the bus, it sometimes ends up that the handler's stop is called multiple times (directly, then via a 'stop' publish event) during normal operations. There's no inherent badness to this, so while it should be noted/logged, it should be logged at a much lower level.

**Checklist**:

### DIFF
--- a/cherrypy/process/win32.py
+++ b/cherrypy/process/win32.py
@@ -20,7 +20,7 @@ class ConsoleCtrlHandler(plugins.SimplePlugin):
 
     def start(self):
         if self.is_set:
-            self.bus.log('Handler for console events already set.', level=40)
+            self.bus.log('Handler for console events already set.', level=20)
             return
 
         result = win32api.SetConsoleCtrlHandler(self.handle, 1)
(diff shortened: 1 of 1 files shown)


robinchow:
Hi, any update on this? Is there any additional work/discussion required?

webknjaz:
I've restarted tests, will merge after success

robinchow:
I scanned the failed builds, and I don't see anything that links the failed tests to this code change. I could be wrong though, happy to fix up someone points me in the general direction.

webknjaz:
Yeah, TravisCI build got fixed upon restart

webknjaz:
Thanks!

PR #1771, opened 2019-02-20
Hookmap.run does not swallow tracebacks

### LINKED ISSUE #1770: HookMap.run swallows tracebacks
**I'm submitting a ...**
- [x] bug report
- [ ] feature request

Report a bug.

`HookMap.run` does not preserve traceback information when passing an exception.

If the exception is not `HTTPError`, `HTTPRedirect`, `InternalRedirect`, `KeyboardInterrupt` or `SystemExit`, it is logged with traceback through `cherrypy.log`, but then the exception is re-raised without preserving the original traceback.

I could provide all that if that's really needed, but come on...

Re-raise exceptions from hooks in a way that preserves the original traceback.

Making it easier to debug tools from automated tests using `webtest`. In my environment, logging is suppressed to reduce noise, but then tracebacks originating from tools are lost.

- Cheroot version: 6.5.4
- CherryPy version: 17.4.1
- Python version: 2.7.15
- OS: Not relevant
- Browser: Not relevant

**Other information**

The fix is trivial,
https://github.com/cherrypy/cherrypy/blob/master/cherrypy/_cprequest.py#L114
Replace with a bare `raise` statement.

The bare `raise` statement was introduced in Python 1.5.1p1… It provides exactly the semantics wanted here.

https://docs.python.org/release/1.5.1p1/ref/raise.html
https://docs.python.org/release/2.7.15/reference/simple_stmts.html#the-raise-statement
https://docs.python.org/release/3.7.2/reference/simple_stmts.html#the-raise-statement

PR upcoming.

ddaanet:
  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

Fixes #1770

See issue.

See issue.

**Other information**:

Would be nice to have it backported in 17.

**Checklist**:

### DIFF
--- a/cherrypy/_cprequest.py
+++ b/cherrypy/_cprequest.py
@@ -111,7 +111,7 @@ class HookMap(dict):
                     exc = sys.exc_info()[1]
                     cherrypy.log(traceback=True, severity=40)
         if exc:
-            raise exc
+            raise
 
     def __copy__(self):
         newmap = self.__class__()
--- a/cherrypy/_cprequest.py
+++ b/cherrypy/_cprequest.py
@@ -90,7 +90,7 @@ class HookMap(dict):
 
     def run(self, point):
         """Execute all registered Hooks (callbacks) for the given point."""
-        exc = None
+        raising = False
         hooks = self[point]
         hooks.sort()
         for hook in hooks:
@@ -99,18 +99,18 @@ class HookMap(dict):
             # but proceed on to the next hook. The only way
             # to stop all processing from one of these hooks is
             # to raise SystemExit and stop the whole server.
-            if exc is None or hook.failsafe:
+            if not raising or hook.failsafe:
                 try:
                     hook()
                 except (KeyboardInterrupt, SystemExit):
                     raise
                 except (cherrypy.HTTPError, cherrypy.HTTPRedirect,
                         cherrypy.InternalRedirect):
-                    exc = sys.exc_info()[1]
+                    raising = True
                 except Exception:
-                    exc = sys.exc_info()[1]
+                    raising = True
                     cherrypy.log(traceback=True, severity=40)
-        if exc:
+        if raising:
             raise
 
     def __copy__(self):


ddaanet:
Really, this is a lot of hoops jumping for such a simple fix, but this one has been itching me one time too many.

cherrypy/_cprequest.py
```diff
@@ -99,19 +99,19 @@ def run(self, point):
...
-                    exc = sys.exc_info()[1]
+                    raising = True
                 except Exception:
-                    exc = sys.exc_info()[1]
+                    raising = True
                     cherrypy.log(traceback=True, severity=40)
-        if exc:
-            raise exc
+        if raising:
+            raise
```
webknjaz, line 114: I'm pretty sure this will not work because there's nothing to raise in this context (outside of the except block)...

webknjaz:
Is this covered by a known test? Can you provide a minimal reproducer demonstrating what you mean to a stranger?

ddaanet:
Ok. This is getting harder than I expected. Apparrently that relies on a poorly documented behavior of Python. I will work on it.

webknjaz:
What's poorly documented? In Python 3 it should be working just as you want due to automatic exception chaining and storing `__cause__` or/and `__context__` as per https://www.python.org/dev/peps/pep-3134/

jaraco:
It would be particularly nice if this solution could emit _all_ tracebacks encountered during the hook. In other words:

```python
def run(self):
  run_hooks(iter(sorted(self[point])))

def run_hooks(hooks):
  for hook in hooks:
    try:
      hook.run()
    except Exception:
      run_hooks(filter(operator.attrgetter('failsafe'), hooks))
      raise
```

webknjaz:
@jaraco yeah, it'd be nice. I know that some of asyncio devs were thinking about adding another attribute for multiple exception causes to the base exception. While this is useful in the async world in particular, you've correctly pointed out that it's applicable to our architecture as well...

jaraco:
Thanks to webknjaz's reference and looking at this problem while at the CPython sprints, Lisa Roach was able to assist with a couple of references of existing art around handling multiple exceptions in an async model:

- [edb MultiError](https://github.com/edgedb/edgedb/blob/4623b0bc0172999d26357e0a4b9b6bb47a45b47d/edb/common/taskgroup.py#L267-L286)
- [trio MultiError](https://github.com/python-trio/trio/blob/c47c08cbe65790e5f0dfef15571a17b727dbc338/trio/_core/_multierror.py)

Perhaps one of these could be re-used or at least used for inspiration.

Although in our case, we're operating within a single thread synchronously, so maybe the suggestion I've made would be best - rely on the default multiple exception handling behavior in Python to catch the exceptions in the hooks.

Regardless how this is handled, this problem needs a different approach than the one proposed.

PR #1777, opened 2019-05-03
Fix serving files with diacritics in name under their original name properly

### LINKED ISSUE #1776: Cherrypy doesn`t encode UTF-8 filename in HTTP header "Content-Disposition" and returns "download" browser default filename instead.
**I'm submitting a ...**
- [X] bug report
- [ ] feature request

Report a bug.

cherrypy.lib.static.serve_file() method sets wrong "Content-Disposition" and doesn't encode UTF-8 filename properly. It causes error with browser suggested filename - when downloading the file, the value is "download" and it doesn't matter what the original filename is. This could be fixed with RFC6266 (https://tools.ietf.org/html/rfc6266#appendix-D) solution, e.g. filename="my-resume.pdf"; filename*=utf-8''My%20R%C3%A9sum%C3%A9.pdf.

Code:
```
return cherrypy.lib.static.serve_file(
    path='file_path',
    content_type='file_mime_type',
    disposition='attachment',
    name='has_utf-8_character_☃.html',
    debug=True)
```
DEBUG:
TOOLS.STATIC Content-Disposition: 'attachment; filename="has_utf-8_character_☃.html"'

DEBUG:
TOOLS.STATIC Content-Disposition: 'attachment; filename="has_utf-8_character_.html"; filename*=utf-8''has_utf-8_character_%E2%98%83.html'

To be able return file with czech diacritics in name for download under its original name.

- Cheroot version: v6.5.4
- CherryPy version: v18.0.1
- Python version: v3.6.6
- OS: Windows 10
- Browser: [all]

https://stackoverflow.com/questions/1361604/how-to-encode-utf8-filename-for-http-headers-python-django/8996249#8996249

jrxpert:
  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

Fixes #1776

https://github.com/cherrypy/cherrypy/issues/1776

**Other information**:

**Checklist**:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,9 @@ params = dict(
         'Docs: RTD': 'https://docs.cherrypy.org',
         'GitHub: issues': '{}/issues'.format(repo_url),
         'GitHub: repo': repo_url,
+        'Tidelift: funding':
+        'https://tidelift.com/subscription/pkg/pypi-cherrypy'
+        '?utm_source=pypi-cherrypy&utm_medium=referral&utm_campaign=pypi',
     },
     packages=[
         'cherrypy', 'cherrypy.lib',


webknjaz:
Doesn't #1695 address this?

jrxpert:
Hello, thank you for response. Pull request you linked solves diacritics for file upload, this pull request solves diacritics for file download.

cherrypy/lib/static.py
```diff
@@ -28,6 +29,21 @@ def _setup_mimetypes():
...
 
 
+def _utf8_content_disposition(disposition, file_name):
+    """Create HTTP header for downloading a file with UTF-8 filename.
+
+    See this and related answers:
+    https://stackoverflow.com/a/8996249/2173868
+    """
+    ascii_name = unicodedata.normalize(
+        'NFKD', file_name).encode('ascii', 'ignore').decode()
```
webknjaz: I'm not in favor of using `ignore` for error processing. I think it shadows errors. What is the reason behind using it?

cherrypy/lib/static.py
```diff
@@ -28,6 +29,21 @@ def _setup_mimetypes():
...
+    header = '{}; filename="{}"'.format(disposition, ascii_name)
+    if ascii_name != file_name:
```
webknjaz: Should we make this conditional? Adding it unconditionally would simplify the code.

cherrypy/lib/static.py
```diff
@@ -28,6 +29,21 @@ def _setup_mimetypes():
...
 
+def _utf8_content_disposition(disposition, file_name):
+    """Create HTTP header for downloading a file with UTF-8 filename.
+
+    See this and related answers:
+    https://stackoverflow.com/a/8996249/2173868
+    """
+    ascii_name = unicodedata.normalize(
+        'NFKD', file_name).encode('ascii', 'ignore').decode()
+    header = '{}; filename="{}"'.format(disposition, ascii_name)
```
webknjaz: Let's use a Unicode literal here to be explicit:
```suggestion
    header = u'{}; filename="{}"'.format(disposition, ascii_name)
```

cherrypy/lib/static.py
```diff
@@ -28,6 +29,21 @@ def _setup_mimetypes():
 _setup_mimetypes()
 
 
+def _utf8_content_disposition(disposition, file_name):
+    """Create HTTP header for downloading a file with UTF-8 filename.
+
+    See this and related answers:
+    https://stackoverflow.com/a/8996249/2173868
+    """
+    ascii_name = unicodedata.normalize(
```
webknjaz: Could you please change the style here to
```python
var_name = (
    multi().line().
    chain().of().calls()
)
```
?

cherrypy/lib/static.py
```diff
@@ -28,6 +29,21 @@ def _setup_mimetypes():
...
+        'NFKD', file_name).encode('ascii', 'ignore').decode()
+    header = '{}; filename="{}"'.format(disposition, ascii_name)
+    if ascii_name != file_name:
+        quoted_name = urllib.parse.quote(file_name)
+        header += '; filename*=UTF-8\'\'{}'.format(quoted_name)
```
webknjaz:
```suggestion
        header += u'; filename*=UTF-8\'\'{}'.format(quoted_name)
```

webknjaz (approved):
I like this! Please edit it a bit and I'll hit merge :)

webknjaz:
@jrxpert

A few more notes:
1) Commit message `Fix serving files with diacritics in name under their original name properly` doesn't fit 50 chars limit, _could you please change it to_:

    ```
    Fix serving files with diacritics in name under their original name

    This implements adding extra ``filename*`` parameter in Content-disposition
    header when serving content with non-ASCII filenames. As per RFC 6266
    recommendations in the Appendix D.

    Refs:
    * https://stackoverflow.com/a/8996249/2173868
    * https://tools.ietf.org/html/rfc6266#appendix-D

    Fixes #1776
    ```
2) I noted that your commit isn't linked you your GitHub account. That's probably because your local Git configuration contains email which doesn't match the one you set up in GitHub. You may want to fix that :)

webknjaz:
@jrxpert should I expect updates here anytime soon?

cherrypy/lib/static.py
jrxpert: Try use it with default "errors" setting (strict), you will get "UnicodeEncodeError: 'ascii' codec can't encode character '\u2603' in position 20: ordinal not in range(128)" when running tests. Thats protect it against 'ascii' codec unknown UTF8 character error.

cherrypy/lib/static.py
jrxpert: We don`t have to, its ok to simplify it.

cherrypy/lib/static.py
jrxpert: Yes, good idea.

cherrypy/lib/static.py
jrxpert: Yes, no problem.

jrxpert:
I revert previous commit and done it again with few changes related to comments. Because of problems with local `pre-commit` installation, i did this commit without that. Tests are working.

cherrypy/lib/static.py
jrxpert: We don`t have to, its ok to simplify it.

cherrypy/lib/static.py
jrxpert: Yes, good idea.

cherrypy/lib/static.py
jrxpert: Yes, no problem.

cherrypy/lib/static.py
jrxpert: Try use it with default "errors" setting (strict), you will get "UnicodeEncodeError: 'ascii' codec can't encode character '\u2603' in position 20: ordinal not in range(128)" when running tests. Thats protect it against 'ascii' codec unknown UTF8 character error.

Issue #1779, opened 2019-05-13
Potential race conditions in FileSessions  due to deletion of lock files


ralhei:
**I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

**What is the current behavior?**
In the current implementation (using zc.lockfile) and removing the lockfile in lib/sessions.py it is possible that two processes/threads obtain a lock on the identical session lock file. This of course can lead to problems when one thread/process writes into a session file (but hasn't finished) and another thread/process reads incomplete data from it (leading to unpickle errors).

The scenario works because of this:
P1 opens file
P1 locks file
P2 opens file
P1 unlocks file
P1 removes file
P2 locks file (creating a lock on the removed file)
P3 opens file (creates a new lock file)
P3 locks file (now both P2 and P3 hold a lock on the same file)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**

```python
import fcntl, os
flags = fcntl.LOCK_EX | fcntl.LOCK_NB
fp1 = open('x.lock', 'a+')  # first file handle and file descriptor for the lock file
os.remove('x.lock') 
fp2 = open('x.lock', 'a+')  # second file handle and descriptor for the lock file
lock1 = fcntl.flock(fp1.fileno(), flags)   # obtain lock in deleted file
lock2 = fcntl.flock(fp2.fileno(), flags)   # obtain 2nd lock on new file
```

The simple (but maybe not the desired solution) would be to *not* remove the lock file.

**What is the expected behavior?**
The 2nd lock should fail

I went back to a very old Cherrypy Version (3.2.2), there the so called 'naive' lock file approach 
via `fd = os.open(path, os.O_CREAT | os.O_WRONLY | os.O_EXCL)` was used, this indeed is very reliable, I could not reproduce my session lock errors with that ;-)

**Please tell us about your environment:**

- Cheroot version: 6.5.5
- CherryPy version: 18.1.1
- Python version: 3.6.3
- OS: Centos7
- Browser: all

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)

webknjaz:
@ralhei hi, do you have any ideas on how to improve this? We've seen some issues with this but haven't figured it out yet. So any help is welcome :)

webknjaz:
[...]

Issue #1781, opened 2019-05-31
After the python3 fix for CVE-2019-9740 test_null_bytes fails


mcepl:
**I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When running the testsuite StaticTest.test_null_bytes fails with:
```
[  107s] __________________________ StaticTest.test_null_bytes __________________________
[  107s] 
[  107s] self = <cherrypy.test.test_static.StaticTest testMethod=test_null_bytes>
[  107s] 
[  107s]     def test_null_bytes(self):
[  107s] >       self.getPage('/static/\x00')
[  107s] 
[  107s] cherrypy/test/test_static.py:401: 
[  107s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[  107s] cherrypy/test/helper.py:326: in getPage
[  107s]     protocol, raise_subcls)
[  107s] /usr/lib/python3.7/site-packages/cheroot/test/webtest.py:212: in getPage
[  107s]     raise_subcls, ssl_context=self.ssl_context,
[  107s] /usr/lib/python3.7/site-packages/cheroot/test/webtest.py:514: in openURL
[  107s]     skip_accept_encoding=True,
[  107s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[  107s] 
[  107s] self = <http.client.HTTPConnection object at 0x7f1697a032b0>, method = 'GET'
[  107s] url = '/static/\x00', skip_host = True, skip_accept_encoding = True
[  107s] 
[  107s]     def putrequest(self, method, url, skip_host=False,
[  107s]                    skip_accept_encoding=False):
[  107s]         """Send a request to the server.
[  107s]     
[  107s]         `method' specifies an HTTP request method, e.g. 'GET'.
[  107s]         `url' specifies the object being requested, e.g. '/index.html'.
[  107s]         `skip_host' if True does not add automatically a 'Host:' header
[  107s]         `skip_accept_encoding' if True does not add automatically an
[  107s]            'Accept-Encoding:' header
[  107s]         """
[  107s]     
[...]

Issue #1540, opened 2016-12-27
test_4_File_deletion fails intermittently


jaraco:
As observed in [this job](https://travis-ci.org/cherrypy/cherrypy/jobs/187052318), the sessions test_4_File_deletion fails intermittently with this error:

```
_______________________ SessionTest.test_4_File_deletion _______________________

self = <cherrypy.test.test_session.SessionTest testMethod=test_4_File_deletion>
    def tearDown(self):
        # Clean up sessions.
        for fname in os.listdir(localDir):
            if fname.startswith(sessions.FileSession.SESSION_PREFIX):
>               os.unlink(os.path.join(localDir, fname))
E               OSError: [Errno 2] No such file or directory: '/home/travis/build/cherrypy/cherrypy/cherrypy/test/session-7029320719c2c8a76f8eb4ce0cea3d0a29b36fec.lock'

cherrypy/test/test_session.py:144: OSError
```

Issue #1306, opened 2014-03-21
test_2_File_Concurrency fails intermittently


bb-migration:
Originally reported by: **Nathan Kinder (Bitbucket: [nkinder](http://bitbucket.org/nkinder), GitHub: [nkinder](http://github.com/nkinder))**

---

I've noticed intermittent failures of the test_2_File_Concurrency test in test_session.py.  I'm using a recent checkout from the default branch (CherryPy 3.2.5 plus whatever has been committed since) with Python 2.7.5.

The failures can be seen by simply running the test_session.py tests in a loop as follows:

```
  for i in `seq 1 100`; do nosetests -s ./test_session.py; done
```

I'm seeing failures about 5% of the time.  Here is an example of the failure output:

```
S..........
----------------------------------------------------------------------
Ran 11 tests in 10.387s

OK (SKIP=1)
S...F......
======================================================================
FAIL: test_2_File_Concurrency (cherrypy.test.test_session.SessionTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/nathank/Source/cherrypy/cherrypy/cherrypy/test/test_session.py", line 218, in test_2_File_Concurrency
    self._test_Concurrency()
  File "/home/nathank/Source/cherrypy/cherrypy/cherrypy/test/test_session.py", line 268, in _test_Concurrency
    self.assertEqual(hitcount, expected)
AssertionError: 150 != 151
-------------------- >> begin captured logging << --------------------
cherrypy.error: INFO: [19/Mar/2014:14:41:18] ENGINE Stopped thread 'Session cleanup'.
cherrypy.access.49232208: INFO: 127.0.0.1 - - [19/Mar/2014:14:41:18] "GET /setsessiontype/file HTTP/1.1" 200 - "" ""
cherrypy.error: INFO: [19/Mar/2014:14:41:18] ENGINE Started monitor thread 'Session cleanup'.
cherrypy.access.49232208: INFO: 127.0.0.1 - - [19/Mar/2014:14:41:18] "GET / HTTP/1.1" 200 1 "" ""
cherrypy.access.49232208: INFO: 127.0.0.1 - - [19/Mar/2014:14:41:18] "GET / HTTP/1.1" 200 1 "" ""
...
cherrypy.access.49232208: INFO: 127.0.0.1 - - [19/Mar/2014:14:41:20] "GET / HTTP/1.1" 200 3 "" ""
[...]

PR #1787, opened 2019-06-23
Do not remove session lock files

suheb:
Signed-off-by: Suhaib Khan <suheb.work@gmail.com>

  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

#1391 (#1779)

We remove the lock file when clearing sessions.

We do not remove the lock files when clearing sessions

**Other information**:
The issue author #1779 provides two possible solutions. This PR uses the first one to fix the issue.
I have been not able to reproduce the issue so did not write tests to verify if this actually fixes the issue. Would really like some help on how to reproduce this so that I can include tests to verify this as well.      

**Checklist**:


cherrypy/lib/sessions.py
```diff
@@ -566,8 +565,6 @@ def acquire_lock(self, path=None):
     def release_lock(self, path=None):
         """Release the lock on the currently-loaded session data."""
         self.lock.close()
-        with contextlib.suppress(FileNotFoundError):
-            os.remove(self.lock._path)
```
webknjaz, line 570: I think we need something smarter than not removing files.

jaraco: My instinct is that less code is better (especially since the file lock should ideally be managed by zc.lockfile). It looks like this change was proposed way back in #1391 also. I don't like the idea of leaving artifacts around after exiting, but since Unix allows two manifestations of the same file to exist at the same time, I don't think there's any option but to always leave the file.

webknjaz: I'm curious whether we could somehow introduce better locking here.

jaraco:
If making this change removes the intermittent failures in the failing session tests (#1540/#1534/#1306 and maybe others), I'd say that's sufficient to qualify as the reproducer and regression test.

jaraco (changes requested):
Please explore if removing the xfail on the related intermittently-failing tests demonstrates the superiority of this change.

cherrypy/lib/sessions.py
jaraco: In my experience, locking is hard to get right, which is why I've been seeking to rely on libraries that are battle tested and relied upon by other frameworks. I'd be open to another implementation, but would prefer if CherryPy didn't have to host it. I don't see anything that CherryPy is doing special that warrants a custom lock behavior.

webknjaz: I absolutely agree that it should be fully implemented in a separate dist.

suheb:
@jaraco I removed xfail on the `File_Concurrency`(#1306) test. The test passes locally and in travis but fails in the appveyor build.

webknjaz:
Right, AFAIR that test was failing specifically under Windows because the fs/win doesn't allow opening files when there's another file descriptor open on those files.

suheb:
@webknjaz Ah!  Good to know.
@jaraco The File_Concurrency(#1306) test *now* passes on Unix systems but still fails on Windows. Does that qualify as the reproducer and regression test?

jaraco:
@suheb - yes, quite possibly, especially if this patch also corrects that failure. I'm confident that these failures were intermittent and more sensitive on some platforms than others. I'll try running the tests in this PR on my macOS workstation and see how it compares to master.

jaraco:
I ran `tox -- -k File_Concurrency` many (10+) times on this branch and was unable to get the test to fail, whereas on master, the test xfails frequently. So yes, I believe that test is sufficient for a reproducer. Now to figure out why tests are failing on Windows.

It also looks like tests are failing for `nightly`, which is preventing some of the other tests from running. That issue (reported in #1781) is independent of this one, but needs to be corrected so the remaining tests can run.

webknjaz:
@jaraco #1781 is not just a problem with a test, it actually reveals a need to fix Cheroot...

suheb:
If I remember correctly, last week the Travis build was passing and the Appveyor build was failing. Now, the Appveyor build seems to be passing but the Travis build is failing due to #1781 (as correctly pointed by you @jaraco). Strange!

webknjaz:
Yes, which probably means that it's still unstable.

webknjaz:
@suheb you need to rebase your changes on top of master.

suheb:
> Yes, which probably means that it's still unstable.

@webknjaz But the test failing on Travis was due to #178, which is completely unrelated to this change.

suheb:
@webknjaz @jaraco Are there any other changes that we need to do in this PR?

webknjaz:
Honestly, I don't feel confident applying this change so I'll defer it to @jaraco to merge if he thinks it's a good idea. But yes, no other changes seem to be necessary.

jaraco (approved):

PR #1793, opened 2019-07-25
HTTP redirect 308

ingoogni:
  - [x] bug fix
  - [x] feature

Cherrypy currently lacks a HTTP redirect 308

Adds a redirect 308 option to _cperror.py

### DIFF
--- a/cherrypy/_cperror.py
+++ b/cherrypy/_cperror.py
@@ -34,6 +34,7 @@ user:
                                               POST should not raise this error
 305      Use Proxy                            Confirm with the user
 307      Temporary Redirect                   Confirm with the user
+308      Moved Permanently                    Confirm with the user
 =====    =================================    ===========
 
 However, browsers have historically implemented these restrictions poorly;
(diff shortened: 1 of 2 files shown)


ingoogni:
Meh, something goes wrong and cant find to undo it. The PATCH for cache (05ff41a) should not be there

ingoogni:
Try again

webknjaz:
Hey! If you're having troubles with Git, here's a useful book you could read — https://git-scm.com/book/en/v2/ :)

cherrypy/_cperror.py
```diff
@@ -34,6 +34,7 @@
                                               POST should not raise this error
 305      Use Proxy                            Confirm with the user
 307      Temporary Redirect                   Confirm with the user
+308      Moved Permanently                    Confirm with the user
```
webknjaz, line 37: There's line 42 starting with `or 307 when redirecting POST`, it should probably also mention 308.

PR #1804, opened 2019-08-15
fix RamSession Cleanup Thread Death #1803

### LINKED ISSUE #1803: RamSession clean_up() occasionally causes cleanup thread death
**I'm submitting a ...**
- [x] bug report
- [ ] feature request

bug

Under heavy load (25K req/hr) and a large session cache (10K sessions), the `RamSession.clean_up()` method will occasionally (~1/week) raise a `RuntimeError` with the message "dictionary changed size during iteration". This causes the cleanup thread to die, and the cache then grows to use all available memory.

```python-traceback
[15/Aug/2019:02:06:11] ENGINE Error in background task thread function <bound method 
RamSession.clean_up of <cherrypy.lib.sessions.RamSession object at 0x7f138d920438>>.
  File "/var/lib/autocat/.local/share/virtualenvs/autocat3-mGIyuTut/lib/python3.6/site-packages/cherrypy/process/plugins.py", line 517, in run
    self.function(*self.args, **self.kwargs)
  File "/var/lib/autocat/.local/share/virtualenvs/autocat3-mGIyuTut/lib/python3.6/site-packages/cherrypy/lib/sessions.py", line 407, in clean_up
    for _id, (data, expiration_time) in list(self.cache.items()):
RuntimeError: dictionary changed size during iteration

```
The autocat3 service serves the dynamic content on the Project Gutenberg website.

the cleanup thread should not die.

not having the server die in the night every week or so.

- Cheroot version: 6.5.5
- CherryPy version: 18.1.1
- Python version: 3.6.8
- OS: CentOS Linux release 7.6.1810 (Core)
- Browser: all
- Number of server threads: 10

I am preparing a PR which catches `RuntimeError`s from `list(self.cache.items())` in `RamSession`; a patched class is now in production. and I'd like to see it run for a week.

The other, more general way to fix this would be to change line 301 in `lib.sessions` from ```if self.clean_freq and not cls.clean_thread``` to ```if self.clean_freq and (not cls.clean_thread or not cls.clean_thread.is_alive())``` but I do not know the code well enough to judge whether this is overkill.

eshellman:
So the cleanup thread doesn't die

  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

Fixes #1803

#1803

cleanup thread survives `RuntimeError`

**Other information**:

`test_session.test_8_Ram_Cleanup` covers the function but I'm not sure how you could generate the exception in a test environment.

The other, more general way to fix this would be to change line 301 in `lib.sessions` from `if self.clean_freq and not cls.clean_thread` to `if self.clean_freq and (not cls.clean_thread or not cls.clean_thread.is_alive())` but I do not know the code well enough to judge whether this is overkill.

**Checklist**:

  - [?] Unit tests for the changes exist
  - [?] Integration tests for the changes exist (if applicable)
  - [x] Documentation reflects the changes

### DIFF
--- a/cherrypy/lib/sessions.py
+++ b/cherrypy/lib/sessions.py
@@ -404,18 +404,28 @@ class RamSession(Session):
         """Clean up expired sessions."""
 
         now = self.now()
-        for _id, (data, expiration_time) in list(self.cache.items()):
-            if expiration_time <= now:
-                try:
-                    del self.cache[_id]
-                except KeyError:
-                    pass
-                try:
-                    if self.locks[_id].acquire(blocking=False):
-                        lock = self.locks.pop(_id)
-                        lock.release()
-                except KeyError:
-                    pass
+        try:
+            for _id, (data, expiration_time) in list(self.cache.items()):
+                if expiration_time <= now:
+                    try:
+                        del self.cache[_id]
+                    except KeyError:
+                        pass
+                    try:
+                        if self.locks[_id].acquire(blocking=False):
+                            lock = self.locks.pop(_id)
+                            lock.release()
+                    except KeyError:
+                        pass
+
+        except RuntimeError:
+            """Under heavy load, list(self.cache.items()) will occasionally raise this
+            error for large session caches with message
+            "dictionary changed size during iteration"
+            Better to pause the cleanup than to let the cleanup thread die.
+            """
+
+            return
 
         # added to remove obsolete lock objects
         for _id in list(self.locks):


cherrypy/lib/sessions.py
```diff
@@ -404,18 +404,28 @@ def clean_up(self):
...
-                except KeyError:
-                    pass
-                try:
-                    if self.locks[_id].acquire(blocking=False):
-                        lock = self.locks.pop(_id)
-                        lock.release()
-                except KeyError:
-                    pass
+        try:
+            for _id, (data, expiration_time) in list(self.cache.items()):
```
webknjaz, line 408: Honestly, I'm not sure about the try/except. It just hides the problem.

I think that the initial purpose of `list()` was to copy the contents so that the for-loop would iterate over an immutable iterable. But it seems like in the case you describe the copying process (new list creation) *lazily* iterates over the dict entries (`.items()` returns an iterator). And that's where it fails, not in the for-loop as you might've thought.

That said, the proper patch following your way would be to wrap just `list(self.cache.items())` with try/except. Yet, swallowing a RuntimeError like that doesn't feel right.

I think that the ultimate solution would be to actually copy the dict which, I think, is an atomic operation. So I'd apply this instead of wrapping everything with try/except:
```suggestion
            for _id, (data, expiration_time) in self.cache.copy().items():
```

webknjaz (changes requested):
Thanks for the PR, I think I've found a better way. Would you mind adjusting the patch accordingly?

cherrypy/lib/sessions.py
eshellman: any suggestions on how to test this alternative?

webknjaz: I've done some research and I'm pretty sure we can rely on this:
```python
In [1]: d1 = {1: 2}                                                                                                                     

In [2]: import dis                                                                                                                      

In [3]: dis.dis('d2 = d1.copy()')                                                                                                       
  1           0 LOAD_NAME                0 (d1)
              2 LOAD_METHOD              1 (copy)
              4 CALL_METHOD              0
              6 STORE_NAME               2 (d2)
              8 LOAD_CONST               0 (None)
             10 RETURN_VALUE

In [4]: dis.dis('d3 = dict(d1)')                                                                                                        
  1           0 LOAD_NAME                0 (dict)
              2 LOAD_NAME                1 (d1)
              4 CALL_FUNCTION            1
              6 STORE_NAME               2 (d3)
              8 LOAD_CONST               0 (None)
             10 RETURN_VALUE

In [5]: dis.dis('list(d1.items())')                                                                                                     
  1           0 LOAD_NAME                0 (list)
              2 LOAD_NAME                1 (d1)
              4 LOAD_METHOD              2 (items)
              6 CALL_METHOD              0
              8 CALL_FUNCTION            1
             10 RETURN_VALUE

In [6]: dis.dis('list(d1.copy().items())')                                                                                              
  1           0 LOAD_NAME                0 (list)
              2 LOAD_NAME                1 (d1)
              4 LOAD_METHOD              2 (copy)
              6 CALL_METHOD              0
              8 LOAD_METHOD              3 (items)
             10 CALL_METHOD              0
             12 CALL_FUNCTION            1
             14 RETURN_VALUE
```

OTOH if you want to test you could deploy the following in production and see if the error is still happening:
```suggestion
            try:
                cache_items_copy = self.cache.copy().items()
            except RuntimeError as re:
                cherrypy.log(f'Runtime Error happened while copying the cache entries: {re}')
                cherrypy.log('Ref: https://github.com/cherrypy/cherrypy/pull/1804')
                return
            for _id, (data, expiration_time) in cache_items_copy:
```
without that outer try/except.

Alternatively, you could try reproducing this using `ab`/`Locust` or similar.

eshellman: I tried to reproduce the error in some test code without success. I will implement your suggestion in production and see if the runtimeerror occurs with cache.copy().items()

PR #1813, opened 2019-10-02
Refer enterprise customers to Tidelift for commercial support

jaraco:
Alongside cherrypy/cherrypy.github.io#8, this PR updates the documentation and README to refer potential commercial clients to commercial support offerings through Tidelift.

### DIFF
--- a/docs/enterprise.rst
+++ b/docs/enterprise.rst
@@ -1,3 +1,6 @@
+.. todo: consider linking to the same text in README.rst.
+.. https://github.com/cherrypy/cherrypy/pull/1813#discussion_r330805322
+
 For Enterprise
 ==============
 
--- a/docs/enterprise.rst
+++ b/docs/enterprise.rst
@@ -1,0 +1,10 @@
+For Enterprise
+==============
+
+CherryPy is available as part of the Tidelift Subscription.
+
+The CherryPy maintainers and the maintainers of thousands of other packages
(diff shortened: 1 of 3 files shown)


docs/enterprise.rst
```diff
@@ -0,0 +1,10 @@
+For Enterprise
```
webknjaz, line 4: FYI you can include that chunk from README using sphinx.

jaraco: Can I? I'm not aware of how to include a section of another file in a file. If I could, that would be great. 

_does the minimal amount of research_

I do see [this example](https://muffinresearch.co.uk/selectively-including-parts-readme-rst-in-your-docs/) which leads me to the `start-after` and `end-before` directives in [the docs](http://docutils.sourceforge.net/docs/ref/rst/directives.html#including-an-external-document-fragment).

If I could include a section by name, I think I'd implement that, but since I can't and I must bound the included content with start/stop sentinels, I'm more inclined to just live with the duplication for now.

webknjaz: Yes, that's the one. Use `include` or `literalinclude`. The trick is to use comment lines which won't be rendered anyway. You could also add some `-do-not-remove-` thing into those comments.

jaraco: In dca4a517, I added a "todo" so as not to forget.

Issue #1817, opened 2019-10-16
PipelineTests.test_HTTP11_Timeout_after_request failing


jaraco:
The `PipelineTests.test_HTTP11_Timeout_after_request` test is failing consistently [in Travis-CI](https://travis-ci.org/cherrypy/cherrypy/jobs/598346436#L546-L638) and on my local workstation.

jaraco:
Pinning to `cheroot<8.1` bypasses the failure, so it seems the [changes in 8.1](https://docs.cherrypy.org/projects/cheroot/en/latest/history.html#v8-1-0) (request-based workers) are implicated.

/cc @the-allanc

jaraco:
For now, I've disabled the tests to suppress the failures. To reproduce the failure, run `tox -- -k Timeout_after_request --runxfail`.

jaraco:
I've confirmed that [this line](https://github.com/cherrypy/cheroot/blob/02f7a17cbc98398def138c6ce42d8cd6487955dd/cheroot/connections.py#L192) is being reached and the value is `1`, so I would expect the socket objects to have 1-second timeouts. So why is the timeout not being hit with the test?

webknjaz:
I bet that timeouts happen on read/write (`socket.recv()`/`socket.write()`) which probably means that this may be happening in a place different than you expect.

PR #1820, opened 2019-10-17
Fix test_HTTP11_Timeout_after_request to expect 408 responses

the-allanc:
  - [X] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

#1817 

Test fails.

Test passes!

**Other information**:

The changes in cheroot now mean that we're better at writing a 408 response to indicate that the time waiting for a request has timed out - it's just that the underlying test in CherryPy wasn't expecting this consistent behaviour, and so expected a 408 response in some circumstances and a socket error in others.

### DIFF
--- a/cherrypy/test/test_conn.py
+++ b/cherrypy/test/test_conn.py
@@ -390,12 +390,10 @@ class PipelineTests(helper.CPWebCase):
         except Exception:
             if not isinstance(sys.exc_info()[1],
                               (socket.error, BadStatusLine)):
-                self.fail("Writing to timed out socket didn't fail"
-                          ' as it should have: %s' % sys.exc_info()[1])
+                self.fail(msg % sys.exc_info()[1])
         else:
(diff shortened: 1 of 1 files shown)


webknjaz:
I don't understand this. You only changed the error message, right?

webknjaz:
Plz remove `@pytest.mark.xfail` from the test.

the-allanc:
> I don't understand this. You only changed the error message, right?

The test itself tries in two different places to trigger a timeout. The first expects the timeout to manifest itself as either a socket error, a BadStatusLine error, or a 408 response. The second didn't accept a 408 response, but it should have done - and the changes in Cheroot now make that more likely.

I've just made the second check consistent with the first check (and reused the already defined error message).

webknjaz (approved):

webknjaz:
Thanks, Allan!

Issue #1818, opened 2019-10-17
Test failures in test_wsgi_unix_socket


jaraco:
When running tests on my macOS workstation, I see frequent test failures in test_wsgi_unix_socket like this:

```
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― WSGI_UnixSocket_Test.test_simple_request ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

func = functools.partial(<function _open_url_once at 0x10d18fdc0>, b'/', None, 'GET', None, '127.0.0.1', 54583, <cherrypy.test.test_wsgi_unix_socket.USocketHTTPConnection object at 0x10d83ca90>, 'HTTP/1.1', ssl_context=None)
cleanup = <function openURL.<locals>.on_exception at 0x10e195700>, retries = 9, trap = <class 'OSError'>

    def retry_call(func, cleanup=lambda: None, retries=0, trap=()):
        """
        Given a callable func, trap the indicated exceptions
        for up to 'retries' times, invoking cleanup on the
        exception. On the final attempt, allow any exceptions
        to propagate.
        """
        attempts = count() if retries == float('inf') else range(retries)
        for attempt in attempts:
                try:
>                       return func()

attempt    = 0
attempts   = range(0, 9)
cleanup    = <function openURL.<locals>.on_exception at 0x10e195700>
func       = functools.partial(<function _open_url_once at 0x10d18fdc0>, b'/', None, 'GET', None, '127.0.0.1', 54583, <cherrypy.test.test_wsgi_unix_socket.USocketHTTPConnection object at 0x10d83ca90>, 'HTTP/1.1', ssl_context=None)
retries    = 9
trap       = <class 'OSError'>

.tox/python/lib/python3.8/site-packages/jaraco/functools.py:323: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

url = '/', headers = [('Host', '127.0.0.1:54583')], method = 'GET', body = None, host = '127.0.0.1', port = 54583
[...]

PR #1821, opened 2019-10-18
Pass arguments from CPWebCase.getPage to WebCase.getPage without replicating signature. 

jaraco:
Ref #1818.

This change simplifies the code and avoids coupling the two function call signatures.

Because it bumps the cheroot dependency to 8.2.1, I want to wait some time (2 weeks) before requiring that dependency.


webknjaz:
@jaraco more than two weeks passed now

PR #1825, opened 2019-11-20
use pytest 3.4 compatible fixtures for tmp paths

### LINKED ISSUE #1824: `tmp_path_factory` fixture is incompatible with old pytest
**I'm submitting a ...**
- [x] bug report
- [ ] feature request

This could be considered a bug, it is really a minor issue for packaging CherryPy on CentOS8 where pytest 3.4.2 is available.

The `tmp_path_factory` fixture is unavailable in pytest 3.4.2 which is the latest version available in EPEL8, and it breaks the `check` step in the spec file that runs pytest against CherryPy

```python-traceback
==================================== ERRORS ====================================
_____________________ ERROR at setup of test_normal_return _____________________
file /builddir/build/BUILD/CherryPy-18.4.0/cherrypy/test/test_logging.py, line 109
  def test_normal_return(log_tracker, server):
file /builddir/build/BUILD/CherryPy-18.4.0/cherrypy/test/test_logging.py, line 102
  @pytest.fixture
  def log_tracker(access_log_file):
file /builddir/build/BUILD/CherryPy-18.4.0/cherrypy/test/test_logging.py, line 18
  @pytest.fixture
  def access_log_file(tmp_path_factory):
E       fixture 'tmp_path_factory' not found
>       available fixtures: access_log_file, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, configure_server, cov, doctest_namespace, error_log_file, log_tracker, monkeypatch, no_cover, pytestconfig, record_xml_attribute, record_xml_property, recwarn, server, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.

/builddir/build/BUILD/CherryPy-18.4.0/cherrypy/test/test_logging.py:18
```

Tests can run with pytest 3.4.2 with a backward compatible fixture

I'm packaging CherryPy for EPEL8 and this is preventing doing so with the constraint of an older pytest version

- Cheroot version: X.X.X
- CherryPy version: tip of master
- Python version: 3.6.X
- OS: CentOS8

alfredodeza:
  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

**What is the related issue number (starting with `#`)** Resolves #1824 

**What is the current behavior?**  See #1824 

**What is the new behavior (if this is a feature change)?** No behavior is changed.

**Other information**:

**Checklist**:

### DIFF
--- a/cherrypy/test/test_logging.py
+++ b/cherrypy/test/test_logging.py
@@ -16,13 +16,17 @@ erebos = u'\u0388\u03c1\u03b5\u03b2\u03bf\u03c2.com'
 
 
 @pytest.fixture
-def access_log_file(tmp_path_factory):
-    return tmp_path_factory.mktemp('logs') / 'access.log'
+def access_log_file(tmpdir):
+    p = tmpdir.mkdir("access_logs").join("access.log")
+    p.write("")
+    return p
 
 
 @pytest.fixture
-def error_log_file(tmp_path_factory):
-    return tmp_path_factory.mktemp('logs') / 'access.log'
+def error_log_file(tmpdir):
+    p = tmpdir.mkdir("error_logs").join("error.log")
+    p.write("")
+    return p
 
 
 @pytest.fixture


webknjaz:
Hi,

I don't think we ever intended to support that old version of pytest. The supported way testing is using `tox` which is configured properly.

I'd rather raise the required version in the dependencies to 5.0+.

I don't know why you'd want to rely on system-installed versions of libs rather than use virtualenvs per project but you could patch it on the RPM level or build an RPM with newer pytest.

alfredodeza:
I understand the flexibility that virtualenv and tox provides and that one can enable testing with those tools. 

Distributions and their packages are regulated under strict constraints that package maintainers need to work with. As a Python developer I do prefer to use tox, virtualenv, and non-system packages, but this is not the case.

This PR makes it easier to package cherrypy for EPEL8 so that other system libraries can use it and require it.

In this particular case, the Ceph project is currently blocked on not being able to work with Python 3 and CentOS8 because there is no CherryPy package https://github.com/ceph/ceph/pull/25969

My hope is that the CherryPy maintainers can be lenient enough to help get us past this issue to enable other projects in distributions like CentOS8.

As a side note, we can't _build an RPM with newer pytest_, although there is a plan to introduce newer versions in the future.

webknjaz:
The problem here is introducing more maintenance burden for the CherryPy team
which is outnumbered.
That's why we decided, a long time ago, that if somebody wants to package it
using other package managers, it's their responsibility to apply patches that
are unique to their problems, not ours.
Otherwise, every each esoteric package manager would come and demand that we
support their problematic cases on our side which would pull us back into the Stone Age,
not allowing us to use modern tech at all.

You're lucky that you hit just one place that depends on modern `pytest` features. That is
because most of the tests still rely on `unittest` but my plan is to get rid of it and fully rely
on the newest `pytest` at all times because it's not a runtime dependency and shouldn't
prevent us from simplifying the maintenance.
So even if I accept this, soon you'll hit similar problems again. It's not a sustainable
solution for either of us.

Also, we only support what we run in the CI. If we were to support things like CentOS,
this would require someone to implement testing against it in the CI.
And then keep it healthy, fix related bugs and so on. While this is possible with GitHub Actions,
somebody has to really commit to doing this. None of the current maintainers are sponsored to
work on CherryPy full-time by their employers, not even part-time. So we just don't have time
to do this ourselves and are not interested in making the maintenance move complicated than it is.

alfredodeza:
> The problem here is introducing more maintenance burden for the CherryPy team
> which is outnumbered.

The proposed changes can't possibly be labeled a _maintenance burden_. It was a minor change to use a supported pytest fixture.

> That's why we decided, a long time ago, that if somebody wants to package it
> using other package managers, it's their responsibility to apply patches that
> are unique to their problems, not ours.

This is reasonable and I agree here, but nothing should prevent small changes to make that work easier on the rest.

> Otherwise, every each esoteric package manager would come and demand that we
> support their problematic cases on our side which would pull us back into the Stone Age,
> not allowing us to use modern tech at all.
> 

Again, I am not demanding anything here, I identified this would be a very small patch which uses a *currently supported* fixture from pytest, instead of using one that is so new that it isn't available everywhere. This patch is not pulling CherryPy back to the Stone Age.

> You're lucky that you hit just one place that depends on modern `pytest` features. That is
> because most of the tests still rely on `unittest` but my plan is to get rid of it and fully rely
> on the newest `pytest` at all times because it's not a runtime dependency and shouldn't
> prevent us from simplifying the maintenance.
> So even if I accept this, soon you'll hit similar problems again. It's not a sustainable
> solution for either of us.

Even though you are raising hypothetical reasons here, all I am asking is to accept this minor patch so that it buys us time to get pytest to a newer version. I am not expecting the CherryPy project to keep fixing things to support an older pytest version. We are working to move forward, and this PR would help (and a subsequent release).

> 
> Also, we only support what we run in the CI. If we were to support things like CentOS,
> this would require someone to implement testing against it in the CI.

I am not sure where this is coming from. There is no need for you or the CherryPy project to "support CentOS" or "test it in the CI". This is on us.

> And then keep it healthy, fix related bugs and so on. While this is possible with GitHub Actions,
> somebody has to really commit to doing this. None of the current maintainers are sponsored to
> work on CherryPy full-time by their employers, not even part-time. So we just don't have time
> to do this ourselves and are not interested in making the maintenance move complicated than it is.

Again, I am not sure where you are coming with all these requirements, I am not suggesting any kind of extra responsibilities, or testing.

My hope is that this PR can be merged and a minor (point) release can be made so that CentOS8 can get the latest version of CherryPy with as little effort as possible from both sides.  Doing this will buy us time to get newer versions of Pytest available.

webknjaz:
Okay, I don't agree with all of your points but I'm willing to accept this change and release a patch
or even a smaller post-release just this time (because it doesn't fix anything in runtime,
I believe that a tests-only change doesn't affect anything semver-wise).

But I have to re-emphasize that because nothing in the CI is testing this change, it will
get lost eventually during refactoring.

Still, this PR isn't fully ready: https://travis-ci.org/cherrypy/cherrypy/jobs/614785272#L532.
To fix the linter, you can run `tox -e pre-commit` locally.

webknjaz:
Please also raise the minimum version to 3.4 here: https://github.com/cherrypy/cherrypy/blob/8265854/setup.py#L84.

jaraco:
Alfredo, hey! Hope you're well.

The main problem I have with this fix is that it can be easily optimized away. That is, another PR could come in tomorrow and switch `tmpdir` to the newer, preferred `tmp_path_factory`, and there's nothing that would prevent its acceptance. It's a footgun waiting to happen.

Moreover, this compatibility change is implying a requirement (to support older Pytests) without any timeline. When should this support expire?

I'd be happy to accept this change with a few modifications:

- [ ] Implement a compatibility shim, something that wraps `tmpdir` with something that implements the interfaces that CherryPy is using.
- [ ] In that shim, document why this is there instead of `tmp_path_factory`, probably linking to the ticket.
- [ ] Document in that ticket or in this shim when that shim can be removed. What are the conditions under which this requirement would go away.

It's possible CherryPy will at some point need to make a tradeoff between keeping this compatibility shim and dropping support for CENTOS 8, but if it has these changes I've outlined, that would make it straightforward to communicate that loss of compatibility (with a backward-incompatible release) and giving CentOS options to support (pin to older version).

jaraco (changes requested):
Alfredo, hey! Hope you're well.

The main problem I have with this fix is that it can be easily optimized away. That is, another PR could come in tomorrow and switch `tmpdir` to the newer, preferred `tmp_path_factory`, and there's nothing that would prevent its acceptance. It's a footgun waiting to happen.

Moreover, this compatibility change is implying a requirement (to support older Pytests) without any timeline. When should this support expire?

I'd be happy to accept this change with a few modifications:

- [ ] Implement a compatibility shim, something that wraps `tmpdir` with something that implements the interfaces that CherryPy is using.
- [ ] In that shim, document why this is there instead of `tmp_path_factory`, probably linking to the ticket.
- [ ] Document in that ticket or in this shim when that shim can be removed. What are the conditions under which this requirement would go away.

It's possible CherryPy will at some point need to make a tradeoff between keeping this compatibility shim and dropping support for CENTOS 8, but if it has these changes I've outlined, that would make it straightforward to communicate that loss of compatibility (with a backward-incompatible release) and giving CentOS options to support (pin to older version).

webknjaz:
@jaraco I think we agreed earlier that we could release just one version of CherryPy with this patch and then stop caring. This would allow CentOS package maintainers to release just that one version of CherryPy RPM so that they'd have it in the system. After that, we can ignore pytest compat and whenever they solve their packaging problem they could proceed with packaging newer CherryPy versions.
This trade-off sounds acceptable to me as a one-time solution.

jaraco:
I admit. I didn’t read the whole thread. Yes, that approach seems suitable. In that case maybe we should consider just tagging a specialized release for CentOS. But then I wonder, why not just apply this patch downstream in the CentOS package?

alfredodeza:
Thanks all for your suggestions, since this PR was getting a lot of push back, I created a patch file to make the fixture changes which has allowed to build CherryPy for EPEL8.

I will not pursue this PR further.

jaraco:
I think that's the best approach - it gives you control and avoids entangling with the CherryPy release process.

~~In the spirit of goodwill, I also cherry-picked your commit onto v18.4.0 and tagged it as [v18.4.0.post.centos8.p1](https://github.com/cherrypy/cherrypy/releases/tag/v18.4.0.post.centos8.p1), which [should create a comparable release](https://travis-ci.org/cherrypy/cherrypy/jobs/617780985) in PyPI, so if you'd like to link to that release, you're welcome to do so.~~

webknjaz:
@jaraco no, such version format is forbidden in PyPI just like local version parts (that's why the Travis job failed). We'd have to do just `.post0`.

jaraco:
Yeah, you're right. I was actually thinking of a [Local Version Identifier](https://www.python.org/dev/peps/pep-0440/#id25), but that wouldn't be allowed in PyPI either (probably for the best). So nevermind.

alfredodeza:
Thanks @jaraco and @webknjaz for the effort! But there is no need to cut a release, I've already linked to the issue raised so that the patch is approved which should be sufficient in this case.

Issue #1827, opened 2019-11-24
Dictionary entries by class HeaderReader are strange


bmxp:
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐞 **Describe the bug. What is the current behavior?**

From cheroot server the class HeaderReader gives strange dict entries like ``{"B'Status'": b'302 Found' }`` when using a predefined dictionary like the one in ``cherrypy.serving.response.headers``.

❓ **What is the motivation / use case for changing the behavior?**
The result should not contain these strange dictionary entries.

💡 **To Reproduce**

Consider the following code:
```python
import tempfile
import cherrypy
from cheroot import server

response = tempfile.SpooledTemporaryFile(max_size = 1000, mode = "w+b")

response.write(b'Status: 302 Found\r\nLocation: index.php?\r\nContent-type: text/html; charset=UTF-8\r\n\r\n')

emptydict ={} 
response.seek(0)
hr = server.HeaderReader()
headers = hr(response,cherrypy.serving.response.headers)
print(headers)

print(80*'-')

response.seek(0)
headers = hr(response,emptydict)
print(headers)


print(80*'-')

otherdict ={ 'Content-Type':'text/html'} 
response.seek(0)
headers = hr(response,otherdict)
print(headers)
```
This will result in output like
```console
{'Content-Type': 'text/html', 'Server': 'CherryPy/18.4.0', 'Date': 'Sat, 23 Nov 2019 20:10:51 GMT', "B'Status'": b'302 Found', "B'Location'": b'index.php?', "B'Content-Type'": b'text/html; charset=UTF-8'}
--------------------------------------------------------------------------------
{b'Status': b'302 Found', b'Location': b'index.php?', b'Content-Type': b'text/html; charset=UTF-8'}
--------------------------------------------------------------------------------
{'Content-Type': 'text/html', b'Status': b'302 Found', b'Location': b'index.php?', b'Content-Type': b'text/html; charset=UTF-8'}
```

💡 **Expected behavior**

The first printed dict is not ok IMHO. The second and third printed dicts show the expected entries in dict.
A lookup like ``if 'Status' in dict ...`` will thus fail and a workaround ist needed.

📋 **Environment**

- Cheroot version: 8.2.1
- CherryPy version: 18.4.0
- Python version: 3.7.3
- OS: Debian 10 (buster)
- Browser: not relevant

[...]

PR #1828, opened 2019-11-27
Potential fix for HeaderMap issues

jaraco:
  - [X] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

#1827

# Discussion

I'm not sure this fix is the right one. I did not attempt to resolve what values _should_ be allowed for keys to a `HeaderMap`, only to address the issue that _if_ they're bytes, they're treated duck-like strings. Perhaps the right thing to do is just crash on non-text keys and require the callers to decode any bytes values.

I've called out `None` values as a special case, as tests fail without them. Perhaps that should be an error too, but for now, because the tests demand it, they're supported.

### DIFF
--- a/cherrypy/lib/httputil.py
+++ b/cherrypy/lib/httputil.py
@@ -456,7 +456,7 @@ class CaseInsensitiveDict(KeyTransformingDict):
     @staticmethod
     def transform_key(key):
         if key is None:
-            # TODO: why?
+            # TODO(#1830): why?
             return 'None'
         return key.title()
 
--- a/cherrypy/lib/httputil.py
+++ b/cherrypy/lib/httputil.py
@@ -450,12 +450,15 @@ class CaseInsensitiveDict(KeyTransformingDict):
 
     """A case-insensitive dict subclass.
 
-    Each key is changed on entry to str(key).title().
+    Each key is changed on entry to title case.
     """
 
     @staticmethod
     def transform_key(key):
-        return str(key).title()
+        if key is None:
+            # TODO: why?
+            return 'None'
+        return key.title()
 
 
 #   TEXT = <any OCTET except CTLs, but including LWS>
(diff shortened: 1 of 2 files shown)


cherrypy/lib/httputil.py
```diff
@@ -450,12 +450,15 @@ class CaseInsensitiveDict(KeyTransformingDict):
...
-    Each key is changed on entry to str(key).title().
+    Each key is changed on entry to title case.
     """
 
     @staticmethod
     def transform_key(key):
-        return str(key).title()
+        if key is None:
+            # TODO: why?
+            return 'None'
```
webknjaz, line 401: This should probably be
```suggestion
            raise ValueError
```

webknjaz:
@jaraco I think we should make the public API of HeaderMap compatible with [multidict] which would allow us to use it as a speed-up option.

[multidict]: https://pypi.org/project/multidict/

cherrypy/lib/httputil.py
jaraco: Have you tried that? Lots of tests fail when key is `None`. I'd like to see that issue fixed too, but it's outside the scope of the reported issue.

jaraco:
> I think we should make the public API of HeaderMap compatible with [multidict](https://pypi.org/project/multidict/) which would allow us to use it as a speed-up option.

I appreciate the reference. That seems reasonable, but I'm pretty sure that class has a public interface, so you'll need to support both interfaces for some time. And it's not obvious to me it will be possible. In any case, that sounds like a feature request separate from the issue at hand. Feel free to file one (and implement it).

cherrypy/lib/httputil.py
webknjaz: Oh. Then this is worth mentioning in a comment.
This should probably be
```suggestion
            # NOTE: Setting `raise ValueError` here causes many tests
            # to fail so before refactoring we'll need to fix that.
            return 'None'
```

jaraco: Right. I was trying to be more concise with `TODO: why?`... and let the failing tests tell the rest of the story until a ticket could be filed about it. With this change, we have both a TODO and a NOTE. That seems excessive, but thanks for the suggestion.

jaraco: In b07830a1, I've updated the TODO to reference a new issue (#1830).

Issue #1345, opened 2014-12-08
Can't explicitly release FileSessions


ghost:
Originally reported by: **tilboerner (Bitbucket: [tilboerner](http://bitbucket.org/tilboerner), GitHub: [tilboerner](http://github.com/tilboerner))**

---

A [change](https://bitbucket.org/cherrypy/cherrypy/commits/c5f10269f7ed25ff0c85d246c5483c3a6f0f2b60?at=default) introduced in CherryPy 3.3.0 makes it impossible to explicitly save or release the current session, if FileSessions are used.

This happens because FileSessions [assert](https://bitbucket.org/cherrypy/cherrypy/src/6279bd60ca3b51dbe16e4a28eff3659c223d559c/cherrypy/lib/sessions.py?at=default#cl-518) that they are locked on `_save`, and the SessionTool adds a [hook](https://bitbucket.org/cherrypy/cherrypy/src/6279bd60ca3b51dbe16e4a28eff3659c223d559c/cherrypy/_cptools.py?at=default#cl-315) to call `sessions.save()` before the response is finalized.

If a controller explicitly releases the session (for instance, before a time-consuming operation), CherryPy ultimately raises an AssertionError and returns an error response with status `500`.

This result runs contrary to the [session locking documentation](http://cherrypy.readthedocs.org/en/latest/pkg/cherrypy.lib.html#locking-sessions), which suggests that it is possible to explicitly call `cherrypy.session.release_lock()`.

To resolve this issue, either the behavior of automatic session release should be changed to accomodate already released sessions, or the documentation should be updated with instructions detailing how such a case should be handled.

Example AssertionError:

```
#!python
File "/.../lib/python3.4/site-packages/cherrypy/_cprequest.py", line 674, in respond
    self.hooks.run('before_finalize')
  File "/.../lib/python3.4/site-packages/cherrypy/_cprequest.py", line 114, in run
    raise exc
  File "/.../lib/python3.4/site-packages/cherrypy/_cprequest.py", line 104, in run
    hook()
  File "/.../lib/python3.4/site-packages/cherrypy/_cprequest.py", line 63, in __call__
    return self.callback(**self.kwargs)
  File "/.../lib/python3.4/site-packages/cherrypy/lib/sessions.py", line 784, in save
    cherrypy.session.save()
  File "/.../lib/python3.4/site-packages/cherrypy/lib/sessions.py", line 253, in save
    self._save(expiration_time)
[...]

PR #1834, opened 2019-12-15
Fix explicitly releasing FileSessions

dj51d:
  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

#1345 

#1345 

**Other information**:

**Checklist**:

### DIFF
--- a/cherrypy/lib/sessions.py
+++ b/cherrypy/lib/sessions.py
@@ -529,6 +529,8 @@ class FileSession(Session):
             return None
 
     def _save(self, expiration_time):
+        if getattr(self, 'locking', '') == 'explicit' and not self.locked:
+            self.acquire_lock()
         assert self.locked, ('The session was saved without being locked.  '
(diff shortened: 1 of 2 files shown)


Sxderp:
I'm surprised this hasn't been resolved. But I do think this should not be limited to "explicit" locking [1]. "implicit" and "early" would also greatly benefit from this. The doc seems to imply that you can call `cherrypy.session.release_lock()` for long-running requests while still using "early" / "implicit", which would cause the save error encountered here.

[1] I do find it weird that "explicit" locking has an "implicit" lock here. It would seem "explicit" should be the one method were the lock is not reacquired. /shrug

webknjaz:
@Sxderp it's a very old mechanism and IIRC there's some weird race condition that nobody's been able to figure out for a long time, that's making the tests flaky... Hence, long review time.

Issue #1772, opened 2019-02-26
Strange behavior when script_name does not start with /


Arvedui:
**I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**
cherrypy returns 404 instead of the requested page.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**
``` python
import cherrypy
class Test():

    @cherrypy.expose
    def bar(self,):
        return "foobar"

cherrypy.quickstart(Test(), 'foo')
```

Run this and try to open `localhost:8080/foo/bar`, it will return 404 without a payload.

**What is the expected behavior?**
I would expect it to return `foobar`

**Please tell us about your environment:**

- Cheroot version: 6.5.4
- CherryPy version: 18.1.0
- Python version: 3.7.2
- OS: archlinux

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)

The problem occurs because of an off by one error here:
https://github.com/cherrypy/cherrypy/blob/9d21bda648dc804b8a330ffd24439c823379bc14/cherrypy/_cprequest.py#L591

In my example `self.script_name` is "foo", `path` is "/foo/var" which results in `pi` = "o/bar".

I realize that the documentation says script_name *should* start with /, but that does not mean it must.
I think there should at least be an error message for this.

webknjaz:
Hi @Arvedui,

Thanks for the report! We'll gladly accept any PRs improving this :)

Issue #1750, opened 2018-10-23
File-based sessions on Windows, locking issues


pannal:
* **I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

* **Do you want to request a *feature* or report a *bug*?**
Report a bug.

* **What is the current behavior?**
This is the setup:
  * A CherryPy application that serves assets through a defined handler, which makes `requests.Request` requests to another HTTP server
  * Both the default handler and the assets handler make use of sessions
  * When the default handler is hit, there are intermediate requests to the assets handler from the browser, which might occur while the default handler is still active

When using `cherrypy.lib.sessions.FileSession` on a Windows environment, there seem to be issues in this particular setup.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**
```python-traceback
[23/Oct/2018:15:34:36]
Traceback (most recent call last):
  File "C:\Python35\lib\site-packages\cheroot\wsgi.py", line 147, in respond
    self.write(chunk)
  File "C:\Python35\lib\site-packages\cheroot\wsgi.py", line 228, in write
    self.req.write(chunk)
  File "C:\Python35\lib\site-packages\cheroot\server.py", line 1089, in write
    self.conn.wfile.write(chunk)
  File "C:\Python35\lib\site-packages\cheroot\makefile.py", line 31, in write
    self._flush_unlocked()
  File "C:\Python35\lib\site-packages\cheroot\makefile.py", line 40, in _flush_unlocked
    n = self.raw.write(bytes(self._write_buf))
  File "C:\Python35\lib\socket.py", line 594, in write
    return self._sock.send(b)
ConnectionAbortedError: [WinError 10053] An established connection was aborted by the software in your host machine

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python35\lib\site-packages\cherrypy\_cptree.py", line 160, in release_serving
    req.close()
  File "C:\Python35\lib\site-packages\cherrypy\_cprequest.py", line 511, in close
    self.hooks.run('on_end_request')
  File "C:\Python35\lib\site-packages\cherrypy\_cprequest.py", line 114, in run
    raise exc
  File "C:\Python35\lib\site-packages\cherrypy\_cprequest.py", line 104, in run
    hook()
  File "C:\Python35\lib\site-packages\cherrypy\_cprequest.py", line 63, in __call__
[...]

PR #1835, opened 2019-12-28
fix script_name without / at beginning

abtinmo:

  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

#1772 

#1772 

1.  shows an error, telling: script_name foo should start with /

2. fixes the script_name to prevent 404 error (not sure if it's necessary)

**Other information**:

**Checklist**:


webknjaz:
Hey @abtinmo, thanks for the PR!

Please **add a test for this change**. And make sure the CI is green.

abtinmo:
Hi @webknjaz,

it seems there is a problem with your macos config for ci.
and failed with  AppVeyor is for #1750

webknjaz:
@abtinmo don't worry Circle CI and that flaky test. Just add a test case for your change.

abtinmo:
@webknjaz its done.

cherrypy/test/test_core.py
```diff
@@ -824,3 +824,19 @@ def test_bind_ephemeral_port(self):
...
+class TestScriptName(helper.CPWebCase):
+    @staticmethod
+    def setup_server():
+        class Root:
+
+            @cherrypy.expose
+            def bar(self):
+                return 'foobar'
+
+        cherrypy.tree.mount(Root(), script_name='foo')
```
webknjaz, line 839: Please also test `script_name=None`, `script_name=''` and `script_name='/'`.

It looks like it changes the behavior described in the docstring of `Tree.mount()`:
>         script_name
>             ... 
>
>             It MUST NOT end in a slash. If the script_name refers to the
>             root of the URI, it MUST be an empty string (not "/").

webknjaz (changes requested):

cherrypy/_cptree.py
```diff
@@ -217,6 +217,11 @@ def mount(self, root, script_name='', config=None):
                 'request). You cannot mount such Applications on this Tree; '
                 'you must pass them to a WSGI server interface directly.')
 
+        if script_name and not script_name.startswith('/'):
+            cherrypy.log.error(
+                msg='script_name {0} should start with /'.format(script_name)
+            )
+            script_name = '/' + script_name
```
webknjaz, line 224: I'm not sure if it's a good idea for nested trees. Could you please check this too?

webknjaz:
This change feels dangerous so I would also require a clear approval from @jaraco, not just myself.

cherrypy/test/test_core.py
jaraco: Normally, I'd be reluctant to ask someone to add tests for behavior outside the scope of the behavior they're trying to support, but in this case, I'm pretty sure there's a regression for untested behavior here, so adding those tests is important. If your `git` foo is good enough, consider adding the tests against master first, so you can see the expected passing behavior on the examples that webknjaz suggested, but also the failing behavior on `script_name='foo'`.

cherrypy/_cptree.py
jaraco, line 224: I'm disinclined to continue to let the user supply an erroneous value and correct it for them. Why not instead hard-fail (raise an exception) when invalid values are input (similar to how `script_name=None` is handled)?

abtinmo: fixes the script_name to prevent 404 error (not sure if it's necessary) , its what i said in my first comment. 
i don't think letting user to continue is a right solution. but in doc i saw this:
This should start with a slash

so i choose to fix error myself and throw a console error.
raise an exception is indeed a better solution

webknjaz (changes requested):
AFAIR we discussed raising an exception for the invalid input. I'll wait for this change before attempting to review again.

jaraco:
It's been a while since I last reviewed this change.

I remember many years ago trying to understand what is the purpose of this code and fixing bugs around it, so I'm uneasy adding more uncertainty to the implementation.

As I understand, we discussed and agreed to make the invalid input an error (Exception) instead of a warning, but there hasn't been any activity since, so I'm closing this as languishing. Feel free to revisit with a new PR.

PR #1851, opened 2020-04-15
Fix serving files with diacritics in name under their original name

### LINKED ISSUE #1776: Cherrypy doesn`t encode UTF-8 filename in HTTP header "Content-Disposition" and returns "download" browser default filename instead.

jrxpert:
This implements adding extra ``filename*`` parameter in Content-disposition
header when serving content with non-ASCII filenames. As per RFC 6266
recommendations in the Appendix D.

Refs:
* https://stackoverflow.com/a/8996249/2173868
* https://tools.ietf.org/html/rfc6266#appendix-D

Fixes #1776

  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

Fixes #1776

#1776

**Other information**:
Previous pull request #1777 was closed due hard reset of the fork base to current version of cherrypy.

**Checklist**:

### DIFF
--- a/cherrypy/lib/static.py
+++ b/cherrypy/lib/static.py
@@ -35,12 +35,12 @@ def _make_content_disposition(disposition, file_name):
     This function implements the recommendations of :rfc:`6266#appendix-D`.
     See this and related answers: https://stackoverflow.com/a/8996249/2173868.
     """
-    # As normalization algorithm for `unicodedata` is used composed form (NFC and NFKC)
-    # with compatibility equivalence criteria (NFK), so "NFKC" is the one.
-    # It first applies the compatibility decomposition, followed by the canonical
-    # composition. Should be displayed in the same manner, should be treated in
-    # the same way by applications such as alphabetizing names or searching,
-    # and may be substituted for each other.
+    # As normalization algorithm for `unicodedata` is used composed form (NFC
+    # and NFKC) with compatibility equivalence criteria (NFK), so "NFKC" is the
+    # one. It first applies the compatibility decomposition, followed by the
+    # canonical composition. Should be displayed in the same manner, should be
+    # treated in the same way by applications such as alphabetizing names or
+    # searching, and may be substituted for each other.
     # See: https://en.wikipedia.org/wiki/Unicode_equivalence.
     ascii_name = (
         unicodedata.normalize('NFKC', file_name).
--- a/cherrypy/lib/static.py
+++ b/cherrypy/lib/static.py
@@ -39,10 +39,10 @@ def _make_content_disposition(disposition, file_name):
         unicodedata.normalize('NFKD', file_name).
         encode('ascii', errors='ignore').decode()
     )
-    header = u'{}; filename="{}"'.format(disposition, ascii_name)
+    header = '{}; filename="{}"'.format(disposition, ascii_name)
     if ascii_name != file_name:
         quoted_name = urllib.parse.quote(file_name)
-        header += u'; filename*=UTF-8\'\'{}'.format(quoted_name)
+        header += '; filename*=UTF-8\'\'{}'.format(quoted_name)
     return header
 
 
--- a/cherrypy/lib/static.py
+++ b/cherrypy/lib/static.py
@@ -30,9 +30,10 @@ _setup_mimetypes()
 
 
 def _utf8_content_disposition(disposition, file_name):
-    """Create HTTP header for downloading a file with UTF-8 filename.
-    See this and related answers:
-    https://stackoverflow.com/a/8996249/2173868
+    """Create HTTP header for downloading a file with a UTF-8 filename.
+
+    This function implements the recommendations of :rfc:`6266#appendix-D`.
+    See this and related answers: https://stackoverflow.com/a/8996249/2173868.
     """
     ascii_name = (
         unicodedata.normalize('NFKD', file_name).
--- a/cherrypy/lib/static.py
+++ b/cherrypy/lib/static.py
@@ -29,7 +29,7 @@ def _setup_mimetypes():
 _setup_mimetypes()
 
 
-def _utf8_content_disposition(disposition, file_name):
+def _make_content_disposition(disposition, file_name):
     """Create HTTP header for downloading a file with a UTF-8 filename.
 
     This function implements the recommendations of :rfc:`6266#appendix-D`.
@@ -110,7 +110,7 @@ def serve_file(path, content_type=None, disposition=None, name=None,
     if disposition is not None:
         if name is None:
             name = os.path.basename(path)
-        cd = _utf8_content_disposition(disposition, name)
+        cd = _make_content_disposition(disposition, name)
         response.headers['Content-Disposition'] = cd
     if debug:
         cherrypy.log('Content-Disposition: %r' % cd, 'TOOLS.STATIC')
@@ -168,7 +168,7 @@ def serve_fileobj(fileobj, content_type=None, disposition=None, name=None,
         if name is None:
             cd = disposition
         else:
-            cd = _utf8_content_disposition(disposition, name)
+            cd = _make_content_disposition(disposition, name)
         response.headers['Content-Disposition'] = cd
     if debug:
         cherrypy.log('Content-Disposition: %r' % cd, 'TOOLS.STATIC')
--- a/cherrypy/lib/static.py
+++ b/cherrypy/lib/static.py
@@ -28,6 +28,22 @@ def _setup_mimetypes():
 _setup_mimetypes()
 
 
+def _utf8_content_disposition(disposition, file_name):
+    """Create HTTP header for downloading a file with UTF-8 filename.
+
+    See this and related answers:
+    https://stackoverflow.com/a/8996249/2173868
+    """
+    ascii_name = (
+        unicodedata.normalize('NFKD', file_name).
+        encode('ascii', errors='ignore').decode()
+    )
+    quoted_name = urllib.parse.quote(file_name)
+    header = u'{}; filename="{}"'.format(disposition, ascii_name)
+    header += u'; filename*=UTF-8\'\'{}'.format(quoted_name)
+    return header
+
+
 def serve_file(path, content_type=None, disposition=None, name=None,
                debug=False):
     """Set status, headers, and body in order to serve the given path.
@@ -37,9 +53,10 @@ def serve_file(path, content_type=None, disposition=None, name=None,
     of the 'path' argument.
 
     If disposition is not None, the Content-Disposition header will be set
-    to "<disposition>; filename=<name>". If name is None, it will be set
-    to the basename of path. If disposition is None, no Content-Disposition
-    header will be written.
+    to "<disposition>; filename=<name>; filename*=utf-8''<name>"
+    as described in RFC6266 (https://tools.ietf.org/html/rfc6266#appendix-D).
+    If name is None, it will be set to the basename of path.
+    If disposition is None, no Content-Disposition header will be written.
     """
     response = cherrypy.serving.response
 
@@ -92,7 +109,7 @@ def serve_file(path, content_type=None, disposition=None, name=None,
     if disposition is not None:
         if name is None:
             name = os.path.basename(path)
-        cd = '%s; filename="%s"' % (disposition, name)
+        cd = _utf8_content_disposition(disposition, name)
         response.headers['Content-Disposition'] = cd
     if debug:
         cherrypy.log('Content-Disposition: %r' % cd, 'TOOLS.STATIC')
@@ -111,9 +128,10 @@ def serve_fileobj(fileobj, content_type=None, disposition=None, name=None,
     The Content-Type header will be set to the content_type arg, if provided.
 
     If disposition is not None, the Content-Disposition header will be set
-    to "<disposition>; filename=<name>". If name is None, 'filename' will
-    not be set. If disposition is None, no Content-Disposition header will
-    be written.
+    to "<disposition>; filename=<name>; filename*=utf-8''<name>"
+    as described in RFC6266 (https://tools.ietf.org/html/rfc6266#appendix-D).
+    If name is None, 'filename' will not be set.
+    If disposition is None, no Content-Disposition header will be written.
 
     CAUTION: If the request contains a 'Range' header, one or more seek()s will
     be performed on the file object.  This may cause undesired behavior if
@@ -149,7 +167,7 @@ def serve_fileobj(fileobj, content_type=None, disposition=None, name=None,
         if name is None:
             cd = disposition
         else:
-            cd = '%s; filename="%s"' % (disposition, name)
+            cd = _utf8_content_disposition(disposition, name)
         response.headers['Content-Disposition'] = cd
     if debug:
         cherrypy.log('Content-Disposition: %r' % cd, 'TOOLS.STATIC')
--- a/cherrypy/lib/static.py
+++ b/cherrypy/lib/static.py
@@ -39,9 +39,10 @@ def _make_content_disposition(disposition, file_name):
         unicodedata.normalize('NFKD', file_name).
         encode('ascii', errors='ignore').decode()
     )
-    quoted_name = urllib.parse.quote(file_name)
     header = u'{}; filename="{}"'.format(disposition, ascii_name)
-    header += u'; filename*=UTF-8\'\'{}'.format(quoted_name)
+    if ascii_name != file_name:
+        quoted_name = urllib.parse.quote(file_name)
+        header += u'; filename*=UTF-8\'\'{}'.format(quoted_name)
     return header
 
 
--- a/cherrypy/lib/static.py
+++ b/cherrypy/lib/static.py
@@ -28,22 +28,6 @@ def _setup_mimetypes():
 _setup_mimetypes()
 
 
-def _utf8_content_disposition(disposition, file_name):
-    """Create HTTP header for downloading a file with UTF-8 filename.
-
-    See this and related answers:
-    https://stackoverflow.com/a/8996249/2173868
-    """
-    ascii_name = (
-        unicodedata.normalize('NFKD', file_name).
-        encode('ascii', errors='ignore').decode()
-    )
-    quoted_name = urllib.parse.quote(file_name)
-    header = u'{}; filename="{}"'.format(disposition, ascii_name)
-    header += u'; filename*=UTF-8\'\'{}'.format(quoted_name)
-    return header
-
-
 def serve_file(path, content_type=None, disposition=None, name=None,
                debug=False):
     """Set status, headers, and body in order to serve the given path.
@@ -53,10 +37,9 @@ def serve_file(path, content_type=None, disposition=None, name=None,
     of the 'path' argument.
 
     If disposition is not None, the Content-Disposition header will be set
-    to "<disposition>; filename=<name>; filename*=utf-8''<name>"
-    as described in RFC6266 (https://tools.ietf.org/html/rfc6266#appendix-D).
-    If name is None, it will be set to the basename of path.
-    If disposition is None, no Content-Disposition header will be written.
+    to "<disposition>; filename=<name>". If name is None, it will be set
+    to the basename of path. If disposition is None, no Content-Disposition
+    header will be written.
     """
     response = cherrypy.serving.response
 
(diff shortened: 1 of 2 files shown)


cherrypy/test/test_static.py
```diff
@@ -193,6 +209,19 @@ def test_static(self):
         #   we just check the content
         self.assertMatchesBody('^Dummy stylesheet')
 
+        # Check a filename with utf-8 characters in it
+        ascii_fn = 'has_utf-8_character_.html'
+        url_quote_fn = 'has_utf-8_character_%E2%98%83.html'
+        cd = '''attachment; filename="%s"; filename*=UTF-8\'\'%s'''
```
webknjaz: no need to have a multiline literal here:
```suggestion
        cd = 'attachment; filename="%s"; filename*=UTF-8\'\'%s'
```

cherrypy/test/test_static.py
```diff
@@ -193,6 +209,19 @@ def test_static(self):
         #   we just check the content
         self.assertMatchesBody('^Dummy stylesheet')
 
+        # Check a filename with utf-8 characters in it
+        ascii_fn = 'has_utf-8_character_.html'
+        url_quote_fn = 'has_utf-8_character_%E2%98%83.html'
```
webknjaz:
```suggestion
        url_quote_fn = 'has_utf-8_character_%E2%98%83.html'  # %E2%98%83 == ☃
```

cherrypy/test/test_static.py
```diff
@@ -97,6 +97,22 @@ def bytesio(self):
...
+                return static.serve_file(
+                    file_path,
+                    disposition='attachment',
+                    name='has_utf-8_character_☃.html')
+
+            @cherrypy.expose
+            def serve_fileobj_utf8_filename(self):
+                f = open(os.path.join(curdir, 'style.css'), 'rb')
+                return static.serve_fileobj(
+                    f,
```
webknjaz: Let's not hit the disk here:
```suggestion
                return static.serve_fileobj(
                    io.BytesIO(b'☃\nfie\nfo\nfum'),
```

cherrypy/test/test_static.py
```diff
@@ -193,6 +209,19 @@ def test_static(self):
...
+        url_quote_fn = 'has_utf-8_character_%E2%98%83.html'
+        cd = '''attachment; filename="%s"; filename*=UTF-8\'\'%s'''
+
+        self.getPage('/serve_file_utf8_filename')
+        self.assertStatus('200 OK')
+        self.assertHeader('Content-Disposition', cd % (ascii_fn, url_quote_fn))
+
+        self.getPage('/serve_fileobj_utf8_filename')
+        self.assertStatus('200 OK')
+        self.assertHeader('Content-Disposition', cd % (ascii_fn, url_quote_fn))
```
webknjaz: no need to have a multiline literal here, also let's use `str.format()`:
```suggestion
        expected_content_disposition = (
            'attachment; filename="{!s}"; filename*=UTF-8\'\'{!s}'.
            format(ascii_fn, url_quote_fn)
        )

        self.getPage('/serve_file_utf8_filename')
        self.assertStatus('200 OK')
        self.assertHeader('Content-Disposition', expected_content_disposition)

        self.getPage('/serve_fileobj_utf8_filename')
        self.assertStatus('200 OK')
        self.assertHeader('Content-Disposition', expected_content_disposition)
```

cherrypy/lib/static.py
```diff
@@ -28,6 +29,21 @@ def _setup_mimetypes():
...
+    """Create HTTP header for downloading a file with UTF-8 filename.
+    See this and related answers:
+    https://stackoverflow.com/a/8996249/2173868
+    """
+    ascii_name = (
+        unicodedata.normalize('NFKD', file_name).
+        encode('ascii', errors='ignore').decode()
+    )
+    quoted_name = urllib.parse.quote(file_name)
+    header = u'{}; filename="{}"'.format(disposition, ascii_name)
```
webknjaz: Appendix D says:
>    o  Include a "filename" parameter when US-ASCII ([US-ASCII]) is
>       sufficiently expressive.

Does this mean that `filename=` should be omitted if it's not expressive? Maybe when `not ascii_name.trim()`?

cherrypy/lib/static.py
```diff
@@ -28,6 +29,21 @@ def _setup_mimetypes():
 _setup_mimetypes()
 
 
+def _utf8_content_disposition(disposition, file_name):
+    """Create HTTP header for downloading a file with UTF-8 filename.
+    See this and related answers:
+    https://stackoverflow.com/a/8996249/2173868
+    """
+    ascii_name = (
+        unicodedata.normalize('NFKD', file_name).
```
webknjaz: I think this needs a code comment explaining why specifically `NKFD` is used here and not some other normalization type.

cherrypy/lib/static.py
```diff
@@ -28,6 +29,21 @@ def _setup_mimetypes():
 _setup_mimetypes()
 
 
+def _utf8_content_disposition(disposition, file_name):
+    """Create HTTP header for downloading a file with UTF-8 filename.
+    See this and related answers:
+    https://stackoverflow.com/a/8996249/2173868
+    """
```
webknjaz: As per PEP257, the docstring title should be followed by an empty line:
```suggestion
    """Create HTTP header for downloading a file with a UTF-8 filename.

    This function implements the recommendations of :rfc:`6266#appendix-D`.
    See this and related answers: https://stackoverflow.com/a/8996249/2173868.
    """
```

cherrypy/lib/static.py
```diff
@@ -37,9 +53,10 @@ def serve_file(path, content_type=None, disposition=None, name=None,
     of the 'path' argument.
 
     If disposition is not None, the Content-Disposition header will be set
-    to "<disposition>; filename=<name>". If name is None, it will be set
-    to the basename of path. If disposition is None, no Content-Disposition
-    header will be written.
+    to "<disposition>; filename=<name>; filename*=utf-8''<name>"
+    as described in RFC6266 (https://tools.ietf.org/html/rfc6266#appendix-D).
```
webknjaz: We can use a sphinx role for this:
```suggestion
    as described in :rfc:`6266#appendix-D`.
```

cherrypy/lib/static.py
```diff
@@ -111,9 +128,10 @@ def serve_fileobj(fileobj, content_type=None, disposition=None, name=None,
     The Content-Type header will be set to the content_type arg, if provided.
 
     If disposition is not None, the Content-Disposition header will be set
-    to "<disposition>; filename=<name>". If name is None, 'filename' will
-    not be set. If disposition is None, no Content-Disposition header will
-    be written.
+    to "<disposition>; filename=<name>; filename*=utf-8''<name>"
+    as described in RFC6266 (https://tools.ietf.org/html/rfc6266#appendix-D).
```
webknjaz:
```suggestion
    as described in :rfc:`6266#appendix-D`.
```

webknjaz:
> Previous pull request #1777 was closed due hard reset of the fork base to current version of cherrypy.

To avoid this in the future, I recommend having separate (non-master) branches for PRs.

cherrypy/lib/static.py
```diff
@@ -28,6 +29,21 @@ def _setup_mimetypes():
...
+    See this and related answers:
+    https://stackoverflow.com/a/8996249/2173868
+    """
+    ascii_name = (
+        unicodedata.normalize('NFKD', file_name).
+        encode('ascii', errors='ignore').decode()
+    )
+    quoted_name = urllib.parse.quote(file_name)
+    header = u'{}; filename="{}"'.format(disposition, ascii_name)
+    header += u'; filename*=UTF-8\'\'{}'.format(quoted_name)
```
webknjaz: >    o  Include a "filename*" parameter where the desired filename cannot
      be expressed faithfully using the "filename" form.  Note that
      legacy user agents will not process this, and will fall back to
      using the "filename" parameter's content.

Looks like `filename*` should only be included when different from `filename`.

cherrypy/lib/static.py
```diff
@@ -28,6 +29,22 @@ def _setup_mimetypes():
 _setup_mimetypes()
 
 
+def _utf8_content_disposition(disposition, file_name):
```
webknjaz: I think the function name can be improved here:
1) functions (unlike variables) do some actions which should be reflected in their name with a verb, the current name looks rather like a property.
2) "utf8" seems to be an implementation detail from the caller's PoV so I'd drop it.


```suggestion
def _make_content_disposition(disposition, file_name):
```

cherrypy/lib/static.py
```diff
@@ -92,7 +110,7 @@ def serve_file(path, content_type=None, disposition=None, name=None,
     if disposition is not None:
         if name is None:
             name = os.path.basename(path)
-        cd = '%s; filename="%s"' % (disposition, name)
+        cd = _utf8_content_disposition(disposition, name)
```
webknjaz:
```suggestion
        cd = _make_content_disposition(disposition, name)
```

cherrypy/lib/static.py
```diff
@@ -149,7 +168,7 @@ def serve_fileobj(fileobj, content_type=None, disposition=None, name=None,
         if name is None:
             cd = disposition
         else:
-            cd = '%s; filename="%s"' % (disposition, name)
+            cd = _utf8_content_disposition(disposition, name)
```
webknjaz:
```suggestion
            cd = _make_content_disposition(disposition, name)
```

cherrypy/test/test_static.py
```diff
@@ -97,6 +97,21 @@ def bytesio(self):
                 f = io.BytesIO(b'Fee\nfie\nfo\nfum')
                 return static.serve_fileobj(f, content_type='text/plain')
 
+            @cherrypy.expose
+            def serve_file_utf8_filename(self):
+                file_path = os.path.join(curdir, 'style.css')
+                return static.serve_file(
+                    file_path,
```
webknjaz: How about just serving this test module?
```suggestion
                return static.serve_file(
                    __file__,
```

webknjaz:
@jrxpert I've committed a few cosmetic changes on top but some of the questions still need to be addressed. Do you mind replying to those comments?

cherrypy/test/test_static.py
```diff
@@ -97,6 +97,20 @@ def bytesio(self):
...
+            def serve_file_utf8_filename(self):
+                return static.serve_file(
+                    __file__,
+                    disposition='attachment',
+                    name='has_utf-8_character_☃.html')
+
+            @cherrypy.expose
+            def serve_fileobj_utf8_filename(self):
+                return static.serve_fileobj(
+                    io.BytesIO(b'☃\nfie\nfo\nfum'),
```
webknjaz:
```suggestion
                    io.BytesIO('☃\nfie\nfo\nfum'.encode('utf-8')),
```

cherrypy/lib/static.py
jrxpert: Yeah, this is true, i forgot it, that`s why in first pull request there was a condition before second part of header variable assignment, I will return it

cherrypy/lib/static.py
jrxpert: There is good description of basic differences on wiki:  https://en.wikipedia.org/wiki/Unicode_equivalence

https://docs.python.org/3.7/library/unicodedata.html

__NFC__: Normal form C (NFC) first applies a canonical decomposition, then composes pre-combined characters again.
vs.
__NFKC__: The normal form KC (NFKC) first applies the compatibility decomposition, followed by the canonical composition.
vs.
__NFD__: Normal form D (NFD) is also known as canonical decomposition, and translates each character into its decomposed form.
vs.
__NFKD__: The normal form KD (NFKD) will apply the compatibility decomposition, i.e. replace all compatibility characters with their equivalents.

= ???

It looks like better solution to use __NFKC__ finally, because of what they say on wiki "_Sequences that are canonically equivalent are also compatible, but the opposite is not necessarily true._". What do you think?

cherrypy/lib/static.py
jrxpert: I don't think so, if you ommit `filename=`, there will be no filename specified for browser.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

> __filename__
    Is followed by a string containing the original name of the file transmitted. The filename is always optional and must not be used blindly by the application: path information should be stripped, and conversion to the server file system rules should be done. This parameter provides mostly indicative information. When used in combination with Content-Disposition: attachment, it is used as the default filename for an eventual "Save As" dialog presented to the user.

webknjaz: Okay, agreed.

cherrypy/lib/static.py
webknjaz: Okay, `NFKC` it is. But make sure to add explanation and references as `# code comments`.

cherrypy/lib/static.py
```diff
@@ -28,6 +29,23 @@ def _setup_mimetypes():
...
+    See this and related answers: https://stackoverflow.com/a/8996249/2173868.
+    """
+    ascii_name = (
+        unicodedata.normalize('NFKD', file_name).
+        encode('ascii', errors='ignore').decode()
+    )
+    header = u'{}; filename="{}"'.format(disposition, ascii_name)
+    if ascii_name != file_name:
+        quoted_name = urllib.parse.quote(file_name)
+        header += u'; filename*=UTF-8\'\'{}'.format(quoted_name)
```
webknjaz: Since we no longer support Python 2, no need to use u-prefix in string literals
```suggestion
    header = '{}; filename="{}"'.format(disposition, ascii_name)
    if ascii_name != file_name:
        quoted_name = urllib.parse.quote(file_name)
        header += '; filename*=UTF-8\'\'{}'.format(quoted_name)
```

cherrypy/lib/static.py
```diff
@@ -28,6 +29,31 @@ def _setup_mimetypes():
 _setup_mimetypes()
 
 
+def _make_content_disposition(disposition, file_name):
+    """Create HTTP header for downloading a file with a UTF-8 filename.
+
+    This function implements the recommendations of :rfc:`6266#appendix-D`.
+    See this and related answers: https://stackoverflow.com/a/8996249/2173868.
+
+    As normalization algorithm for `unicodedata` is used composed form (NFC and NKFC)
```
webknjaz: I don't think this should go to the docstring but to a comment. Docstrings should only contain high-level usage implications information while comments answer to the question why a certain technical decision was made.

webknjaz: (also this exceeds the line limit as you can see from the linter reports which has to be fixed for use to be able to merge this)

jrxpert: I moved it below as code comment

webknjaz: Good, but it still needs to pass the linter checks.

webknjaz:
@jrxpert looks like besides linting (https://zuul.vexxhost.dev/t/cherrypy/build/79f7d6dc2ccc44078b36775cbd895787 / https://github.com/cherrypy/cherrypy/pull/1851/files#annotation_170326549 / https://travis-ci.org/github/cherrypy/cherrypy/jobs/676112806#L535-L536), this change also breaks some of the tests: https://ci.appveyor.com/project/CherryPy/cherrypy/builds/32246508/job/pj8nid12vp1yi991/tests

Here's one of the tracebacks there:
```python-traceback
  File "C:\projects\cherrypy\cherrypy\_cprequest.py", line 638, in respond
    self._do_respond(path_info)
  File "C:\projects\cherrypy\cherrypy\_cprequest.py", line 697, in _do_respond
    response.body = self.handler()
  File "C:\projects\cherrypy\cherrypy\lib\encoding.py", line 219, in __call__
    self.body = self.oldhandler(*args, **kwargs)
  File "C:\projects\cherrypy\cherrypy\_cpdispatch.py", line 54, in __call__
    return self.callable(*self.args, **self.kwargs)
  File "C:\projects\cherrypy\cherrypy\test\test_static.py", line 105, in serve_file_utf8_filename
    name='has_utf-8_character_\u2603.html')
  File "C:\projects\cherrypy\cherrypy\lib\static.py", line 121, in serve_file
    cd = _make_content_disposition(disposition, name)
  File "C:\projects\cherrypy\cherrypy\lib\static.py", line 46, in _make_content_disposition
    unicodedata.normalize('NKFC', file_name).
ValueError: invalid normalization form
```

cherrypy/lib/static.py
```diff
@@ -28,6 +29,30 @@ def _setup_mimetypes():
 _setup_mimetypes()
 
 
+def _make_content_disposition(disposition, file_name):
+    """Create HTTP header for downloading a file with a UTF-8 filename.
+
+    This function implements the recommendations of :rfc:`6266#appendix-D`.
+    See this and related answers: https://stackoverflow.com/a/8996249/2173868.
+    """
+    # As normalization algorithm for `unicodedata` is used composed form (NFC and NKFC)
```
webknjaz:
```suggestion
    # As normalization algorithm for `unicodedata` is used composed form (NFC and NFKC)
```

cherrypy/lib/static.py
```diff
@@ -28,6 +29,30 @@ def _setup_mimetypes():
...
+    # with compatibility equivalence criteria (NFK), so "NKFC" is the one.
```
webknjaz:
```suggestion
    # with compatibility equivalence criteria (NFK), so "NFKC" is the one.
```

cherrypy/lib/static.py
```diff
@@ -28,6 +29,30 @@ def _setup_mimetypes():
...
+    # It first applies the compatibility decomposition, followed by the canonical
+    # composition. Should be displayed in the same manner, should be treated in
+    # the same way by applications such as alphabetizing names or searching,
+    # and may be substituted for each other.
+    # See: https://en.wikipedia.org/wiki/Unicode_equivalence.
+    ascii_name = (
+        unicodedata.normalize('NKFC', file_name).
```
webknjaz: This is why the tests fail:
```suggestion
        unicodedata.normalize('NFKC', file_name).
```

webknjaz:
I've fixed those errors. Now I'm waiting for the CI statuses and will probably merge this PR later today.

webknjaz:
I rebased it, no need to have foxtrot merges here.

jrxpert:
> I've fixed those errors. Now I'm waiting for the CI statuses and will probably merge this PR later today.

pre-commit is working for me now, so I noticed error from it:
Fix double quoted strings................................................Failed
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 16055: character maps to <undefined>

I don't know if it has connection with new code and used %E2%98%83 character or not. Traceback points somewhere to cache, so it's little bit harder to find.

webknjaz:
@jrxpert do you run `pre-commit` directly or via `tox -e pre-commit`? Also are you sure it runs under Python 3? It's green in CIs.

jrxpert:
> @jrxpert do you run `pre-commit` directly or via `tox -e pre-commit`? Also are you sure it runs under Python 3?

via `tox -e pre-commit`, python 3.6

webknjaz:
> via `tox -e pre-commit`, python 3.6

What does `.tox/pre-commit/bin/python -V` say?

jrxpert:
> What does `.tox/pre-commit/bin/python -V` say?

Python 3.6.6

webknjaz:
I'd probably clean up `~/.pre-commit`, `~/.cache/pre-commit` and `.tox` and try again then. This looks weird and maybe some of the pre-cached pre-commit envs was created with py2 earlier and gets picked up now.

webknjaz:
Should be pip-installable in 10 min: https://pypi.org/project/CherryPy/18.6.0/

Issue #1849, opened 2020-04-11
ToolTests::testCombinedTools fails in NixOS


unode:
**I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

**What is the current behavior?**

Test `ToolTests::testCombinedTools` fails with the output mentioned in [this Nixpkgs issue](https://github.com/NixOS/nixpkgs/issues/84897#issuecomment-612383018).
The version listed is `18.3.0` but the failure is also reproducible in `18.5.0`.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**

1. Install [NixOS](https://nixos.org) or [nix](https://nixos.org/nix/) or use https://hub.docker.com/r/nixos/nix/
2. `git clone --depth 1 https://github.com/NixOS/nixpkgs`
3. `cd nixpkgs`
4. `git checkout 4e4e57bdd3bbfaaf27eab08c6502530c0f52d448`
5. `export NIXPKGS=$(pwd)`
6. `nix-build $NIXPKGS -A python37Packages.cherrypy`

**What is the expected behavior?**

Test `ToolTests::testCombinedTools` fails. All others pass (except 3 XPASS + 3 SKIPPED).

**Please tell us about your environment:**
```
platform linux -- Python 3.7.7, pytest-5.3.5, py-1.8.1, pluggy-0.13.1 -- /nix/store/vs4vj1yzqj1bkcqkf3b6sxm6jfy1gb4j-python3-3.7.7/bin/python3.7
cachedir: .pytest_cache
rootdir: /build/CherryPy-18.5.0, inifile: pytest.ini, testpaths: cherrypy/test/
plugins: cov-2.8.1, services-2.0.1
collected 298 items / 1 deselected / 297 selected
```
```
Requirement already satisfied: portend>=2.1.1 in /nix/store/8ynzwhpbl3f8hxx0w8x11b15izck1q1x-python3.7-portend-2.6/lib/python3.7/site-packages (from CherryPy==18.5.0) (2.6)
Requirement already satisfied: more-itertools in /nix/store/nlxnby9qcnykhbpaxk4wa3jdivj7mzb3-python3.7-more-itertools-8.0.2/lib/python3.7/site-packages (from CherryPy==18.5.0) (8.0.2)
[...]

PR #1877, opened 2020-11-12
CircleCI: Use supported versions of macOS Xcode

cclauss:
Related to ##1878

https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions

  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

CircleCI macOS tests fail.

**Other information**:

**Checklist**:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,8 +14,12 @@ jobs:
           export PATH="$PYENV_ROOT/bin:$PATH"
         ' >> $BASH_ENV
 
+    - run: pyenv install --list
+    # CFLAGS and LDFLAGS added for pypy3 build... Which still fails.
     - run: |-
-        for py_ver in 3.7.0 3.6.4 3.5.4 pypy3.5-6.0.0
+        CFLAGS="-I$(brew --prefix openssl)/include"
+        LDFLAGS="-L$(brew --prefix openssl)/lib"
+        for py_ver in 3.9.0 3.8.6 3.7.9 3.6.12 pypy3.5-6.0.0
         do
           pyenv install "$py_ver" &
         done
@@ -25,7 +29,7 @@ jobs:
     - run: python3 -m pip install --upgrade pip wheel
     - run: python3 -m pip install tox tox-pyenv
     - checkout
-    - run: tox -e py35,py36,py37,pypy3 -- -p no:sugar
+    - run: tox -e py36,py37,py38,py39,pypy3 -- -p no:sugar
     - store_test_results:
         path: .test-results
     - store_artifacts:
@@ -38,7 +42,7 @@ jobs:
     steps:
     - checkout
     - run: pip install tox
-    - run: tox -e py35,py36,py37
+    - run: tox -e py36,py37,py38  # Not yet... ,py39
     - store_test_results:
         path: .test-results
     - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,17 +14,13 @@ jobs:
           export PATH="$PYENV_ROOT/bin:$PATH"
         ' >> $BASH_ENV
 
-    - run: pyenv install --list
-    # CFLAGS and LDFLAGS added for pypy3 build... Which still fails.
     - run: |-
-        CFLAGS="-I$(brew --prefix openssl)/include"
-        LDFLAGS="-L$(brew --prefix openssl)/lib"
         for py_ver in 3.9.0 3.8.6 3.7.9 3.6.12 pypy3.5-6.0.0
         do
           pyenv install "$py_ver" &
         done
         wait
-    - run: pyenv global 3.7.0 3.6.4 3.5.4 pypy3.5-6.0.0
+    - run: pyenv global 3.9.0 3.8.6 3.7.9 3.6.12 pypy3.5-6.0.0
 
     - run: python3 -m pip install --upgrade pip wheel
     - run: python3 -m pip install tox tox-pyenv
--- a/cherrypy/test/test_bus.py
+++ b/cherrypy/test/test_bus.py
@@ -11,6 +11,7 @@ import pytest
 
 from cherrypy.process import wspbus
 
+
 CI_ON_MACOS = bool(os.getenv("CI")) and sys.platform == "darwin"
 msg = 'Listener %d on channel %s: %s.'  # pylint: disable=invalid-name
 
--- a/cherrypy/test/test_bus.py
+++ b/cherrypy/test/test_bus.py
@@ -11,7 +11,7 @@ import pytest
 
 from cherrypy.process import wspbus
 
-CI_ON_MACOS = bool(os.getenv('CI')) and sys.platform == "darwin"
+CI_ON_MACOS = bool(os.getenv("CI")) and sys.platform == "darwin"
 msg = 'Listener %d on channel %s: %s.'  # pylint: disable=invalid-name
 
 
--- a/cherrypy/test/test_bus.py
+++ b/cherrypy/test/test_bus.py
@@ -1,6 +1,8 @@
 """Publish-subscribe bus tests."""
 # pylint: disable=redefined-outer-name
 
+import os
+import sys
 import threading
 import time
 import unittest.mock
@@ -9,7 +11,7 @@ import pytest
 
 from cherrypy.process import wspbus
 
-
+CI_ON_MACOS = bool(os.getenv('CI')) and sys.platform == "darwin"
 msg = 'Listener %d on channel %s: %s.'  # pylint: disable=invalid-name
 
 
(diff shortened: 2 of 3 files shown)


cherrypy/test/test_bus.py
```diff
@@ -9,7 +11,7 @@
 
 from cherrypy.process import wspbus
 
-
+ci_on_macOS = bool(os.getenv('CI')) and sys.platform == "darwin"
```
webknjaz: Please use uppercase constant names

cherrypy/test/test_states.py
```diff
@@ -125,6 +125,9 @@ def exittest():
         engine.block()
         self.assertEqual(engine.state, engine.states.EXITING)
 
+    @pytest.mark.skipif(
```
webknjaz: Tests mustn't be skipped just because they fail. Instead, xfail should be used: https://pganssle-talks.github.io/xfail-lightning/

setup.py
```diff
@@ -31,7 +31,6 @@
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: Implementation',
         'Programming Language :: Python :: Implementation :: CPython',
-        'Programming Language :: Python :: Implementation :: Jython',
```
webknjaz: This change is wrong. A similar one has been rejected some time ago.

setup.py
```diff
@@ -82,15 +81,9 @@
         'testing': [
             'coverage',  # inspects tests coverage
             'codecov',   # sends tests coverage to codecov.io
-
```
webknjaz: don't remove the delimiter

webknjaz (changes requested):
Looks like this change is still WIP so I'll mark it as draft. Also, I've made a few notes about things that should be fixed

.circleci/config.yml
```diff
@@ -2,29 +2,37 @@ version: 2
...
     - run: |-
-        for py_ver in 3.7.0 3.6.4 3.5.4 pypy3.5-6.0.0
+        for py_ver in 3.9.0 3.8.6 3.7.9 3.6.12 pypy3.6-7.3.1
         do
           pyenv install "$py_ver" &
         done
         wait
-    - run: pyenv global 3.7.0 3.6.4 3.5.4 pypy3.5-6.0.0
-
-    - run: pip install tox tox-pyenv
```
webknjaz: all the empty line separators should be kept

setup.py
cclauss: Please explain.  This repo does not support Python 2 and Jython does not support Python 3.

cherrypy/test/test_states.py
cclauss: They don’t fail.  They hang so xfail would be a useless change.

setup.py
webknjaz: Jason's opinion is that if Jython 3 ever becomes stable, our code will support it. Anyway, this has nothing to do with Circle CI and so mustn't be in the same PR.

cherrypy/test/test_states.py
webknjaz: @cclauss `skipif` is still semantically wrong here. `xfail` has `run=False` exactly for this case.

P.S. https://github.com/cherrypy/cheroot/pull/331 that I'm planning to merge soon should fix this problem.

.circleci/config.yml
```diff
@@ -2,29 +2,37 @@ version: 2
...
+    - run: pyenv versions
+    - run: pyenv global 3.9.0 3.8.6 3.7.9 3.6.12 pypy3.6
```
webknjaz: I'd not drop `pypy3.5` as it's separate

cherrypy/test/test_states.py
webknjaz: P.P.S. Also, the best way to isolate this test is to use `pytest-forked` so that one test wouldn't crash the whole process.

setup.py
```diff
@@ -82,15 +81,9 @@
...
             # cherrypy.lib.gctools
             'objgraph',
-
-            # pytest >= 5.4.0 doesn't work with pytest-sugar <= 0.9.2
-            # once https://github.com/Teemu/pytest-sugar/pull/188 is
-            # merged, the upper bound can be removed.
-            # Ref: https://github.com/pytest-dev/pytest/issues/6931
-            'pytest>=5.3.5,<6.1.0',
+            'pytest',
```
webknjaz: Per comment, the lower bound shouldn't be removed. It should be in sync with that `pytest.ini` says.

cherrypy/test/test_tools.py
```diff
@@ -343,6 +347,7 @@ def testGuaranteedHooks(self):
         expected_msg = tmpl.format(attr='items')
         self.assertInBody(expected_msg)
 
+    @pytest.mark.skipif(ci_on_macOS, reason="continuous integration on macOS fails")
```
webknjaz: The problem here was unrelated to the OS but to a patch in CPython. I've fixed it on master via https://github.com/cherrypy/cherrypy/pull/1879 (see https://github.com/cherrypy/cherrypy/issues/1849 / https://bugs.python.org/issue39389 / https://docs.python.org/3.7/whatsnew/changelog.html#python-3-7-7-final).

But not that it's fixed, this mark shouldn't be necessary. Also, even if it was necessary, it'd be xfail, not skipif — this one doesn't crash the test runner.

cherrypy/test/test_states.py
cclauss: Removed because 331 should fix this.

setup.py
cclauss: LOL

cherrypy/test/test_tools.py
cclauss: Removed because 1879 should fix this.

cherrypy/test/test_bus.py
```diff
@@ -218,6 +220,7 @@ def f(method):  # pylint: disable=invalid-name
         assert bus.state in states, 'State %r not in %r' % (bus.state, states)
 
 
+@pytest.mark.skipif(CI_ON_MACOS, reason="continuous integration on macOS fails")
```
webknjaz: xfail plz

setup.py
```diff
@@ -85,12 +85,7 @@
 
             # cherrypy.lib.gctools
             'objgraph',
-
```
webknjaz, line 88: don't remove the delimiter

webknjaz (changes requested):

setup.py
webknjaz: Well, this is totally reasonable: CherryPy is pure-python and this is why it's expected for it to work under any Python runtime.

cherrypy/test/test_bus.py
```diff
@@ -218,6 +220,7 @@ def f(method):  # pylint: disable=invalid-name
         assert bus.state in states, 'State %r not in %r' % (bus.state, states)
 
 
+@pytest.mark.xfail(CI_ON_MACOS, reason="continuous integration on macOS fails")
```
webknjaz, line 224: Does this happen in any macOS env or only in Circle CI?
I've restarted macOS jobs @ https://travis-ci.org/github/cherrypy/cherrypy/builds/743540522 to figure out if it's also true for Travis CI.

cherrypy/test/test_bus.py
```diff
@@ -9,7 +11,7 @@
 
 from cherrypy.process import wspbus
 
-
+CI_ON_MACOS = bool(os.getenv("CI")) and sys.platform == "darwin"
```
webknjaz, line 15: Preserve the "two lines after imports" spacing
```suggestion

CI_ON_MACOS = bool(os.getenv("CI")) and sys.platform == "darwin"
```

cclauss: Can we please format the codebase with [psf/black](https://github.com/psf/black) and eliminate the arbitrary code formatting rules?

webknjaz: No, black is so terrible that I avoid projects pushing it down the throat.
Also, it's not arbitrary. Common sense tells us not to change the formatting unrelated to other changes in the PR so why do that at all? I'm not asking for anything that doesn't make sense: there's a visual separation that's already maintained there — just don't break it, I don't ask to reformat anything in the actual changes. Also, none of the existing formatters is clever enough to keep the visual separators meant for humans, not machines. If anything's not right beside what pre-commit reports, I'd rather fix it myself than annoy the contributors.
Besides, there's enough autoformatters in our pre-commit setup: https://travis-ci.org/github/cherrypy/cherrypy/jobs/743940552#L291.

cherrypy/test/test_bus.py
webknjaz: Confirmed. This is also happening on Travis CI: https://travis-ci.org/github/cherrypy/cherrypy/jobs/743540538#L17898.
I was worried that the condition may be not precise enough and would xfail unrelated envs but it seems it's okay.

PR #1892, opened 2020-11-28
Fix `test_block()` to ignore the final message with a thread name

cclauss:
  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

Continuous Integration tests fail.

Continuous Integration tests fail.

**Other Information**:

**Checklist**:

### DIFF
--- a/cherrypy/test/test_bus.py
+++ b/cherrypy/test/test_bus.py
@@ -269,7 +269,11 @@ def test_block(bus, log_tracker):
         'Waiting for child threads to terminate...',
     ]
     # If the last message mentions an indeterminable thread name then ignore it
-    assert expected in (log_tracker.log_entries, log_tracker.log_entries[:-1])
+    len_expected = len(expected)
+    assert log_tracker.log_entries[:len_expected] == expected
+    assert len(log_tracker.log_entries[len_expected:]) - len_expected in (0, 1), (
+        'No more than one extra log line with the thread name expected'
+    )
 
 
 def test_start_with_callback(bus):
--- a/cherrypy/test/test_bus.py
+++ b/cherrypy/test/test_bus.py
@@ -271,7 +271,7 @@ def test_block(bus, log_tracker):
     # If the last message mentions an indeterminable thread name then ignore it
     len_expected = len(expected)
     assert log_tracker.log_entries[:len_expected] == expected
-    assert len(log_tracker.log_entries) - len_expected in (0, 1), (
+    assert len(log_tracker.log_entries) - len_expected <= 1, (
         'No more than one extra log line with the thread name expected'
     )
 
--- a/cherrypy/test/test_bus.py
+++ b/cherrypy/test/test_bus.py
@@ -271,7 +271,7 @@ def test_block(bus, log_tracker):
     # If the last message mentions an indeterminable thread name then ignore it
     len_expected = len(expected)
     assert log_tracker.log_entries[:len_expected] == expected
-    assert len(log_tracker.log_entries[len_expected:]) - len_expected in (0, 1), (
+    assert len(log_tracker.log_entries) - len_expected in (0, 1), (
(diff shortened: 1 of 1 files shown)


cherrypy/test/test_bus.py
```diff
@@ -261,10 +261,15 @@ def g():  # pylint: disable=invalid-name
...
-             'Waiting for child threads to terminate...'])
+    expected = [
+        "Bus STOPPING",
+        "Bus STOPPED",
+        "Bus EXITING",
+        "Bus EXITED",
+        "Waiting for child threads to terminate...",
+    ]
+    # If the last message mentions an indeterminable thread name then ignore it
+    assert expected in (log_tracker.log_entries, log_tracker.log_entries[:-1])
```
webknjaz: I'm not sure this is reasonable. If there's a bug, it should be fixed. Making the test runner shut up just buries the bug and that is bad.

cclauss: I do not think that it is a bug.  This is about macOS behaving differently than other platforms.  The final message is not returned on macOS but is on other platforms.  The test already ignores the final message.

webknjaz: Interesting... I'm concerned because the related code is probably about a decade old and so if something changed in the env (like newer Pythons or OSs), the code either needs to be updated or it should be made clear in tests to expect one thing in some envs and another thing in different envs. It'd also need some code comment explaining the motivation to prevent regressions.

cclauss: \# If the last message mentions an indeterminable thread name then ignore it

webknjaz: Oh, I've probably misread the code and was thinking that you're cutting off https://github.com/cherrypy/cherrypy/blob/4912fca/cherrypy/process/wspbus.py#L352 for some reason.
How about
```suggestion
    expected_len = len(expected)
    assert expected == log_tracker.log_entries[:expected_len]
    assert len(log_tracker.log_entries[expected_len:]) <= 1, (
        'No more than one extra log line with the thread name expected'
    )
```

This way, the code should communicate the intent clearer.

cherrypy/test/test_bus.py
```diff
@@ -261,10 +261,19 @@ def g():  # pylint: disable=invalid-name
...
+        'Bus STOPPING',
+        'Bus STOPPED',
+        'Bus EXITING',
+        'Bus EXITED',
+        'Waiting for child threads to terminate...',
+    ]
+    # If the last message mentions an indeterminable thread name then ignore it
+    len_expected = len(expected)
+    assert log_tracker.log_entries[:len_expected] == expected
+    assert len(log_tracker.log_entries) - len_expected in (0, 1), (
```
webknjaz: Why did you obfuscate the check? Using `len(log_tracker.log_entries[len_expected:]) < 2` is way cleaner

cclauss: This box is checked on this PR.
> Allow edits by maintainers

That means that maintainers have the ability to modify the code in this PR to match cherrypy’s needs.  If there are cleaner ways to express these tests then please feel free to make appropriate mods.  Our goal is to work together have a strong test suite that passes more than it fails.  This is required to give users and contributors confidence in the health of cherrypy.

webknjaz: Sure, you are absolutely right. That's why when you do something that I didn't expect I first try to ask about the motivation. Instead of assuming that you've made a mistake, I assume that you may have some serious reason behind applying the change in a different way. And in such unobvious cases, I prefer to ask if that's intentional rather than forcefully override the submission.

webknjaz: It may not seem like it but I'm very grateful for your effort to fix the CIs. It's something that I'm struggling to find time for lately.

PR #1894, opened 2020-12-02
test_params.test_syntax() will fail on Python > 3.9

cclauss:
  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

**Other information**:

**Checklist**:

### DIFF
--- a/cherrypy/test/test_params.py
+++ b/cherrypy/test/test_params.py
@@ -43,8 +43,8 @@ class ParamsTest(helper.CPWebCase):
         self.assertStatus(500)
 
     def test_syntax(self):
-        if sys.version_info < (3,):
-            return self.skip('skipped (Python 3 only)')
+        if not (3,) < sys.version_info < (3, 10):
+            return self.skip('skipped (Python 3.0 to 3.9 only)')
         code = textwrap.dedent("""
             class Root:
                 @cherrypy.expose


jaraco (changes requested):
I'm okay with skipping tests for older Python versions, but skipping a test for a newer version is problemmatic. If the test is not expected to pass on future versions, then it's probably not a test worth having. I can't tell from this PR what the test is for or why it fails on Python 3.10 and later. It seems to me more analysis is needed.

webknjaz (changes requested):
I agree with Jason: this doesn't look right.

webknjaz:
> I can't tell from this PR what the test is for or why it fails on Python 3.10 and later. It seems to me more analysis is needed.

After looking closer, I assume that the failure was to be caused by
> ```python
> def resource(self, limit: int):
> ```

having an unquoted `int`. It could've been easily fixed by just changing it to use `'int'`. There was a plan to change how the annotations are handled in Python 3.10 but that change has been reverted last-minute and the decision to reintroduce it was postponed because it breaks pydantic. So now there's no need to fix anything.

FWIW the patch could've been
```diff
-def resource(self, limit: int):
+def resource(self, limit: 'int'):
```

Issue #1308, opened 2014-03-30
testsuite failures X ~5; test_file_stream and 4 more in test_logging.py under pypy


ghost:
Originally reported by: **idella5 (Bitbucket: [idella5](http://bitbucket.org/idella5), GitHub: [idella5](http://github.com/idella5))**

---

The first appears to fail for all python versions.

``` python
====================================================================
FAIL: test_file_stream (cherrypy.test.test_static.StaticTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/mnt/gen2/TmpDir/portage/dev-python/cherrypy-3.2.2-r1/work/CherryPy-3.2.2-python2_7/cherrypy/test/test_static.py", line 259, in test_file_stream
    expected)
AssertionError: The file should have advanced to position 65536, but has already advanced to the end of the file. It may not be streamed as intended, or at the wrong chunk size (64k)
-------------------- >> begin captured logging << --------------------
cherrypy.access.22126352: INFO: 127.0.0.1 - - [30/Mar/2014:17:38:40] "GET /bigfile HTTP/1.1" 200 1048576 "" ""
cherrypy.access.22126352: INFO: 127.0.0.1 - - [30/Mar/2014:17:38:40] "GET /tell HTTP/1.1" 200 - "" ""
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
Ran 219 tests in 62.558s

FAILED (SKIP=2, failures=1)
```

---

The second only for py3.2, however it's also intermittent.  

``` python
===================================================================
ERROR: test_HTTP11_pipelining (cherrypy.test.test_conn.PipelineTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/mnt/gen2/TmpDir/portage/dev-python/cherrypy-3.2.2-r1/work/CherryPy-3.2.2-python3_2/cherrypy/test/test_conn.py", line 428, in test_HTTP11_pipelining
    response.begin()
  File "/usr/lib64/python3.2/http/client.py", line 346, in begin
    version, status, reason = self._read_status()
  File "/usr/lib64/python3.2/http/client.py", line 328, in _read_status
    raise BadStatusLine(line)
nose.proxy.BadStatusLine: Hello, world!
-------------------- >> begin captured logging << --------------------
cherrypy.access.140312555645200: INFO: 127.0.0.1 - - [30/Mar/2014:18:04:35] "GET /hello HTTP/1.1" 200 13 "" ""
[...]

PR #1946, opened 2022-01-31
Use context managers to close files properly and fix tests on PyPy

mgorny:
  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [x] refactoring
  - [ ] other

#1308 (partially related, it's an old bug filed for PyPy2)

The code prior to this change used four approaches for closing files:

1. Using a context manager (`with` clause).
2. Using a try/finally clause.
3. Closing the file in the same scope (unreliable: file object can leak on exception).
4. Not closing open files at all.
    
The last point is a real problem for PyPy since it does not GC unreachable objects as aggressively as CPython does.  While leaving a function scope on CPython causes the file objects private to it to be destroyed (and therefore closed), in PyPy they can stay dangling
for some time.  When combines with buffered writes, this means that writes can still remain pending after returning from function.

Context managers are used consistently everywhere. This gives me 100% test pass with PyPy3.7.

**Other information**:

**Checklist**:

  - [x] Documentation reflects the changes (-- I don't think any doc changes are necessary)

### DIFF
--- a/cherrypy/lib/covercp.py
+++ b/cherrypy/lib/covercp.py
@@ -334,9 +334,10 @@ class CoverStats(object):
         yield '</body></html>'
 
     def annotated_file(self, filename, statements, excluded, missing):
-        source = open(filename, 'r')
+        with open(filename, 'r') as source:
+            lines = source.readlines()
         buffer = []
-        for lineno, line in enumerate(source.readlines()):
+        for lineno, line in enumerate(lines):
             lineno += 1
             line = line.strip('\n\r')
             empty_the_buffer = True
--- a/cherrypy/_cperror.py
+++ b/cherrypy/_cperror.py
@@ -532,7 +532,8 @@ def get_error_page(status, **kwargs):
                     return result
             else:
                 # Load the template from this path.
-                template = io.open(error_page, newline='').read()
+                with io.open(error_page, newline='') as f:
+                    template = f.read()
         except Exception:
             e = _format_exception(*_exc_info())[-1]
             m = kwargs['message']
--- a/cherrypy/_cpmodpy.py
+++ b/cherrypy/_cpmodpy.py
@@ -339,11 +339,8 @@ LoadModule python_module modules/mod_python.so
                                      }
 
         mpconf = os.path.join(os.path.dirname(__file__), 'cpmodpy.conf')
-        f = open(mpconf, 'wb')
-        try:
+        with open(mpconf, 'wb') as f:
             f.write(conf_data)
-        finally:
-            f.close()
 
         response = read_process(self.apache_path, '-k start -f %s' % mpconf)
         self.ready = True
--- a/cherrypy/lib/auth_digest.py
+++ b/cherrypy/lib/auth_digest.py
@@ -101,13 +101,12 @@ def get_ha1_file_htdigest(filename):
     """
     def get_ha1(realm, username):
         result = None
-        f = open(filename, 'r')
-        for line in f:
-            u, r, ha1 = line.rstrip().split(':')
-            if u == username and r == realm:
-                result = ha1
-                break
-        f.close()
+        with open(filename, 'r') as f:
+            for line in f:
+                u, r, ha1 = line.rstrip().split(':')
+                if u == username and r == realm:
+                    result = ha1
+                    break
         return result
 
     return get_ha1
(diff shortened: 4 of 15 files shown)


mgorny:
It seems that the AppVeyor failure is a flaky test.

graingert:
Can this warning ignore be removed now? https://github.com/cherrypy/cherrypy/blob/main/pytest.ini#L41

cherrypy/lib/covercp.py
```diff
@@ -334,26 +334,26 @@ def menu(self, base='/', pct='50', showpct='',
...
+                elif lineno in statements:
+                    template = TEMPLATE_LOC_COVERED
+                else:
+                    empty_the_buffer = False
+                    buffer.append((lineno, line))
+                if empty_the_buffer:
+                    for lno, pastline in buffer:
+                        yield template % (lno, cgi.escape(pastline))
+                    buffer = []
+                    yield template % (lineno, cgi.escape(line))
```
graingert: because there's a `yield` inside this `with ` you're relying on the GC to `.close()` this generator to exit the `with` block using a `GeneratorExit`

I think you can fix it by using:

```python
yield from self.annotated_file(filename, statements, excluded,
                                        missing):
```

in `def report(self, name):`

mgorny: Ok, I've changed that. Alternatively, I suppose we could just put the list from `readlines()` into a variable and have it closed before the iteration.

graingert: :facepalm: oh I only just noticed the `readlines()` call.

```python
with open(filename, 'r') as f:
    lines = f.readlines()
```

would be much more clear that the file is immediately consumed and closed.

however maybe that's a mistake and the file was intended to be streamed, and the code should be:

```python
for lineno, line in enumerate(source):
```

mgorny:
> Can this warning ignore be removed now? https://github.com/cherrypy/cherrypy/blob/main/pytest.ini#L41

I'm afraid not. There are more leaking files but the traceback is entirely unhelpful to me. I'm not even sure if they're bugs in cherrypy or just the test suite taking shortcuts somewhere.

cherrypy/lib/covercp.py
graingert: Originally the file was streamed in this generator and then it was changed to `.readlines()` https://github.com/cherrypy/cherrypy/commit/596b1f603047a407bf8949b018671c7de13e9f2f#diff-342872edf8db6cab0554b70351960b4850b8eb7120003bad9e2d074c69a7c758L335-R333

mgorny: Ok, I'll do the first snippet but I suppose there's no harm in leaving the `yield from` change as well.

webknjaz:
> I'm not even sure if they're bugs in cherrypy or just the test suite taking shortcuts somewhere.

I think this may be both. There were some corner cases I've been unable to debug. Ideally, we shouldn't have any ignores like that, of course. The only reason it is there is that some of them are hard to track down, especially, when they are flaky.

mgorny:
Just to be clear, I'm only @Gentoo packager for Python stuff and I don't even really know how to use CherryPy (nor wanna learn, sorry), let alone figure out how it deals with keeping resources like open files.

cherrypy/lib/covercp.py
```diff
@@ -364,9 +364,8 @@ def report(self, name):
                                        fullpath=name,
                                        pc=pc)
         yield '<table>\n'
-        for line in self.annotated_file(filename, statements, excluded,
```
webknjaz: Was this leaking file descriptors? I don't fully understand why it is in the scope of this PR (I'd appreciate if it was as atomic as possible).

mgorny: No, we thought it was originally necessary because of changes in `annotated_file()` but it's not now. I'll revert it then.

graingert: @webknjaz it was leaking an unclosed generator (`annotated_file()`), which then went on to leak a file descriptor - the `yield` has been moved outside of the `with open(filename) as source): ` block and so now only the generator is leaked

webknjaz:
@mgorny I know and I highly appreciate your contribution! Unfortunately, the current CI setup is also in a bad state so keep in mind that the AppVeyor tests may be failing for unrelated reasons.

P.S. Note that our release automation got broken over time so I won't be able to release it right away. So I suppose you may need to integrate this patch into the ebuild via https://github.com/cherrypy/cherrypy/pull/1946.patch.

webknjaz (approved):
Brilliant! Thanks for the patch! I'll wait a bit for the Circle CI status to be reported and will merge then.

mgorny:
Don't worry about the release. Right now my biggest blocker is still the horror called twisted.

Issue #1958, opened 2022-04-12
DeprecationWarning: This method will be removed in Python 3.12. Use 'parser.read_file()' instead.


radez:
**I'm submitting a ...**
- [ X ] bug report

**What is the current behavior?**

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cherrypy/_cpconfig.py:157: in update
    super(Config, self).update(config)
        __class__  = <class 'cherrypy._cpconfig.Config'>
        config     = <_io.StringIO object at 0x7fde106cb5b0>
        self       = {'tools.log_tracebacks.on': True, 'tools.log_headers.on': False, 'tools.trailing_slash.on': True, 'tools.encode.on': T...lse, 'checker.on': False, 'request.show_tracebacks': True, 'request.show_mismatched_params': True, 'log.screen': False}
cherrypy/lib/reprconf.py:130: in update
    self._apply(Parser.load(config))
        config     = <_io.StringIO object at 0x7fde106cb5b0>
        self       = {'tools.log_tracebacks.on': True, 'tools.log_headers.on': False, 'tools.trailing_slash.on': True, 'tools.encode.on': T...lse, 'checker.on': False, 'request.show_tracebacks': True, 'request.show_mismatched_params': True, 'log.screen': False}
cherrypy/lib/reprconf.py:208: in load
    return Parser().dict_from_file(input) if is_file else input.copy()
        input      = <_io.StringIO object at 0x7fde106cb5b0>
        is_file    = True
        self       = <class 'cherrypy.lib.reprconf.Parser'>
cherrypy/lib/reprconf.py:194: in dict_from_file
    self.readfp(file)
        file       = <_io.StringIO object at 0x7fde106cb5b0>
        self       = <cherrypy.lib.reprconf.Parser object at 0x7fde140a5360>
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <cherrypy.lib.reprconf.Parser object at 0x7fde140a5360>
fp = <_io.StringIO object at 0x7fde106cb5b0>, filename = None

    def readfp(self, fp, filename=None):
        """Deprecated, use read_file instead."""
>       warnings.warn(
            "This method will be removed in Python 3.12. "
[...]

PR #1959, opened 2022-04-12
Replace readfp with read_file: 3.12 compatibility requirement

radez:
Getting Deprecation Warning in python 3.11:
DeprecationWarning: This method will be removed in Python 3.12.
Use 'parser.read_file()' instead.

read_file was introduced in 3.2, so we should not incur cherrypy backwards compat issues.

https://github.com/python/cpython/commit/e8659b47dece5a272111c0af5e340c364a9f807b#diff-ca80d44aeeadd5e4966195b99aff106abb3e7b0b5356b7499125ee021bd84777R757

  - [ X ] bug fix

#1958 

#1958 

No more deprecation warning

**Other information**:
If I'm reading this correct read_file has been around since Python 2.7
https://github.com/python/cpython/commit/a492362f9a2a44e411147fd7b2886466bb0bb17f

**Checklist**:

  - [x] Documentation reflects the changes

### DIFF
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,0 +1,4 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%(describe:tags=true)$
+ref-names: $Format:%D$
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,6 @@
+# Needed for setuptools-scm-git-archive
+.git_archival.txt  export-subst
+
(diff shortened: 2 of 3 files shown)


mgorny:
Ping. This is needed to fix compatibility with Python 3.12.

webknjaz:
I clicked "rebase" and it disappeared...

webknjaz:
Oh, it looks I already made this change last January. That's why.

PR #2007, opened 2023-12-14
Replace deprecated `datetime.utcnow()` w/ a tz-aware variant

radez:
Calling `datetime.datetime.utcnow()` emits a deprecation warning now.
Python 3.12 reports:
```python-traceback
DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```
This patch replaces it with a non-deprecated cross-python implementation.

**What kind of change does this PR introduce?**
  - [x] refactoring

**Checklist**:

  - [x] I think the code is well written
  - [x] I wrote [good commit messages][1]
  - [x] I have [squashed related commits together][2] after the changes have been approved
  - [x] I used the same coding conventions as the rest of the project
  - [x] The new code doesn't generate linter offenses
  - [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences

[1]: http://chris.beams.io/posts/git-commit/
[2]: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

PR #2008, opened 2023-12-14
Upgrade the RTD config with the current schema

radez:
The respective changes in the service have been announced in the following blog post: https://blog.readthedocs.com/migrate-configuration-v2/.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,12 +1,17 @@
 ---
 
 build:
-  image: latest
+  os: ubuntu-22.04
+  tools:
+    python: >-
+      3.6
+
 python:
-  version: 3.6
-  extra_requirements:
-  - docs
-  - testing
-  pip_install: true
+  install:
+  - method: pip
+    path: .
+    extra_requirements:
+    - docs
+    - testing
 
 ...


radez:
Looks like tests have changed since I wrote this, closing it since it's probably not needed

radez:
Seems my assumption was a bit hasty. My PR that was rebased on this one passed CI check and the one that wasn't didn't pass.

cherrypy/test/test_logging.py
```diff
@@ -199,7 +199,8 @@ def test_custom_log_format(log_tracker, monkeypatch, server):
 
 
 def test_utc_in_timez(monkeypatch):
-    """Test that ``LazyRfc3339UtcTime`` is rendered as ``str`` using UTC timestamp."""
+    """Test that ``LazyRfc3339UtcTime``
+    is rendered as ``str`` using UTC timestamp."""
```
webknjaz: This seems broken PEP 257-wise. Any ideas of how to make it look better? The title line is usually a full sentence on a single line. Max 72 chars. Then, there's an empty line and a paragraph of longer text if needed.

.readthedocs.yml
```diff
@@ -1,12 +1,16 @@
 ---
 
 build:
-  image: latest
+  os: ubuntu-22.04
+  tools:
+    python: "3.6"
```
webknjaz: My personal preference is to avoid quotes whenever possible. So I tend to do this:
```suggestion
    python: >-
      3.6
```
(it's a nice practice that if used consistently, helps avoid accidental problems with having to escape stuff)

cherrypy/test/test_logging.py
radez: I think I shortened the sentence to fit 257. Just pushed a patch.

.readthedocs.yml
radez: I'm pretty sure I tried without quotes and RTD didn't accept it as a valid value. Would you like me to try again?

webknjaz:
> Seems my assumption was a bit hasty. My PR that was rebased on this one passed CI check and the one that wasn't didn't pass.

So the Sphinx build is happening in two places — the GHA CI and RTD. You fixed the RTD one an the CI is coming from tox, so that's probably where it'll have to be addressed. Maybe, the `rst.linker` dep or Sphinx versions need adjusting.

In general, I'd very much prefer having PRs focusing on one thing only, rather than trying to address several separate problems. Do you think you could split? Also, this would allow us to have better commit/PR names. "fixes for CI" isn't very accurate or informative.
I've got some link collections relating to this, if that'd help: https://gist.github.com/webknjaz/a5a4fb374b7579de827e6bedb93a5220 / https://gist.github.com/webknjaz/a7362787a80067af8621a85a71746ca1 / https://gist.github.com/webknjaz/cb7d7bf62c3dda4b1342d639d0e78d79 / https://gist.github.com/webknjaz/2aa9c7a43b51c1385260ff87e0cbb9e3.

By the way, you seem to be interested in getting the project back into shape. If you have any help, check out the Matrix/Gitter room @ https://matrix.to/#/#cherrypy:matrix.org — I'm personally more responsive on Matrix.

P.S. I'm willing to merge-in atomic PRs even when the overall CI is failing if it's visible that the changes turn corresponding individual status checks from red to green. This way, things can be fixed bit-by-bit, not all in one go.

radez:
I'm building cherrypy for Fedora so I'm looking at maintenance items whenever they arise.
Happy to split the patch into multiple items.I think the pep257 check will need more fixes than my one line. Other functions have multiline comments too.

I'll work on a few patches for fixing CI stuff and put them in over the next few days.

webknjaz:
> I think the pep257 check will need more fixes than my one line. Other functions have multiline comments too.

Do you mean multiline docstrings? I don't remember seeing multiline comment there as they have a different function.

.readthedocs.yml
webknjaz: Just apply this code suggestion using the GH UI button. It'll work. The `>-` syntax produces strings, just like inline quotes. But without it, the YAML parser would see a float value which is the problem you were seeing earlier. RTD disallows floats and fails, but with an explicit string, it's fine.

cherrypy/test/test_logging.py
webknjaz: It looks like you accidentally marked this as resolved — there's no new commit. OTOH, it's probably best to separate it out. In #2007 you also modify this line in a different way. It'd be easier to accept it as a separate patch that doesn't influence both of the unrelated PRs.

radez:
I've updated this patch to focus on the RTD yml file and made the version update to it.

webknjaz:
> I think the pep257 check will need more fixes than my one line.

Note that there's two such checks — one of them is allowed to fail precisely because of this...

webknjaz:
> I've updated this patch to focus on the RTD yml file and made the version update to it.

Alright.. I'll update the PR title and description to make sense in this context, then.

PR #2010, opened 2023-12-23
Use `os.listdir()` instead of `os.scandir()` in `FileSession` during clean-up

jiangwen365:
`os.scandir()` is more effcient than `os.listdir()` and is generator based.

  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [X] other

when cleaning file-based session storage, `os.listdir()` returns a huge list of file names for very large session directories. so switching to `os.scandir()` function – a better and faster directory iterator should be beneficial.

**Other information**:

**Checklist**:

### DIFF
--- a/cherrypy/lib/sessions.py
+++ b/cherrypy/lib/sessions.py
@@ -578,7 +578,8 @@ class FileSession(Session):
         """Clean up expired sessions."""
         now = self.now()
         # Iterate over all session files in self.storage_path
-        for fname in os.listdir(self.storage_path):
+        for entry in os.scandir(self.storage_path):
+            fname = entry.name
             have_session = (
                 fname.startswith(self.SESSION_PREFIX)
                 and not fname.endswith(self.LOCK_SUFFIX)
@@ -586,7 +587,7 @@ class FileSession(Session):
             if have_session:
                 # We have a session file: lock and load it and check
                 #   if it's expired. If it fails, nevermind.
-                path = os.path.join(self.storage_path, fname)
+                path = entry.path
                 self.acquire_lock(path)
                 if self.debug:
                     # This is a bit of a hack, since we're calling clean_up


cherrypy/lib/sessions.py
```diff
@@ -578,15 +578,16 @@ def clean_up(self):
         """Clean up expired sessions."""
         now = self.now()
         # Iterate over all session files in self.storage_path
-        for fname in os.listdir(self.storage_path):
+        for entry in os.scandir(self.storage_path):
```
webknjaz, line 581: If we're changing this, let's go for the higher level `pathlib` interface already rather then working with the low-level `os` API.

jiangwen365: Surprisingly, when I attempted later on to switch from using os.scandir() to pathlib's iterdir(), the performance significantly deteriorated. I am currently using a Windows server, and the calls to os.listdir() or path.iterdir() hardly ever return, whereas scandir() was able to return and start iterating within approximately 30 seconds. (I have about 500K session files) So I would suggest sticking to os.scandir() for better performance.

webknjaz: `pathlib` is the interface of choice for maintainability reasons. Also, using file-based sessions was never intended for high-performance use-cases specifically because file system scans are dependent on the number of files in directories and the lookups aren't exactly linear. Having a 30-second scan suggests that other file system operations during normal runtime aren't quite fast either.
So I'd say it's an architecture/deployment problem on the app side, not the framework. A typical solution is to use session storages that fit better for this purpose — like memcached or redis.

Could you post the timings/metrics showing the difference? And perhaps, report this issue on the upstream CPython tracker?

jaraco: I'm inclined to say we should accept this change as proposed as it provides a demonstrated performance benefit. The problem, as webknjaz has illustrated is that a subsequent developer might propose to change it from scandir to Path.iterdir, and there's nothing in the test suite that would prevent that. That is, if CherryPy wishes to support this use-case, and my instinct is that it should if possible (and jiangwen's experience shows it is possible), there should be a test to capture the expectation that this change sets. That is, let's not support the use-case incidentally, but intentionally, and add a test for it.

PR #1993, opened 2023-04-04
Replace the use of `pkg_resources` with `importlib.metadata`

radez:
This patch replaces references to `pkg_resources` with `importlib.metadata`
The latest setuptools emits the following warning: `DeprecationWarning: pkg_resources is deprecated as an API`.

**What kind of change does this PR introduce?**
  - [ ] bug fix
  - [ ] feature
  - [x] docs update
  - [x] tests/coverage improvement
  - [ ] refactoring
  - [x] other

**What is the related issue number (starting with `#`)**
Fixes #1973 

**Checklist**:

  - [x] I think the code is well written
  - [x] I wrote [good commit messages][1]
  - [x] I have [squashed related commits together][2] after the changes have been approved
  - [x] Unit tests for the changes exist
  - [x] Integration tests for the changes exist (if applicable)
  - [x] I used the same coding conventions as the rest of the project
  - [x] The new code doesn't generate linter offenses
  - [x] Documentation reflects the changes
  - [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences

[1]: http://chris.beams.io/posts/git-commit/
[2]: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

PR #2011, opened 2023-12-24
Handle importlib-metadata fallback in docs/conf.py

radez:
- also add install_requires for importlib-metadata

This is a followup pr to #1993

Missed handling fallback in py 3.6/3.7 and the install_requires for it.

### DIFF
--- a/cherrypy/__init__.py
+++ b/cherrypy/__init__.py
@@ -60,11 +60,8 @@ try:
     import importlib.metadata as importlib_metadata
 except ImportError:
     # fall back for python <= 3.7
-    # can simply pass when py 3.6/3.7 no longer supported
-    try:
-        import importlib_metadata
-    except ImportError:
-        pass
+    # This try/except can be removed with py <= 3.7 support
+    import importlib_metadata
 
 from threading import local as _local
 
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -16,6 +16,16 @@
...
 
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    # fall back for python <= 3.7
+    # This try/catch can be removed with py <= 3.7 support
+    try:
+        import importlib_metadata
+    except ImportError:
+        pass
```
webknjaz: That inner try-except isn't required because it's always installed as a dependency when needed.
```suggestion
    import importlib_metadata
```

cherrypy/__init__.py
```diff
@@ -60,7 +60,7 @@
     import importlib.metadata as importlib_metadata
 except ImportError:
     # fall back for python <= 3.7
-    # can simply pass when py 3.6/3.7 no longer supported
+    # This try/catch can be removed with py <= 3.7 support
     try:
         import importlib_metadata
     except ImportError:
```
webknjaz: This one should also be unwrapped.

radez: done

docs/conf.py
radez: done

webknjaz (approved):
Great, thanks!

PR #2012, opened 2023-12-24
Clean the session files upon unpackling Error in FileSession clean up

jiangwen365:
I've seen clean failures in my production error log files due to `UnpicklingError: invalid load key, '\x00'`. I suspect the session file was created by attackers.  This would 1) break the cleansing thread, leaving many expired sessions not cleaned up, also 2) I tend to think it's not good/safe to ignore and keep these "bad" session files. Below is the original traceback info in the error log.
```python-traceback
  File "C:\Program Files (x86)\Python311-32\Lib\site-packages\cherrypy\process\plugins.py", line 518, in run
    self.function(*self.args, **self.kwargs)
  File "C:\Program Files (x86)\Python311-32\Lib\site-packages\cherrypy\lib\sessions.py", line 586, in clean_up
    contents = self._load(path)
               ^^^^^^^^^^^^^^^^
  File "C:\Program Files (x86)\Python311-32\Lib\site-packages\cherrypy\lib\sessions.py", line 520, in _load
    return pickle.load(f)
           ^^^^^^^^^^^^^^
_pickle.UnpicklingError: invalid load key, '\x00'.
```
 

  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [X] other

`UnpicklingError: invalid load key, '\x00'` would break the cleansing thread, leaving many expired sessions not cleaned up, also I tend to think it's not good/safe to ignore and keep these "bad" session files.

If pickling a session file is throwing pickling error, then just delete that file.

**Other information**:

**Checklist**:

### DIFF
--- a/cherrypy/lib/sessions.py
+++ b/cherrypy/lib/sessions.py
@@ -603,6 +603,10 @@ class FileSession(Session):
                         if expiration_time < now:
                             # Session expired: deleting it
                             os.unlink(path)
+                
+                except pickle.UnpicklingError:
+                    os.unlink(path)
+                    
                 finally:
                     self.release_lock(path)
 
--- a/cherrypy/lib/sessions.py
+++ b/cherrypy/lib/sessions.py
@@ -603,10 +603,10 @@ class FileSession(Session):
                         if expiration_time < now:
                             # Session expired: deleting it
                             os.unlink(path)
-                
+
                 except pickle.UnpicklingError:
                     os.unlink(path)
-                    
+
                 finally:
                     self.release_lock(path)
 


cherrypy/lib/sessions.py
```diff
@@ -603,6 +603,10 @@ def clean_up(self):
                         if expiration_time < now:
                             # Session expired: deleting it
                             os.unlink(path)
+
+                except pickle.UnpicklingError:
```
webknjaz, line 607: Could you add a test for this?

jiangwen365: I wish I am able to write a test for this - but with limited skill, I couldn't figure out how to write one - if possible could you please provide some guidance and thoughts?

webknjaz: I'm not sure either. It would probably be an integration test that sets up a server with sessions, makes a query (which results in a session file appearing on disk), then corrupts that file by writing arbitrary data into it (a random byte might be enough), followed by tearing down the server and that should raise the exception. Maybe, it'd be easier (or more stable) with calling `clean_up()` and checking that the file disappeared from disk. Additionally, it would be wise to use `pytest-mocker`'s `mocker.spy()` on the `_load()` method to verify that the exception was raised, indeed.

Issue #2014, opened 2024-01-05
Python 3.13: ModuleNotFoundError: No module named 'cgi'


hroncok:
**I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

**What is the current behavior?**

With Python 3.13.0a2, importing cherrypy fails:

```python-traceback
...
    import cherrypy
/usr/lib/python3.13/site-packages/cherrypy/__init__.py:66: in <module>
    from ._cperror import (
/usr/lib/python3.13/site-packages/cherrypy/_cperror.py:135: in <module>
    from cherrypy.lib import httputil as _httputil
/usr/lib/python3.13/site-packages/cherrypy/lib/httputil.py:15: in <module>
    from cgi import parse_header
E   ModuleNotFoundError: No module named 'cgi'
```

This is reproducible by:

```console
$ python3.13 -m venv venv3.13
$ venv3.13/bin/pip install cherrypy
...
$ venv3.13/bin/python
>>> import cherrypy
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    import cherrypy
  File ".../venv3.13/lib64/python3.13/site-packages/cherrypy/__init__.py", line 66, in <module>
    from ._cperror import (
  File ".../venv3.13/lib64/python3.13/site-packages/cherrypy/_cperror.py", line 135, in <module>
    from cherrypy.lib import httputil as _httputil
  File ".../venv3.13/lib64/python3.13/site-packages/cherrypy/lib/httputil.py", line 15, in <module>
    from cgi import parse_header
ModuleNotFoundError: No module named 'cgi'

```

I see the code in the main branch:

https://github.com/cherrypy/cherrypy/blob/30d9a141c67ecb958ce305810203d89d60ef318b/cherrypy/lib/httputil.py#L15

**What is the expected behavior?**

The CGI module was deprecated and removed from Python. Expected behavior is that it is not used.

**What is the motivation / use case for changing the behavior?**

We are testing all Fedora packages with early pre-releases of Python 3.13 to ensure everything is working once the final 3.13 is out.

**Please tell us about your environment:**

- Cheroot version: ?
- CherryPy version: main
- Python version: 3.13.0a2
- OS: Fedora Linux 39, 40
- Browser: N/A

webknjaz:
Thanks for the report, Miro!

[...]

PR #2021, opened 2024-02-19
🔧 Stop using the `cgi` stdlib module

radez:
the module was deprecatied in py 3.11 and removed in py 3.13 there are examples of using the email module to resolve this: PEP 594
https://github.com/python-babel/babel/issues/873
https://stackoverflow.com/questions/69068527/python-3-cgi-parse-header

This method doesn't seem to work for cherrypy.
The email.message module is trying to create a higher level object that is intelligent about the headers it has defined. The cgi.parse_header function is a very low level unintelligent function that simply parses header content based on structure and does not inspect the contents. Because of the intelligence of the email.message object cherrypy can't use it in the very generic way that the cgi.parse_header function was being used.

Fix #2014

**What kind of change does this PR introduce?**
  - [x] refactoring
  - [x] other - debt

**What is the related issue number (starting with `#`)**
#2014

**What is the current behavior?** (You can also link to an open issue here)
Deprecation warnings in 3.11 & and 3.12
Module import error in 3.13
#2014 

**What is the new behavior (if this is a feature change)?**

**Other information**:

**Checklist**:

  - [x] I think the code is well written
  - [x] I wrote [good commit messages][1]
  - [x] I have [squashed related commits together][2] after the changes have been approved
  - [x] Unit tests for the changes exist
  - [x] Integration tests for the changes exist (if applicable)
  - [x] I used the same coding conventions as the rest of the project
  - [x] The new code doesn't generate linter offenses
  - [x] Documentation reflects the changes
  - [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences

[1]: http://chris.beams.io/posts/git-commit/
[2]: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

PR #2027, opened 2024-04-02
Add test coverage for `HeaderElement` and `AcceptElement`

radez:
This is in preparation for issue #2014
Establishing tests to validate these objects so we can pull in the deprecated cgi.parse_header code

### DIFF
--- a/cherrypy/test/test_httputil.py
+++ b/cherrypy/test/test_httputil.py
@@ -85,8 +85,12 @@ def test_invalid_status(status_code, error_msg):
 @pytest.mark.parametrize(
     ('header_content', 'value', 'params'),
     (
-        ('application/x-www-form-urlencoded',
-         'application/x-www-form-urlencoded', {}),  # default_content_type
+        pytest.param(
+            'application/x-www-form-urlencoded',
+            'application/x-www-form-urlencoded',
+            {},
+            id='default-content-type',
+        ),
         ('application/json;charset="utf8"',
          'application/json', {'charset': 'utf8'}),
         ('audio/*; q=0.2, audio/basic',
@@ -121,8 +125,12 @@ def test_header_element(header_content, value, params):
 @pytest.mark.parametrize(
     ('header_content', 'media_type', 'qvalue'),
     (
-        ('application/x-www-form-urlencoded',
-         'application/x-www-form-urlencoded', 1.0),  # default_content_type
+        pytest.param(
+            'application/x-www-form-urlencoded',
+            'application/x-www-form-urlencoded',
+            1.0,
+            id='default-content-type',
+        ),
         ('application/json;charset="utf8"',
          'application/json', 1.0),
         ('text/*, text/html, text/html;level=1, */*',
--- a/cherrypy/test/test_httputil.py
+++ b/cherrypy/test/test_httputil.py
@@ -84,7 +84,7 @@ def test_invalid_status(status_code, error_msg):
 
 @pytest.mark.parametrize(
     ('header_content', 'value', 'params'),
-    [
+    (
         ('application/x-www-form-urlencoded',
          'application/x-www-form-urlencoded', {}),  # default_content_type
         ('application/json;charset="utf8"',
@@ -105,7 +105,7 @@ def test_invalid_status(status_code, error_msg):
          'text/plain', {}),
         ('text/xml',
          'text/xml', {}),
-    ]
+    ),
 )
 def test_header_element(header_content, value, params):
     """Test that ``value`` and ``params`` are extracted from headers.
@@ -120,7 +120,7 @@ def test_header_element(header_content, value, params):
 
 @pytest.mark.parametrize(
     ('header_content', 'media_type', 'qvalue'),
-    [
+    (
         ('application/x-www-form-urlencoded',
          'application/x-www-form-urlencoded', 1.0),  # default_content_type
         ('application/json;charset="utf8"',
@@ -133,7 +133,7 @@ def test_header_element(header_content, value, params):
          'text/plain', 1.0),
         ('text/xml',
          'text/xml', 1.0),
-    ]
+    ),
 )
 def test_accept_element(header_content, media_type, qvalue):
    """Test that ``value`` and ``qvalue`` are extracted from headers.
@@ -147,7 +147,7 @@ def test_accept_element(header_content, media_type, qvalue):
 
 @pytest.mark.parametrize(
     ('header_content', 'media_type', 'qvalue'),
-    [
+    (
         ('audio/*; q=0.2, audio/basic',
          'audio/*', {'q': '0.2, audio/basic'}),
         ('text/plain; q=0.5, text/html, text/x-dvi; q=0.8, text/x-c',
@@ -156,7 +156,7 @@ def test_accept_element(header_content, media_type, qvalue):
          'gzip', {'q': '0'}),
         ('da, en-gb;q=0.8, en;q=0.7',
          'da, en-gb', 0.8),
-    ]
+    ),
 )
 def test_accept_element_raises_400(header_content, media_type, qvalue):
     """Check bad headers crash :class:`cherrypy.httputil.AcceptElement`.
--- a/cherrypy/test/test_httputil.py
+++ b/cherrypy/test/test_httputil.py
@@ -116,8 +116,7 @@ def test_header_element(header_content, value, params):
     """
     hdr_elem = httputil.HeaderElement.from_str(header_content)
 
-    assert hdr_elem.value == value
-    assert hdr_elem.params == params
+    assert (hdr_elem.value, hdr_elem.params) == (value, params)
 
 
 @pytest.mark.parametrize(
@@ -144,8 +143,7 @@ def test_accept_element(header_content, media_type, qvalue):
    :py:class:`~cherrypy.httputil.AcceptElement` class.
    """
     acc_elem = httputil.AcceptElement.from_str(header_content)
-    assert acc_elem.value == media_type
-    assert acc_elem.qvalue == qvalue
+    assert (acc_elem.value, acc_elem.qvalue) == (media_type, qvalue)
 
 
 @pytest.mark.parametrize(
--- a/cherrypy/test/test_httputil.py
+++ b/cherrypy/test/test_httputil.py
@@ -108,10 +108,12 @@ def test_invalid_status(status_code, error_msg):
     ]
 )
 def test_header_element(header_content, value, params):
-    """ Test that value and params are parsed from headers.
+    """Test that ``value`` and ``params`` are extracted from headers.
 
-        Testing that the value and params are parsed from
-        headers passed to HeaderElement from httputils """
+    This is a positive test case, checking that the value and
+    params are parsed from headers that are being passed into
+    the :py:class:`~cherrypy.httputil.HeaderElement` class.
+    """
     hdr_elem = httputil.HeaderElement.from_str(header_content)
 
     assert hdr_elem.value == value
@@ -136,8 +138,11 @@ def test_header_element(header_content, value, params):
     ]
 )
 def test_accept_element(header_content, media_type, qvalue):
-    """ Test that value and qvalue are parsed from headers passed to
-        AcceptElement from httputils """
+   """Test that ``value`` and ``qvalue`` are extracted from headers.
+
+   This is being checked in the context of the
+   :py:class:`~cherrypy.httputil.AcceptElement` class.
+   """
     acc_elem = httputil.AcceptElement.from_str(header_content)
     assert acc_elem.value == media_type
     assert acc_elem.qvalue == qvalue
@@ -157,7 +162,9 @@ def test_accept_element(header_content, media_type, qvalue):
     ]
 )
 def test_accept_element_raises_400(header_content, media_type, qvalue):
-    """ Test that headers passed to AcceptElement from httputils
-        raise an expected HTTPError """
+    """Check bad headers crash :class:`cherrypy.httputil.AcceptElement`.
+
+    The expected exception is :exc:`~cherrypy.HTTPError`.
+    """
     httputil.AcceptElement.from_str(header_content)
     assert pytest.raises(cherrypy.HTTPError)
--- a/cherrypy/test/test_httputil.py
+++ b/cherrypy/test/test_httputil.py
@@ -120,7 +120,7 @@ def test_header_element(header_content, value, params):
 
 
 @pytest.mark.parametrize(
-    'header_content,media_type,qvalue',
+    ('header_content', 'media_type', 'qvalue'),
     [
         ('application/x-www-form-urlencoded',
          'application/x-www-form-urlencoded', 1.0),  # default_content_type
@@ -147,7 +147,7 @@ def test_accept_element(header_content, media_type, qvalue):
 
 
 @pytest.mark.parametrize(
-    'header_content,media_type,qvalue',
+    ('header_content', 'media_type', 'qvalue'),
     [
         ('audio/*; q=0.2, audio/basic',
          'audio/*', {'q': '0.2, audio/basic'}),
--- a/cherrypy/test/test_httputil.py
+++ b/cherrypy/test/test_httputil.py
@@ -144,11 +144,11 @@ def test_header_element(header_content, value, params):
     ),
 )
 def test_accept_element(header_content, media_type, qvalue):
-   """Test that ``value`` and ``qvalue`` are extracted from headers.
+    """Test that ``value`` and ``qvalue`` are extracted from headers.
 
-   This is being checked in the context of the
-   :py:class:`~cherrypy.httputil.AcceptElement` class.
-   """
+    This is being checked in the context of the
+    :py:class:`~cherrypy.httputil.AcceptElement` class.
+    """
     acc_elem = httputil.AcceptElement.from_str(header_content)
     assert (acc_elem.value, acc_elem.qvalue) == (media_type, qvalue)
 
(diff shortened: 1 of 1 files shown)


cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,82 @@ def test_invalid_status(status_code, error_msg):
     """Check that invalid status cause certain errors."""
     with pytest.raises(ValueError, match=error_msg):
         httputil.valid_status(status_code)
+
+
+@pytest.mark.parametrize(
+    'header_content,value,params',
```
webknjaz: I didn't add the corresponding linter here but a better style is having an iterable of strings ­— it reads better because it's sparse.
```suggestion
    ('header_content', 'value', 'params'),
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,82 @@ def test_invalid_status(status_code, error_msg):
...
+        ('da, en-gb;q=0.8, en;q=0.7',
+         'da, en-gb', {'q': '0.7'}),
+        ('text/plain',
+         'text/plain', {}),
+        ('text/xml',
+         'text/xml', {}),
+    ]
+)
+def test_header_element(header_content, value, params):
+    """ Test that value and params are parsed from headers passed to
```
webknjaz: FYI I'm pretty sure this will cause the linter you're fixing in #2028 to fail.

cherrypy/test/test_httputil.py
radez: Fixed, I'll force push the update.

cherrypy/test/test_httputil.py
radez: heh, yup it will. Fixed and included in the coming patch

radez:
@webknjaz I pushed up some updates to address comments last week. Do I need to do anything else for this to go in?

webknjaz:
Thanks for reminding! I'm going to take another look at this now as I'm on the PyCon US Sprints.

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,84 @@ def test_invalid_status(status_code, error_msg):
...
+         'gzip', {'q': '0'}),
+        ('da, en-gb;q=0.8, en;q=0.7',
+         'da, en-gb', 0.8),
+    ]
+)
+def test_accept_element_raises_400(header_content, media_type, qvalue):
+    """ Test that headers passed to AcceptElement from httputils
+        raise an expected HTTPError """
+    httputil.AcceptElement.from_str(header_content)
+    assert pytest.raises(cherrypy.HTTPError)
```
webknjaz: This is an incorrect way of using `pytest.raises()`. There are two main problems with it:
1) It's always best to use the `match=` argument to state what the expectation is, otherwise, a too broad set of exception might make it pass.
2) This helper returns a context manager, and so it's meant to be used in the `with pytest.raises(Exc, match=r'^stuff$'):` block.

I suppose, it should look something like this:
```suggestion
    with pytest.raises(
            cherrypy.HTTPError,
            match=r'^Malformed HTTP header: `[^`]+`$',
    ):
        httputil.AcceptElement.from_str(header_content)
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,84 @@ def test_invalid_status(status_code, error_msg):
...
+        ('gzip;q=1.0, identity; q=0.5, *;q=0',
+         'gzip', {'q': '0'}),
+        ('da, en-gb;q=0.8, en;q=0.7',
+         'da, en-gb', 0.8),
+    ]
+)
+def test_accept_element_raises_400(header_content, media_type, qvalue):
+    """ Test that headers passed to AcceptElement from httputils
+        raise an expected HTTPError """
+    httputil.AcceptElement.from_str(header_content)
```
webknjaz: Given the other comment I made, I'm surprised that this doesn't fail in the CI — it seems like you expect that it would.

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,84 @@ def test_invalid_status(status_code, error_msg):
...
+def test_accept_element(header_content, media_type, qvalue):
+    """ Test that value and qvalue are parsed from headers passed to
+        AcceptElement from httputils """
+    acc_elem = httputil.AcceptElement.from_str(header_content)
+    assert acc_elem.value == media_type
+    assert acc_elem.qvalue == qvalue
+
+
+@pytest.mark.parametrize(
+    'header_content,media_type,qvalue',
```
webknjaz: I thought we had a linter for this... Let's enumerate the params list as an iterable of strings:
```suggestion
    ('header_content', 'media_type', 'qvalue'),
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,84 @@ def test_invalid_status(status_code, error_msg):
...
+        Testing that the value and params are parsed from
+        headers passed to HeaderElement from httputils """
+    hdr_elem = httputil.HeaderElement.from_str(header_content)
+
+    assert hdr_elem.value == value
+    assert hdr_elem.params == params
+
+
+@pytest.mark.parametrize(
+    'header_content,media_type,qvalue',
```
webknjaz:
```suggestion
    ('header_content', 'media_type', 'qvalue'),
```
*(same)*

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,84 @@ def test_invalid_status(status_code, error_msg):
...
+)
+def test_header_element(header_content, value, params):
+    """ Test that value and params are parsed from headers.
+
+        Testing that the value and params are parsed from
+        headers passed to HeaderElement from httputils """
+    hdr_elem = httputil.HeaderElement.from_str(header_content)
+
+    assert hdr_elem.value == value
+    assert hdr_elem.params == params
```
webknjaz: *(nit, nonblocking)* one assertion per test is usually preferred:
```suggestion
    assert (hdr_elem.value, hdr_elem.params) == (value, params)
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,84 @@ def test_invalid_status(status_code, error_msg):
     """Check that invalid status cause certain errors."""
     with pytest.raises(ValueError, match=error_msg):
         httputil.valid_status(status_code)
+
+
+@pytest.mark.parametrize(
+    ('header_content', 'value', 'params'),
+    [
+        ('application/x-www-form-urlencoded',
+         'application/x-www-form-urlencoded', {}),  # default_content_type
```
webknjaz: Here's an idea. The params get converted into long strings that are hard to follow, in the pytest output. To combat this, a typical solution would be to wrap each param via `pytest.param(..., id='...')` with that `id` made a human-readable kebab-case identifier for each case that would hint the reader what this parameter's high-level meaning is.
I got an example for you:
```suggestion
        pytest.param(
            'application/x-www-form-urlencoded',
            'application/x-www-form-urlencoded',
            {},
            id='default-content-type',
        ),
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,84 @@ def test_invalid_status(status_code, error_msg):
...
+         'text/plain', {}),
+        ('text/xml',
+         'text/xml', {}),
+    ]
+)
+def test_header_element(header_content, value, params):
+    """ Test that value and params are parsed from headers.
+
+        Testing that the value and params are parsed from
+        headers passed to HeaderElement from httputils """
```
webknjaz: Let's have this make more sense in the PEP 257 context:
```suggestion
    """Test that ``value`` and ``params`` are extracted from headers.

    This is a positive test case, checking that the value and
    params are parsed from headers that are being passed into
    the :py:class:`~cherrypy.httputil.HeaderElement` class.
    """
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,84 @@ def test_invalid_status(status_code, error_msg):
...
+         'iso-8859-5, unicode-1-1', 0.8),
+        ('text/plain',
+         'text/plain', 1.0),
+        ('text/xml',
+         'text/xml', 1.0),
+    ]
+)
+def test_accept_element(header_content, media_type, qvalue):
+    """ Test that value and qvalue are parsed from headers passed to
+        AcceptElement from httputils """
```
webknjaz:
```suggestion
    """Test that ``value`` and ``qvalue`` are extracted from headers.

    This is being checked in the context of the
    :py:class:`~cherrypy.httputil.AcceptElement` class.
    """
```

webknjaz (commented):
So.. I identified a few problems here. I'll try applying some patches and rebasing myself and will see where to go from there. Stay tuned!

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,84 @@ def test_invalid_status(status_code, error_msg):
...
+         'text/plain', {'q': '0.8, text/x-c'}),
+        ('gzip;q=1.0, identity; q=0.5, *;q=0',
+         'gzip', {'q': '0'}),
+        ('da, en-gb;q=0.8, en;q=0.7',
+         'da, en-gb', 0.8),
+    ]
+)
+def test_accept_element_raises_400(header_content, media_type, qvalue):
+    """ Test that headers passed to AcceptElement from httputils
+        raise an expected HTTPError """
```
webknjaz:
```suggestion
    """Check bad headers crash :class:`cherrypy.httputil.AcceptElement`.

    The expected exception is :exc:`~cherrypy.HTTPError`.
    """
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,91 @@ def test_invalid_status(status_code, error_msg):
...
+)
+def test_accept_element(header_content, media_type, qvalue):
+   """Test that ``value`` and ``qvalue`` are extracted from headers.
+
+   This is being checked in the context of the
+   :py:class:`~cherrypy.httputil.AcceptElement` class.
+   """
+    acc_elem = httputil.AcceptElement.from_str(header_content)
+    assert acc_elem.value == media_type
+    assert acc_elem.qvalue == qvalue
```
webknjaz:
```suggestion
    assert (acc_elem.value, acc_elem.qvalue) == (media_type, qvalue)
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,89 @@ def test_invalid_status(status_code, error_msg):
...
+)
+def test_header_element(header_content, value, params):
+    """Test that ``value`` and ``params`` are extracted from headers.
+
+    This is a positive test case, checking that the value and
+    params are parsed from headers that are being passed into
+    the :py:class:`~cherrypy.httputil.HeaderElement` class.
+    """
+    hdr_elem = httputil.HeaderElement.from_str(header_content)
+
```
webknjaz:
```suggestion
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,88 @@ def test_invalid_status(status_code, error_msg):
     """Check that invalid status cause certain errors."""
     with pytest.raises(ValueError, match=error_msg):
         httputil.valid_status(status_code)
+
+
+@pytest.mark.parametrize(
+    ('header_content', 'value', 'params'),
+    [
```
webknjaz:
```suggestion
    (
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,88 @@ def test_invalid_status(status_code, error_msg):
...
+         'iso-8859-5, unicode-1-1', {'q': '0.8'}),
+        ('gzip;q=1.0, identity; q=0.5, *;q=0',
+         'gzip', {'q': '0'}),
+        ('da, en-gb;q=0.8, en;q=0.7',
+         'da, en-gb', {'q': '0.7'}),
+        ('text/plain',
+         'text/plain', {}),
+        ('text/xml',
+         'text/xml', {}),
+    ]
```
webknjaz:
```suggestion
    ),
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,88 @@ def test_invalid_status(status_code, error_msg):
...
+    params are parsed from headers that are being passed into
+    the :py:class:`~cherrypy.httputil.HeaderElement` class.
+    """
+    hdr_elem = httputil.HeaderElement.from_str(header_content)
+    assert (hdr_elem.value, hdr_elem.params) == (value, params)
+
+
+@pytest.mark.parametrize(
+    ('header_content', 'media_type', 'qvalue'),
+    [
```
webknjaz:
```suggestion
    (
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,88 @@ def test_invalid_status(status_code, error_msg):
...
+         'application/json', 1.0),
+        ('text/*, text/html, text/html;level=1, */*',
+         'text/*, text/html, text/html', 1.0),
+        ('iso-8859-5, unicode-1-1;q=0.8',
+         'iso-8859-5, unicode-1-1', 0.8),
+        ('text/plain',
+         'text/plain', 1.0),
+        ('text/xml',
+         'text/xml', 1.0),
+    ]
```
webknjaz:
```suggestion
    ),
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,88 @@ def test_invalid_status(status_code, error_msg):
...
+   This is being checked in the context of the
+   :py:class:`~cherrypy.httputil.AcceptElement` class.
+   """
+    acc_elem = httputil.AcceptElement.from_str(header_content)
+    assert (acc_elem.value, acc_elem.qvalue) == (media_type, qvalue)
+
+
+@pytest.mark.parametrize(
+    ('header_content', 'media_type', 'qvalue'),
+    [
```
webknjaz:
```suggestion
    (
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,88 @@ def test_invalid_status(status_code, error_msg):
...
+        ('audio/*; q=0.2, audio/basic',
+         'audio/*', {'q': '0.2, audio/basic'}),
+        ('text/plain; q=0.5, text/html, text/x-dvi; q=0.8, text/x-c',
+         'text/plain', {'q': '0.8, text/x-c'}),
+        ('gzip;q=1.0, identity; q=0.5, *;q=0',
+         'gzip', {'q': '0'}),
+        ('da, en-gb;q=0.8, en;q=0.7',
+         'da, en-gb', 0.8),
+    ]
```
webknjaz:
```suggestion
    ),
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,88 @@ def test_invalid_status(status_code, error_msg):
...
+    """
+    hdr_elem = httputil.HeaderElement.from_str(header_content)
+    assert (hdr_elem.value, hdr_elem.params) == (value, params)
+
+
+@pytest.mark.parametrize(
+    ('header_content', 'media_type', 'qvalue'),
+    (
+        ('application/x-www-form-urlencoded',
+         'application/x-www-form-urlencoded', 1.0),  # default_content_type
```
webknjaz:
```suggestion
        pytest.param(
            'application/x-www-form-urlencoded',
            'application/x-www-form-urlencoded',
            1.0,
            id='default-content-type',
        ),
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,99 @@ def test_invalid_status(status_code, error_msg):
...
+        ('text/xml',
+         'text/xml', 1.0),
+    ),
+)
+def test_accept_element(header_content, media_type, qvalue):
+   """Test that ``value`` and ``qvalue`` are extracted from headers.
+
+   This is being checked in the context of the
+   :py:class:`~cherrypy.httputil.AcceptElement` class.
+   """
```
webknjaz:
```suggestion
    """Test that ``value`` and ``qvalue`` are extracted from headers.

    This is being checked in the context of the
    :py:class:`~cherrypy.httputil.AcceptElement` class.
    """
```

cherrypy/test/test_httputil.py
```diff
@@ -79,3 +80,99 @@ def test_invalid_status(status_code, error_msg):
...
+        ('da, en-gb;q=0.8, en;q=0.7',
+         'da, en-gb', 0.8),
+    ),
+)
+def test_accept_element_raises_400(header_content, media_type, qvalue):
+    """Check bad headers crash :class:`cherrypy.httputil.AcceptElement`.
+
+    The expected exception is :exc:`~cherrypy.HTTPError`.
+    """
+    with pytest.raises(
```
webknjaz, line 174: @radez so as I implemented this, the problem with the expectation in the test got evident. Could you see if you can identify what inputs would actually lead to the validation error expected here...

jaraco:
As mentioned in #2021, I'm re-targeting this PR for the `debt/pre-cgi-drop` branch while we work out the details so we can unblock that PR. The idea is - get tests passing against `debt/pre-cgi-drop` and `main` to uncover any regression introduced by dropping the cgi module.

webknjaz (changes requested):
@radez so I rebased this on top of `main` and 4 tests don't pass — https://github.com/cherrypy/cherrypy/actions/runs/9754497486/job/26921518822?pr=2027#step:25:161, just as I predicted (https://github.com/cherrypy/cherrypy/pull/2027/files#r1608887187):
* `cherrypy/test/test_httputil.py::test_accept_element_raises_400[audio/*; q=0.2, audio/basic-audio/*-qvalue0]`
* `cherrypy/test/test_httputil.py::test_accept_element_raises_400[text/plain; q=0.5, text/html, text/x-dvi; q=0.8, text/x-c-text/plain-qvalue1]`
* `cherrypy/test/test_httputil.py::test_accept_element_raises_400[gzip;q=1.0, identity; q=0.5, *;q=0-gzip-qvalue2]`
* `cherrypy/test/test_httputil.py::test_accept_element_raises_400[da, en-gb;q=0.8, en;q=0.7-da, en-gb-0.8]`

Could you look into fixing them?

Issue #1814, opened 2019-10-04
Utilising Cython with CherryPy


aytey:
**I'm submitting a ...**
- [X] bug report
- [X] feature request
- [ ] question about the decisions made in the repository

**Do you want to request a *feature* or report a *bug*?**

This is probably a feature request, but it could also be classed a bug in the current implementation.

**What is the current behavior?**

If you use Cython to convert certain Python code into executable machine code (e.g., a `.so` on Linux or a `.pyd` on Windows), then CherryPy's ability to detect that the method has been `expose`'d does not seem to work.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**

Some example code would look like this:

```python
import cherrypy

class Router(object):
    config = {}

    @cherrypy.expose
    def index(self):
        return "hello world"

def start_server():
    cherrypy.config.update({
        'server.socket_host': '127.0.0.1',
        'server.socket_port': 8127
    })

    cherrypy.tree.mount(Router(), '/route', Router.config)
    cherrypy.engine.start()
    cherrypy.engine.block()

# Run, then navigate to 127.0.0.1:8127/route in a browser
def main():
    start_server()
```

You can then write a simple "wrapper" just to call into the above:

```python
from test import main

if __name__ == "__main__":
    main()

# EOF
```

If this code is run as "normal" Python and you request `127.0.0.1:8127/route`, then you get the expected "hello world".

However, if you Cythonize this module (`cythonize -a -i test.pyx`) then you obtain the following response:

```python-traceback
        <h2>404 Not Found</h2>
        <p>The path '/route/' was not found.</p>
        <pre id="traceback">Traceback (most recent call last):
  File "/tmp/cython/cherrypy/cherrypy/_cprequest.py", line 638, in respond
    self._do_respond(path_info)
  File "/tmp/cython/cherrypy/cherrypy/_cprequest.py", line 697, in _do_respond
    response.body = self.handler()
  File "/tmp/cython/cherrypy/cherrypy/lib/encoding.py", line 219, in __call__
    self.body = self.oldhandler(*args, **kwargs)
  File "/tmp/cython/cherrypy/cherrypy/_cperror.py", line 416, in __call__
    raise self
cherrypy._cperror.NotFound: (404, "The path '/route/' was not found.")
</pre>
```

**What is the expected behavior?**

I would have expected things to "just work", even if the code had been Cythonised. I do not think that this is a Cython problem.

[...]

PR #2029, opened 2024-04-09
Add support for Cythonized API Classes 

HJJ256:
  - [ ] bug fix
  - [x] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

[#1814](https://github.com/cherrypy/cherrypy/issues/1814)

Currently, when an API class is cythonized, the methods for the class do not pass the decoratable_types check. The issue [#1814](https://github.com/cherrypy/cherrypy/issues/1814) was raised 5 years ago and is still relevant.

Cythonized API class methods are getting mounted in cherrypy.tree and the API runs successfully.

**Other information**:
I tried the approach suggested by @webknjaz to put the types.BuiltinFunctionType in the decoratable_types, however, not all cythonized methods are of types.BuiltinFunctionType. The one thing these functions have in common is the "func_code" attribute.

As for the testing of this change, I am still looking into it. Currently, I can only come up with one way where I create a .whl file and install the cythonized API as a package, then import from the package and test if CherryPy is working, but this way does not seem very efficient. 

At worst, the hasattr check may fail in later python versions (very unlikely since func_code is a core cython method attribute). This feature is important for people who want to speedup and package their code using Cython.

**Checklist**:

### DIFF
--- a/cherrypy/_helper.py
+++ b/cherrypy/_helper.py
@@ -137,7 +137,6 @@ def popargs(*args, **kwargs):
         class Root:
             def index(self):
                 #...
-
     """
     # Since keyword arg comes after *args, we have to process it ourselves
     # for lower versions of python.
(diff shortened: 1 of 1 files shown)


webknjaz:
That's an interesting approach. Do you have any references to the Cython docs that would suggest such a check? We'd need a code comment with such a reference explaining what's listed there.

It's also a good idea to integrate Cython into testing somehow.

> however, not all cythonized methods are of types.BuiltinFunctionType

What are the other types, then? AFAIU, this type corresponds to anything built with C-API, not just Cython. Checking against a known list of types seems like a safer bet to me so I'd prefer such an implementation. My feeling is that we should be able to use Cython on the fly right from within the normal CPython runtime but this needs some explotatory effort.

webknjaz:
@HJJ256 are you still interested in this?

HJJ256:
This [file](https://github.com/cython/cython/blob/master/Cython/Utility/CythonFunction.c) contains a lot of references to setting python type of the method object as "cython_function_or_method". The final type(func) shows the type as `_cython_<version>.cython_function_or_method` where version keeps changing with Cython.

This [doc](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html?utm_source=chatgpt.com#compiler-directives) states that when binding is set to True (which is the default since Cython 3.0.0), the cythonized method "behaves" as a python method. Hence the presence of func_code or `__code__` attributes.

Let me know if one of the references in this comment works for you

HJJ256:
@webknjaz Another alternative is to add this check instead.

```python
func_type = type(func)

is_cython_function = (
    func_type.__name__ == "cython_function_or_method"
    and func_type.__module__.startswith("_cython_") #This one is optional
)
```

This is what I get when I print the type of a cythonized method

```python

print(type(func))
# _cython_3_1_3.cython_function_or_method

print(type(func).__module__)
# _cython_3_1_3

print(type(func).__name__)
# cython_function_or_method
```

PR #2030, opened 2024-04-19
Update GHA sdist name to lowercase for new setuptools

radez:
CI fails because with the setuptools upgrade, the sdist names started being lowercase. The CI declaration needs to be fixed to expect a new name

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -336,7 +336,7 @@ jobs:
                 mode=FILE_APPEND_MODE,
         ) as outputs_file:
             print(
-                "sdist=${{ steps.dist.outputs.name }}-${{
+                "sdist=cherrypy-${{
                     steps.request-check.outputs.release-requested == 'true'
                     && github.event.inputs.release-version
                     || steps.scm-version.outputs.dist-version


webknjaz (commented):
As I'm copying a lot of such configurations over across many projects, it's important that the identifiers remain the same (it makes synchronizing files easier in side-by-side diff mode, for example). Could you adjust the names to be closer to 
https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/master/.github/workflows/ci-cd.yml#L108 and https://github.com/ansible/pylibssh/blob/57e2802/.github/workflows/ci-cd.yml#L358 ?

radez:
Sure thing, I'll update it tomorrow and get a new patch in.

radez:
Looks like sphinx-contrib and pylibssh don't put the sdist and wheel names into the env scope, They hardcode the file name into the step that sets the artifact names.

To match them, I've removed the step that pulls the name from env and hardcoded the names of the two artifact files into the step that sets the artifact names.

.github/workflows/ci-cd.yml
```diff
@@ -312,18 +311,6 @@ jobs:
...
-      run: |
-        from os import environ
-        from pathlib import Path
-
-        FILE_APPEND_MODE = 'a'
-
-        with Path(environ['GITHUB_OUTPUT']).open(
-                mode=FILE_APPEND_MODE,
-        ) as outputs_file:
-            print("name=${{ env.dists-name }}", file=outputs_file)
```
webknjaz: This is referenced in the job outputs around line 115/116 and is than used in the GitHub Environment URL declaration in the PyPI publishing job (line 1007/994). We need to be careful to keep that reference. Especially since that may only become evident at the time of release as that job won't run on PRs.

So instead of a blind removal, let's keep this and maybe the env var too. We could then keep reusing it for the wheel and only update that sdist bit.

webknjaz:
> Looks like sphinx-contrib and pylibssh don't put the sdist and wheel names into the env scope, They hardcode the file name into the step that sets the artifact names.

There's another consideration — they started using reusable workflows and that limits the use of the `env` references in some places. At least, I thought I've already moved some things to reusable workflows there. Maybe not yet. Here's an example where I explicitly left a comment about that limitation: https://github.com/aio-libs/frozenlist/blob/8c15ec9/.github/workflows/ci-cd.yml#L54-L55.

.github/workflows/ci-cd.yml
webknjaz: Marking as resolved.

webknjaz:
Thanks!

PR #2031, opened 2024-05-10
docs: improve pyopenssl docs in deploy.rst

ktdreyer:
Update the docs to say that `builtin` is now the default backend.

Give Python examples of how to set both backends (`builtin` and `pyopenssl`.)

  - [ ] bug fix
  - [ ] feature
  - [x] docs update
  - [ ] refactoring
  - [ ] other

### DIFF
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -208,18 +208,23 @@ openssl will then ask you a series of questions. You can enter whatever values a
 
 3. Decide whether you want to use python's built-in SSL library, or the pyOpenSSL library. CherryPy supports either.
 
-    a) *Built-in.* To use python's built-in SSL, add the following line to your CherryPy config:
+    a) *Built-in.* CherryPy uses Python's built-in SSL by default. To set it explicitly, set this line in your CherryPy config:
 
     .. code-block:: python
 
        cherrypy.server.ssl_module = 'builtin'
 
-    b) *pyOpenSSL*. Because python did not have a built-in SSL library when CherryPy was first created, the default setting is to use pyOpenSSL. To use it you'll need to install it (we could recommend you install `cython <http://cython.org/>`_ first):
+    b) *pyOpenSSL*. For better performance you can use CherryPy's pyOpenSSL backend. To use it you'll need to install it (we could recommend you install `cython <http://cython.org/>`_ first):
 
     .. code-block:: bash
 
        $ pip install cython, pyOpenSSL
 
+    Then, in your CherryPy config:
+
+    .. code-block:: python
+
+       cherrypy.server.ssl_module = 'pyopenssl'
 
 4. Add the following lines in your CherryPy config to point to your certificate files:
 


ktdreyer:
With the simple "hello world" application, I get better performance with the pyopenssl backend (testing in a container on localhost) so I think that's the main reason users would choose pyopenssl now. Is that right?

webknjaz:
I'm not sure. I think many people just put it behind a reverse proxy instead of handling TLS in-app. I have a feeling that the pyopenssl integration is a bit more buggy as of the last Cheroot release. I dream of having enough time to refactor both adapters..

ktdreyer:
Would you please merge this? It makes it much clearer that `builtin` is the default.

docs/deploy.rst
```diff
@@ -208,18 +208,23 @@ openssl will then ask you a series of questions. You can enter whatever values a
...
     .. code-block:: python
 
        cherrypy.server.ssl_module = 'builtin'
 
-    b) *pyOpenSSL*. Because python did not have a built-in SSL library when CherryPy was first created, the default setting is to use pyOpenSSL. To use it you'll need to install it (we could recommend you install `cython <http://cython.org/>`_ first):
+    b) *pyOpenSSL*. For better performance you can use CherryPy's pyOpenSSL backend. To use it you'll need to install it (we could recommend you install `cython <http://cython.org/>`_ first):
 
     .. code-block:: bash
 
        $ pip install cython, pyOpenSSL
```
webknjaz, line 221: Does this syntax with a comma event work? Also, it's unlikely to have any effect these days.

docs/deploy.rst
```diff
@@ -208,18 +208,23 @@ openssl will then ask you a series of questions. You can enter whatever values a
...
 3. Decide whether you want to use python's built-in SSL library, or the pyOpenSSL library. CherryPy supports either.
 
-    a) *Built-in.* To use python's built-in SSL, add the following line to your CherryPy config:
+    a) *Built-in.* CherryPy uses Python's built-in SSL by default. To set it explicitly, set this line in your CherryPy config:
 
     .. code-block:: python
 
        cherrypy.server.ssl_module = 'builtin'
 
-    b) *pyOpenSSL*. Because python did not have a built-in SSL library when CherryPy was first created, the default setting is to use pyOpenSSL. To use it you'll need to install it (we could recommend you install `cython <http://cython.org/>`_ first):
```
webknjaz, line 217: I think, it'd be good to keep the historic reference regarding why it was added. Also, I'm not sure if the performance bit is noticeable. Technically, both are implemented as C-extensions linking against some OpenSSL version.
I'd drop the mention of Cython too since `pyOpenSSL` is a pure-python wheel and `cyptography` ships pre-built wheels — one would have to go out of their way to force it to build from sdist, which would pick up an isolated version of Cython anyway.

docs/deploy.rst
```diff
@@ -208,18 +208,23 @@ openssl will then ask you a series of questions. You can enter whatever values a
 
 3. Decide whether you want to use python's built-in SSL library, or the pyOpenSSL library. CherryPy supports either.
 
-    a) *Built-in.* To use python's built-in SSL, add the following line to your CherryPy config:
+    a) *Built-in.* CherryPy uses Python's built-in SSL by default. To set it explicitly, set this line in your CherryPy config:
```
webknjaz, line 211: It seems like there's at least one place where `builtin` is not the default:
* https://github.com/cherrypy/cherrypy/blob/8db3d07/cherrypy/_cpnative_server.py#L153C56-L153C67

Plus, there's another place that would need to be cleaned up for this docs change to make sense:
* https://github.com/cherrypy/cherrypy/blob/8db3d07/cherrypy/_cpwsgi_server.py#L86-L89

PR #2039, opened 2024-11-15
Replace zc.lockfile with filelock

carlsmedstad:
Hey there, Arch Linux maintainer here 👋

Our  [cherrypy package](https://archlinux.org/packages/extra/any/python-cherrypy/) is currently one of only two packages depending on the orphaned [zc.lockfile package](https://archlinux.org/packages/extra/any/python-zc.lockfile/) in the Arch Linux repositories. Migrating to  [filelock](https://github.com/tox-dev/filelock) would allow us to eventually drop `zc.lockfile` entirely from our repos, reducing the maintenance workload.

We might decide to patch this in anyways, since it was fairly non-intrusive, but I'd thought I'd check if we could upstream it first. Cheers!

### DIFF
--- a/cherrypy/lib/sessions.py
+++ b/cherrypy/lib/sessions.py
@@ -108,7 +108,7 @@ import threading
 import binascii
 import pickle
 
-import zc.lockfile
+from filelock import FileLock, Timeout
 
 import cherrypy
 from cherrypy.lib import httputil
@@ -579,19 +579,21 @@ class FileSession(Session):
         path += self.LOCK_SUFFIX
         checker = locking.LockChecker(self.id, self.lock_timeout)
         while not checker.expired():
+            lock = FileLock(path)
+            timeout = 0.1
             try:
-                self.lock = zc.lockfile.LockFile(path)
-            except zc.lockfile.LockError:
-                time.sleep(0.1)
-            else:
+                lock.acquire(timeout=timeout)
+                self.lock = lock
                 break
+            except Timeout:
+                time.sleep(timeout)
         self.locked = True
         if self.debug:
             cherrypy.log('Lock acquired.', 'TOOLS.SESSIONS')
 
     def release_lock(self, path=None):
         """Release the lock on the currently-loaded session data."""
-        self.lock.close()
+        self.lock.release()
         self.locked = False
 
     def clean_up(self):
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ params = dict(
         'cheroot>=8.2.1',
         'portend>=2.1.1',
         'more_itertools',
-        'zc.lockfile',
+        'filelock',
         'jaraco.collections',
         'importlib-metadata; python_version<="3.7"',
     ],


webknjaz:
Thank you for the suggestion! `zc.lockfile` has a history of causing us problems.
I suggested changing it to `portalocker` in the past (https://github.com/cherrypy/cherrypy/issues/1193#issuecomment-803186321) and I'm not sure if I saw that `filelock` is a thing at that point.

Do you know of any comparison between the two? In particular, it's important for us that it's fully cross-platform.

It looks like the projects are 14 and 11 years old and are both active.

But it doesn't appear like they know of each other: https://github.com/search?q=repo%3Awolph%2Fportalocker%20filelock&type=code / https://github.com/tox-dev//issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+portalocker.

webknjaz:
GH tacks more dependents @ tox's thing: https://github.com/wolph/portalocker/network/dependents / https://github.com/tox-dev/filelock/network/dependents.

carlsmedstad:
> Do you know of any comparison between the two? In particular, it's important for us that it's fully cross-platform.

Regarding functional differences, I'm afraid I don't have any in-depth knowledge. `filelock` seems to be fully cross-platform [according to themselves](https://py-filelock.readthedocs.io/en/latest/index.html):

> This package contains a single module, which implements a platform independent file lock in Python, which provides a simple way of inter-process communication

From an Arch Linux point-of-view, the comparison is heavily in favor of `filelock`:

- We already package [python-filelock](https://archlinux.org/packages/extra-testing/any/python-filelock/), it has 25 downstream dependencies in the repos, and 50 in the AUR:

  ```sh
  $ curl -sSf "https://aur.archlinux.org/rpc/v5/search/python-filelock?by=depends" | jq '.resultcount'
  50
  ```
  
  As for `portalocker`, it's currently only available [in the AUR](https://aur.archlinux.org/packages/python-portalocker), and it only has 18 downstream dependencies there:
  
  ```sh
  $ curl -sSf "https://aur.archlinux.org/rpc/v5/search/python-portalocker?by=depends" | jq '.resultcount'
  18
  ```
  
- `filelock` is packaged in more distributions overall - according to Repology it is packaged in [160 repositories](https://repology.org/project/python%3Afilelock/information) in comparison to [67](https://repology.org/project/python:portalocker/information) for `portalocker`.

jaraco (approved):

carlsmedstad:
Hey again, anything I can do to help get this merged? Thanks!

jaraco:
@webknjaz This PR appears not to be reviewable any longer, due to the pre-commit changes. What's the strategy here and how to get this PR over the line?

webknjaz:
@jaraco should revert https://github.com/cherrypy/cherrypy/commit/5fe3dcef35d44ab60e32e5130d2e37089820bc4a first, as it broke the linting job (it was green before the change incompatible with the linting setup was made).

webknjaz:
I should probably dedicate some time today. I was going to work on upgrading the CI anyway.

jaraco:
> @jaraco should revert [5fe3dce](https://github.com/cherrypy/cherrypy/commit/5fe3dcef35d44ab60e32e5130d2e37089820bc4a) first, as it broke the linting job (it was green before the change incompatible with the linting setup was made).

Done. This change probably happened as part of me applying ruff changes across the projects I maintain.

@carlsmedstad , you may want to reset to 71b69d05240415fb75eb0d0dea13d0c9697d939e rather than resolve the conflicts again.

webknjaz:
Yeah, I figured. I saw a few other bits in CI that need fixing. I'll tend to them.

carlsmedstad:
@jaraco Done! Thanks for helping this get across the line

webknjaz:
@carlsmedstad looks like you still need to rebase?

carlsmedstad:
@webknjaz Ah thanks, should be good now

SpangleLabs:
I've noticed that this is merged, but we've not actually had a new release of CherryPy since this. Would it be possible to make a new release, such that we don't need `zc.lockfile` if we're pulling CherryPy from pypi?

PR #2040, opened 2024-11-22
Fix `test_callable_spec` to handle keyword only parameters

the-allanc:
  - [X] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

None.

If a TypeError occurs inside the execution of the handler, and the handler has had keyword parameters passed to it, _test_callable_spec_ will cause a 404 to be returned, rather than allowing the _TypeError_ to be correctly propagated as a 500 response.  

We delegate validation of the provided args and kwargs against the callable to the inspect module, which adds support for keyword-only parameters while removing the need to perform these checks directly in CherryPy. In the situation mentioned previously, this means we correctly raise a 500 error instead of returning a 404 response.

**Other information**:

I've kept the behaviour compatible with what was previously there regarding reporting user-friendly messages when "show_mismatched_params" is enabled (all previous tests pass), but I've also added additional checks to _testParamErrors_ which will fail if using the current code in _cpdispatch.py. There were also some bugs in the tests which I found while making these changes that I've fixed.

**Checklist**:

  - [X] Documentation reflects the changes

### DIFF
--- a/cherrypy/_cpdispatch.py
+++ b/cherrypy/_cpdispatch.py
@@ -10,6 +10,8 @@ The default dispatcher discovers the page handler by matching path_info
 to a hierarchical arrangement of objects, starting at request.app.root.
 """
 
+import inspect
+import re
 import string
 import sys
 import types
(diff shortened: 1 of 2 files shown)


the-allanc:
Currently tests fail in versions of Python that have reached EOL - see #2041.

webknjaz:
Clicking “rebase” in this PR...

webknjaz (changes requested):
`cherrypy/test/test_request_obj.py::RequestObjectTests::testParamErrors` is broken on Python 3.12. It probably needs some compat fallbacks for `>= 3.12`. I haven't yet added 3.13 to the CI but I expect it to fail the same.

webknjaz:
@the-allanc FYI it looks like the email in the commits (`allan AT sixtyten PERIOD org`) isn't verified in your GitHub account. So GH isn't attributing the commits to you. You can fix that @ https://github.com/settings/emails or by changing the committer with `git commit --author='...'`.

PR #2051, opened 2025-03-26
Tests: don't use an ephemeral port

FliegendeWurst:
Ports >= 32768 may already be [in use for outgoing TCP connections](https://en.wikipedia.org/wiki/Ephemeral_port), so we must use a lower port number here.

  - [x] bug fix

None, as far as I know

Tests sometimes randomly fail, if port 54583 is already used.

Tests only fail if port 24583 is in use. This is not a standard port number so failure is unlikely.

**Checklist**:

  - [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences

### DIFF
--- a/cherrypy/test/benchmark.py
+++ b/cherrypy/test/benchmark.py
@@ -7,8 +7,8 @@ Usage:
 --notests:     start the server but do not run the tests; this allows
                you to check the tested pages with a browser
 --help:        show this help message
---cpmodpy:     run tests via apache on 54583 (with the builtin _cpmodpy)
---modpython:   run tests via apache on 54583 (with modpython_gateway)
+--cpmodpy:     run tests via apache on 24583 (with the builtin _cpmodpy)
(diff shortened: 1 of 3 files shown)


webknjaz:
Why not use `0` then?

FliegendeWurst:
Then it would be more difficult to figure out which port to connect to, right?

webknjaz:
It'd require looking up the actually allocated port on the actual socket, but it's not that difficult. I just don't think that it's worth changing one hardcoded number to another hardcoded number.

FliegendeWurst:
The new hardcoded number has the benefit that it won't fail *randomly* depending on the number of active TCP connections.

FliegendeWurst:
I'll take a look if I can use port 0 and pass along the allocated port number.

webknjaz:
@FliegendeWurst are you still interested in this?

FliegendeWurst:
I'm no longer interested in this change.

Issue #2060, opened 2025-05-22
HTTP Digest Access Authentication on FIPS-compliant hosts


Helene:
Dear cherryPy development team, 
By default, CherryPy uses the [Python hashlib MD5 algorithm](https://docs.python.org/3/library/hashlib.html#hashlib.md5) for HTTP Digest Access Authentication validation. Some operating systems may block md5 if a 'FIPS compliant' build of Python is in use on the host. Here is an example of the code failure:
```python-traceback
File "../cherrypy/lib/auth_digest.py", line 410, in digest_auth
    respond_401()
  File ".../cherrypy/lib/auth_digest.py", line 453, in _respond_401
    header = www_authenticate(
  File ".../cherrypy/lib/auth_digest.py", line 365, in www_authenticate
    nonce = synthesize_nonce(realm, key)
  File ".../cherrypy/lib/auth_digest.py", line 132, in synthesize_nonce
    h = md5_hex('%s:%s:%s' % (timestamp, s, key))
  File ".../cherrypy/lib/auth_digest.py", line 37, in md5_hex
    return md5(ntob(s, 'utf-8')).hexdigest()
ValueError: [digital envelope routines] unsupported
```
Our team would like to continue using the CherryPy framework with HTTP Digest Access Authentication on incoming requests, but we need to find a solution for users who have MD5 blocked in their environment. The SHA512 algorithm, for example, would meet our expectations.
Would it be possible to allow users to optionally set the `HttpDigestAuthorization.algorithm`? 
Example:
```python
    userpassdict = {'alice' : '4x5istwelve'}
    get_ha1 = cherrypy.lib.auth_digest.get_ha1_dict_plain(userpassdict)
    digest_auth = {'tools.auth_digest.on': True,
                   'tools.auth_digest.realm': 'wonderland',
                   'tools.auth_digest.algorithm': 'SHA512',           <<<<
                   'tools.auth_digest.get_ha1': get_ha1,
                   'tools.auth_digest.key': 'a565c27146791cfb',
                   'tools.auth_digest.accept_charset': 'UTF-8',
    }
    app_config = { '/' : digest_auth }
```
Based on the algorithm type provided in the tools.auth_digest option, the get_ha1 method would need to decide which algorithm to use.

This is just an idea for how it could be solved generically.
If you have an alternative solution, please feel free to share it. Your feedback would be much appreciated!

webknjaz:
[...]

PR #2066, opened 2025-07-11
Extend digest_auth tool with all supported hashing methods

Helene:
  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [X] extension

#2060 

**Other information**:

**Checklist**:

### DIFF
--- a/cherrypy/lib/auth_digest.py
+++ b/cherrypy/lib/auth_digest.py
@@ -100,11 +100,7 @@ def get_ha1_dict_plain(user_password_dict):
     """
 
     def get_ha1(**kwargs):
-        if (
-            realm in kwargs
-            and username in kwargs 
-            and algorithm in kwargs
-        ):
+        if realm in kwargs and username in kwargs and algorithm in kwargs:
             password = user_password_dict.get(username)
             if password:
                 return H('%s:%s:%s' % (username, realm, password), algorithm)
@@ -125,10 +121,7 @@ def get_ha1_dict(user_ha1_dict):
     """
 
     def get_ha1(**kwargs):
-        if (
-            realm in kwargs 
-            and username in kwargs
-        ):
+        if realm in kwargs and username in kwargs:
             return user_ha1_dict.get(username)
         return None
 
@@ -153,9 +146,7 @@ def get_ha1_file_htdigest(filename):
 
     def get_ha1(**kwargs):
         result = None
-        if (realm in kwargs 
-            and username in kwargs
-        ):
+        if realm in kwargs and username in kwargs:
             with open(filename, 'r') as f:
                 for line in f:
                     u, r, ha1 = line.rstrip().split(':')
@@ -535,10 +526,10 @@ def digest_auth(
     # perform some correctness checks
     if auth.algorithm not in server_algorithms:
         raise ValueError(
-                self.errmsg(
-                    "Algorithm not supported by server: '%s'" % auth.algorithm,
-                ),
-            )
+            self.errmsg(
+                "Algorithm not supported by server: '%s'" % auth.algorithm,
+            ),
+        )
 
     if debug:
         TRACE(str(auth))
@@ -549,7 +540,7 @@ def digest_auth(
     ha1 = get_ha1(
         realm=realm,
         username=auth.username,
-        algorithm=auth.algorithm
+        algorithm=auth.algorithm,
     )
 
     if ha1 is None:
--- a/cherrypy/lib/auth_digest.py
+++ b/cherrypy/lib/auth_digest.py
@@ -403,12 +403,12 @@ class HttpDigestAuthorization(object):
 
         # RFC 2617 3.2.2.2
         #
-        # If the "algorithm" directive's value is "<algorithm>" or 
+        # If the "algorithm" directive's value is "<algorithm>" or
         # is unspecified, then A1 is:
         #    A1 = unq(username-value) ":" unq(realm-value) ":" passwd
         #
-        # If the "algorithm" directive's value is "<algorithm>-sess", 
-        # then A1 is calculated only once - 
+        # If the "algorithm" directive's value is "<algorithm>-sess",
+        # then A1 is calculated only once -
         # on the first request by the client following
         # receipt of a WWW-Authenticate challenge from the server.
         # A1 = H( unq(username-value) ":" unq(realm-value) ":" passwd )
@@ -569,8 +569,8 @@ def _respond_401(realm, key, accept_charset, debug, **kwargs):
         #        H(data) = SHA-256(data)
         #
         # i.e., the digest is the "<algorithm>" of the secret concatenated with
-        # a colon concatenated with the data. The  "<algorithm>-sess" is 
-        # intended to allow efficient third-party authentication servers; 
+        # a colon concatenated with the data. The  "<algorithm>-sess" is
+        # intended to allow efficient third-party authentication servers;
         # for the difference in usage, see the description in Section 3.4.2.
         if not algorithm.endwith('-sess'):
             header = www_authenticate(
@@ -583,7 +583,7 @@ def _respond_401(realm, key, accept_charset, debug, **kwargs):
             if debug:
                 TRACE(header)
             # TODO: with that solution 'WWW-Authenticate' header will be
-            # overwritten with last one fromm loop iteration. 
+            # overwritten with last one fromm loop iteration.
             # Need a solution for sending a list
             # of 'WWW-Authenticate' headers with single 401 response
             cherrypy.serving.response.headers['WWW-Authenticate'] = header
--- a/cherrypy/lib/auth_digest.py
+++ b/cherrypy/lib/auth_digest.py
@@ -54,7 +54,7 @@ def sha512_hex(s):
 
 
 def choose_hexdigest_method(algorithm):
-    if algorithm.endwith("-sess"):
+    if algorithm.endwith('-sess'):
         algorithm = algorithm.split('-')[0]
     return algorithm.lower() + '_hex'
 
@@ -67,7 +67,14 @@ valid_qops = (qop_auth, qop_auth_int)
 # TODO: validate HTTP response algorithm syntax
 # according rfc7616 it could be 'SHA-256' instead of 'SHA256'
 # if so additional handilng of '-' required
-valid_algorithms = {'MD5', 'MD5-sess', 'SHA256', 'SHA256-sess', 'SHA512', 'SHA512-sess'}
+valid_algorithms = {
+    'MD5',
+    'MD5-sess',
+    'SHA256',
+    'SHA256-sess',
+    'SHA512',
+    'SHA512-sess',
+}
 
 
 def updte_valid_algoritms(algorithms):
@@ -410,7 +417,7 @@ class HttpDigestAuthorization(object):
         # receipt of a WWW-Authenticate challenge from the server.
         # A1 = H( unq(username-value) ":" unq(realm-value) ":" passwd )
         #         ":" unq(nonce-value) ":" unq(cnonce-value)
-        if self.algorithm.endwith("-sess"):
+        if self.algorithm.endwith('-sess'):
             ha1 = H('%s:%s:%s' % (ha1, self.nonce, self.cnonce))
 
         digest = H('%s:%s' % (ha1, req))
@@ -459,7 +466,14 @@ def www_authenticate(
     )
 
 
-def digest_auth(realm, get_ha1, key, algoitms=valid_algorithms, debug=False, accept_charset='utf-8'):
+def digest_auth(
+    realm,
+    get_ha1,
+    key,
+    algoitms=valid_algorithms,
+    debug=False,
+    accept_charset='utf-8',
+):
     """Perform HTTP Digest Access Authentication.
 
     A CherryPy tool that hooks at ``before_handler`` to perform
@@ -487,7 +501,6 @@ def digest_auth(realm, get_ha1, key, algoitms=valid_algorithms, debug=False, acc
         A secret string known only to the server, used in the synthesis
         of nonces.
     """
-
     updte_valid_algoritms(algorithms)
 
     request = cherrypy.serving.request
@@ -560,7 +573,7 @@ def _respond_401(realm, key, accept_charset, debug, **kwargs):
         # colon concatenated with the data. The  "<algorithm>-sess" is intended to
         # allow efficient third-party authentication servers; for the difference in
         # usage, see the description in Section 3.4.2.
-        if not algorithm.endwith("-sess"):
+        if not algorithm.endwith('-sess'):
             header = www_authenticate(
                 realm,
                 key,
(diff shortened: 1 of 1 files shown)


cherrypy/lib/auth_digest.py
```diff
@@ -6,14 +6,19 @@
 An implementation of the server-side of HTTP Digest Access
 Authentication, which is described in :rfc:`2617`.
 
+Digest algorithms, described in :rfc:7616, are `valid_algorithms`
```
webknjaz: this is proper RST
```suggestion
Digest algorithms, described in :rfc:`7616`, are ``valid_algorithms``
```

cherrypy/lib/auth_digest.py
```diff
@@ -6,14 +6,19 @@
...
+but could be limited to a `server_supported_algorithms`
```
webknjaz:
```suggestion
but could be limited to a smaller subset.
```

cherrypy/lib/auth_digest.py
```diff
@@ -38,11 +43,43 @@ def md5_hex(s):
...
+    'MD5',
+    'MD5-sess',
+    'SHA-256',
+    'SHA-256-sess',
+    'SHA-512',
+    'SHA-512-sess',
+}
+
+
+def updte_valid_algorithms(algorithms):
```
webknjaz: There's a typo in the wrd updte here. Also, I don't think that it's acceptable to have a mechanism for manipulating global state like this.

cherrypy/lib/auth_digest.py
```diff
@@ -135,20 +172,25 @@ def synthesize_nonce(s, key, timestamp=None):
...
     return nonce
 
 
-def H(s):
-    """Return an ``md5`` HEX hash."""
-    return md5_hex(s)
+def H(s, algorithm):
+    """Return an "<algorithm>" HEX hash."""
+    hex_method = choose_hexdigest_method(algorithm)
+    func_to_call = eval(hex_method)
```
webknjaz: Please, avoid eval at all cost. It's almost never acceptable in any programming language and would be the main source of security vulnerabilities for years to come. Besides, this sort of dynamism is making it rather difficult to reason about the control flow.

cherrypy/lib/auth_digest.py
```diff
@@ -417,7 +461,14 @@ def www_authenticate(
     )
 
 
-def digest_auth(realm, get_ha1, key, debug=False, accept_charset='utf-8'):
+def digest_auth(
+    realm,
+    get_ha1,
+    key,
+    algorithms=valid_algorithms,
```
webknjaz: When modifying the public API signatures, always add new arguments at the end of the list for backward compatibility.

cherrypy/lib/auth_digest.py
```diff
@@ -445,6 +496,8 @@ def digest_auth(realm, get_ha1, key, debug=False, accept_charset='utf-8'):
         A secret string known only to the server, used in the synthesis
         of nonces.
     """
+    updte_valid_algorithms(algorithms)
```
webknjaz: This should probably be computed here rather than mutating the module globals.

cherrypy/lib/auth_digest.py
```diff
@@ -475,7 +528,10 @@ def digest_auth(realm, get_ha1, key, debug=False, accept_charset='utf-8'):
     if not auth.validate_nonce(realm, key):
         respond_401()
 
-    ha1 = get_ha1(realm, auth.username)
+    if get_ha1.__name__ == 'get_ha1_dict_plain':
```
webknjaz: Please refrain from applying hacks that depend on obscure internal attribute access rather than just structuring code in a way that would call functions as is.

cherrypy/lib/auth_digest.py
Helene: done

cherrypy/lib/auth_digest.py
Helene: done

cherrypy/lib/auth_digest.py
Helene: done

cherrypy/lib/auth_digest.py
Helene: done

cherrypy/lib/auth_digest.py
Helene: done

cherrypy/lib/auth_digest.py
```diff
@@ -109,20 +148,23 @@ def get_ha1_file_htdigest(filename):
     argument be an absolute path, to avoid problems.
     """
 
-    def get_ha1(realm, username):
+    def get_ha1(**kwargs):
```
webknjaz, line 153: Avoid replacing explicit args with dynamic kwargs. It makes it impossible to follow all the premutations of what might happen.

cherrypy/lib/auth_digest.py
```diff
@@ -67,10 +99,14 @@ def get_ha1_dict_plain(user_password_dict):
...
     """
 
-    def get_ha1(realm, username):
-        password = user_password_dict.get(username)
-        if password:
-            return md5_hex('%s:%s:%s' % (username, realm, password))
+    def get_ha1(**kwargs):
+        realm = kwargs.get('realm')
+        username = kwargs.get('username')
+        algorithm = kwargs.get('algorithm')
```
webknjaz, line 107: It's unclear why this is done but looks wrong.

cherrypy/lib/auth_digest.py
```diff
@@ -6,14 +6,19 @@
 An implementation of the server-side of HTTP Digest Access
 Authentication, which is described in :rfc:`2617`.
 
+Digest algorithms, described in :rfc:`7616`, are ``valid_algorithms``
```
webknjaz: This probably shouldn't reference a variable that doesn't even show up in the docs: https://cherrypy--2066.org.readthedocs.build/en/2066/pkg/cherrypy.lib.auth_digest.html#module-cherrypy.lib.auth_digest. This would be misleading. Use plain text to explain things.

cherrypy/lib/auth_digest.py
```diff
@@ -38,11 +43,38 @@ def md5_hex(s):
...
+
+def sha512_hex(s):
+    """Return hexdigest of sha512sum."""
+    return sha512(ntob(s, 'utf-8')).hexdigest()
+
+
+def choose_hexdigest_method(algorithm):
+    """Return Python hexdigest method for given algorithm name."""
+    if algorithm.endwith('-sess'):
+        algorithm = algorithm.rsplit('-', 1)[0]
```
webknjaz:
```suggestion
    algorithm = algorithm.removesuffix('-sess')
```

cherrypy/lib/auth_digest.py
```diff
@@ -417,7 +467,14 @@ def www_authenticate(
...
 
 
-def digest_auth(realm, get_ha1, key, debug=False, accept_charset='utf-8'):
+def digest_auth(
+    realm,
+    get_ha1,
+    key,
+    debug=False,
+    accept_charset='utf-8',
+    algorithms=valid_algorithms,
```
webknjaz, line 478: Instead of using a mutable function default (`set()`), rely on a sentinel object (like `None` or `object()`) and implement a fallback in the function. Alternatively, change it to a `frozenset()`.

cherrypy/lib/auth_digest.py
```diff
@@ -38,11 +43,38 @@ def md5_hex(s):
...
-valid_algorithms = ('MD5', 'MD5-sess')
+
+valid_algorithms = {
+    'MD5',
+    'MD5-sess',
+    'SHA-256',
+    'SHA-256-sess',
+    'SHA-512',
+    'SHA-512-sess',
+}
```
webknjaz:
```suggestion
valid_algorithms = frozenset({
    'MD5',
    'MD5-sess',
    'SHA-256',
    'SHA-256-sess',
    'SHA-512',
    'SHA-512-sess',
})
```

cherrypy/lib/auth_digest.py
```diff
@@ -445,6 +502,8 @@ def digest_auth(realm, get_ha1, key, debug=False, accept_charset='utf-8'):
         A secret string known only to the server, used in the synthesis
         of nonces.
     """
+    server_algorithms = valid_algorithms.intersection(algorithms)
```
webknjaz: Let's just use this equivalent syntax:
```suggestion
    server_algorithms = valid_algorithms & algorithms
```

cherrypy/lib/auth_digest.py
```diff
@@ -469,13 +529,23 @@ def digest_auth(realm, get_ha1, key, debug=False, accept_charset='utf-8'):
             accept_charset=accept_charset,
         )
 
+    # perform some correctness checks
+    if auth.algorithm not in server_algorithms:
+        raise ValueError(
+            "Algorithm not supported by server: '%s'" % auth.algorithm,
+        )
```
webknjaz: We should probably communicate this to the client over HTTP.

cherrypy/lib/auth_digest.py
```diff
@@ -501,17 +571,35 @@ def digest_auth(realm, get_ha1, key, debug=False, accept_charset='utf-8'):
...
-        realm,
-        key,
-        accept_charset=accept_charset,
-        **kwargs,
-    )
-    if debug:
-        TRACE(header)
-    cherrypy.serving.response.headers['WWW-Authenticate'] = header
+    for algorithm in algorithms:
+        # for more details please refer to rfc7616 documentation
```
webknjaz, line 578: It's best if code comments don't point to external places without explaining what those references mean with words. I have no idea which details you'd like me to understand and what conclusion to reach based on this instruction.

cherrypy/lib/auth_digest.py
```diff
@@ -501,17 +571,35 @@ def digest_auth(realm, get_ha1, key, debug=False, accept_charset='utf-8'):
...
+                algorithm,
+                accept_charset=accept_charset,
+                **kwargs,
+            )
+            if debug:
+                TRACE(header)
+            # TODO: with that solution 'WWW-Authenticate' header will be
+            # overwritten with last one fromm loop iteration.
+            # Need a solution for sending a list
+            # of 'WWW-Authenticate' headers with single 401 response
```
webknjaz, line 602: This should be tackled first in a separate PR.

cherrypy/lib/auth_digest.py
Helene: good point. thanks!

cherrypy/lib/auth_digest.py
Helene: @webknjaz,
my problem is, that we have 3 definitions of get_ha1():
- (A) inside get_ha1_dict_plain the function requires 3 argumens for validtaion: username, realm, algorithm
- (B) inside get_ha1_dict only username
- (C) inside get_ha1_file_htdigest reguired username, realm

get_ha1 is passed through as callable to digest_auth.
Depends on server settings the callable (A), (B) or (C) will be executed with respect of required arguments.
I don't know better approach for handling callable with different number of arguments.

Feel free to share your thoughts on possible solutions. It would be much appreciated!

cherrypy/lib/auth_digest.py
Helene: ok. I'll do it.

cherrypy/lib/auth_digest.py
Helene: @webknjaz,
 I have opened #2069 for multiple header type support. Are you going to provide a solution for that?

cherrypy/lib/auth_digest.py
Helene: agree

cherrypy/lib/auth_digest.py
```diff
@@ -6,14 +6,20 @@
 An implementation of the server-side of HTTP Digest Access
 Authentication, which is described in :rfc:`2617`.
 
+All digest algorithms described in :rfc:`7616`, are acceptable.
+Howeever, they may be limited to a smaller subset supported by
```
webknjaz, line 10:
```suggestion
However, they may be limited to a smaller subset supported by
```

cherrypy/lib/auth_digest.py
webknjaz: We could have default values set to sentinel objects or `None`. That's pretty common in Python. Arbitrarily existing args are difficult to analyze statically or cover with tests otherwise.

cherrypy/lib/auth_digest.py
```diff
@@ -135,20 +179,26 @@ def synthesize_nonce(s, key, timestamp=None):
...
     return nonce
 
 
-def H(s):
-    """Return an ``md5`` HEX hash."""
-    return md5_hex(s)
+def H(s, algorithm):
+    """Return an "<algorithm>" HEX hash."""
+    hex_method = choose_hexdigest_method(algorithm)
+    func = globals()[hex_method]
```
webknjaz, line 199: Use a var with a mapping instead of scanning the dynamically populated global env var registry and hoping it works.

cherrypy/lib/auth_digest.py
```diff
@@ -38,11 +44,39 @@ def md5_hex(s):
...
+    """Return hexdigest of sha256sum."""
+    return sha256(ntob(s, 'utf-8')).hexdigest()
+
+
+def sha512_hex(s):
+    """Return hexdigest of sha512sum."""
+    return sha512(ntob(s, 'utf-8')).hexdigest()
+
+
+def choose_hexdigest_method(algorithm):
```
webknjaz, line 57: This function could just return a callable, one of `md5()`, `sha256()` or `sha512()`. And the consumer would invoke it in a uniform way.

cherrypy/lib/auth_digest.py
```diff
@@ -109,20 +150,23 @@ def get_ha1_file_htdigest(filename):
     argument be an absolute path, to avoid problems.
     """
 
-    def get_ha1(realm, username):
+    def get_ha1(**kwargs):
         result = None
```
webknjaz, line 154: This way, we can avoid adding even more nesting. It's called a "guard expression".
```suggestion
        if realm is None or username is None:
            return None

        result = None
```

cherrypy/lib/auth_digest.py
```diff
@@ -109,20 +150,23 @@ def get_ha1_file_htdigest(filename):
...
+                    if u == username and r == realm:
+                        result = ha1
+                        break
         return result
 
     return get_ha1
 
 
-def synthesize_nonce(s, key, timestamp=None):
+def synthesize_nonce(s, key, algorithm, timestamp=None):
```
webknjaz, line 169: Normally, we shouldn't change public API signatures in a way that reorders the args as this would be backwards-incompatible. Append new args to the end.

cherrypy/lib/auth_digest.py
webknjaz: If I'm lucky to have time for this. I don't know yet.

PR #2074, opened 2025-11-16
Add private key password configuration option

### LINKED ISSUE #1583: Password support for SSL certificates
* **I'm submitting a ...**
[ ] bug report
[x] feature request

`SSLContext.load_cert_chain(certfile, keyfile=None, password=None) supports a password argument` (new in python 3.3) to use for encrypted private keys. Please make this available in `CherryPyWSGIServer`.

#1165 related but not relevant anymore (i think)

jatalahd:
- Adds an option to insert private key password in code or configuration file to avoid typing it manually every time server starts.

- With other commit in cheroot, fixes #1583

  - [ ] bug fix
  - [ X] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

#1583

If ssl is used and private key has password protection,
the password needs to be entered manually every time the server starts.

With this change, the private key password can be inserted via code
or in configuration file.

**Other information**:

**Checklist**:

  - [ X] I think the code is well written
  - [ X] I wrote [good commit messages][1]
  - [ X] I have [squashed related commits together][2] after the changes have been approved
  - [ X] I used the same coding conventions as the rest of the project
  - [ X] The new code doesn't generate linter offenses
  - [ X] The PR relates to *only* one subject with a clear title

### DIFF
--- a/cherrypy/_cpwsgi_server.py
+++ b/cherrypy/_cpwsgi_server.py
@@ -89,23 +89,25 @@ class CPWSGIServer(cheroot.wsgi.Server):
             ssl_module = self.server_adapter.ssl_module or 'builtin'
         else:
             ssl_module = self.server_adapter.ssl_module or 'pyopenssl'
-        if self.server_adapter.ssl_context:
+
+        setup_tls = (
+            self.server_adapter.ssl_context
+            or self.server_adapter.ssl_certificate
+        )
+
+        if setup_tls:
             adapter_class = cheroot.server.get_ssl_adapter_class(ssl_module)
+            key_password = self.server_adapter.ssl_private_key_password
             self.ssl_adapter = adapter_class(
-                self.server_adapter.ssl_certificate,
-                self.server_adapter.ssl_private_key,
-                self.server_adapter.ssl_certificate_chain,
-                self.server_adapter.ssl_ciphers,
+                certificate=self.server_adapter.ssl_certificate,
+                private_key=self.server_adapter.ssl_private_key,
+                certificate_chain=self.server_adapter.ssl_certificate_chain,
+                ciphers=self.server_adapter.ssl_ciphers,
+                private_key_password=key_password,
             )
-            self.ssl_adapter.context = self.server_adapter.ssl_context
-        elif self.server_adapter.ssl_certificate:
-            adapter_class = cheroot.server.get_ssl_adapter_class(ssl_module)
-            self.ssl_adapter = adapter_class(
-                self.server_adapter.ssl_certificate,
-                self.server_adapter.ssl_private_key,
-                self.server_adapter.ssl_certificate_chain,
-                self.server_adapter.ssl_ciphers,
-            )
+
+            if self.server_adapter.ssl_context:
+                self.ssl_adapter.context = self.server_adapter.ssl_context
 
         self.stats['Enabled'] = getattr(
             self.server_adapter,
--- a/cherrypy/test/test_config_ssl_adapter.py
+++ b/cherrypy/test/test_config_ssl_adapter.py
@@ -1,0 +1,147 @@
+import pytest
+import trustme
+import ssl
+import cherrypy
+from cherrypy._cpcompat import HTTPSConnection
+from cherrypy import _cpwsgi_server, _cpnative_server
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.serialization import (
+    BestAvailableEncryption,
+    Encoding,
+    PrivateFormat,
+    load_pem_private_key,
+)
+
+
+@pytest.fixture(scope='session')
+def private_key_password():
+    """Provide hardcoded password for private key."""
+    return 'криївка'
+
+
+@pytest.fixture
+def ssl_cert_and_key(tmp_path, private_key_password):
+    """Create certificate chain and encrypted key as one pem."""
+    ca = trustme.CA()
+    leaf_cert = ca.issue_cert('localhost', '127.0.0.1', '::1')
+    key_as_bytes = leaf_cert.private_key_pem.bytes()
+    private_key_object = load_pem_private_key(
+        key_as_bytes,
+        password=None,
+        backend=default_backend(),
+    )
+
+    encrypted_key_as_bytes = private_key_object.private_bytes(
+        encoding=Encoding.PEM,
+        format=PrivateFormat.PKCS8,
+        encryption_algorithm=BestAvailableEncryption(
+            password=private_key_password.encode('utf-8'),
+        ),
+    )
+
+    key_file = tmp_path / 'encrypted_cert.pem'
+    key_file.write_bytes(
+        encrypted_key_as_bytes
+        + b''.join(cert.bytes() for cert in leaf_cert.cert_chain_pems)
+        + ca.cert_pem.bytes(),
+    )
+
+    return key_file
+
+
+@pytest.fixture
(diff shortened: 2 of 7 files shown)


cherrypy/_cpwsgi_server.py
```diff
@@ -89,23 +89,26 @@ def __init__(self, server_adapter=cherrypy.server):
...
             ssl_module = self.server_adapter.ssl_module or 'pyopenssl'
+
+        ssl_args = [
+            self.server_adapter.ssl_certificate,
+            self.server_adapter.ssl_private_key,
+            self.server_adapter.ssl_certificate_chain,
+            self.server_adapter.ssl_ciphers,
+        ]
+        ssl_kwargs = {}
+        if self.server_adapter.ssl_private_key_password is not None:
```
webknjaz: Why make this conditional?

cherrypy/_cpwsgi_server.py
```diff
@@ -89,23 +89,26 @@ def __init__(self, server_adapter=cherrypy.server):
             ssl_module = self.server_adapter.ssl_module or 'builtin'
         else:
             ssl_module = self.server_adapter.ssl_module or 'pyopenssl'
+
+        ssl_args = [
```
webknjaz: Can we just use kwargs for everything?

cherrypy/_cpwsgi_server.py
webknjaz: If this is meant for compatibility with older Cheroot, I think we could just bump the minimum version in the packaging metadata instead.

cherrypy/_cpwsgi_server.py
```diff
@@ -89,23 +89,26 @@ def __init__(self, server_adapter=cherrypy.server):
...
+            self.server_adapter.ssl_ciphers,
+        ]
+        ssl_kwargs = {}
+        if self.server_adapter.ssl_private_key_password is not None:
+            ssl_kwargs['private_key_password'] = (
+                self.server_adapter.ssl_private_key_password
+            )
+
         if self.server_adapter.ssl_context:
             adapter_class = cheroot.server.get_ssl_adapter_class(ssl_module)
```
webknjaz: Perhaps, we could collapse these branches and then get rid of the additional args?
```suggestion
        set_up_tls = self.server_adapter.ssl_context or self.server_adapter.ssl_certificate
        if set_up_tls:
            adapter_class = cheroot.server.get_ssl_adapter_class(ssl_module)
            self.ssl_adapter = adapter_class(*ssl_args, **ssl_kwargs)  # FIXME: drop the indirectly unwrapped args in favor of explicit
            if self.server_adapter.ssl_context:
                self.ssl_adapter.context = self.server_adapter.ssl_context
```

cherrypy/_cpwsgi_server.py
jatalahd: Yes, it was for backwards compatibility. Now removed in the latest push.

cherrypy/_cpwsgi_server.py
jatalahd: Yes, much better that way.

cherrypy/_cpwsgi_server.py
jatalahd: Done as suggested. There was duplicate code, so with this change it is much better.

webknjaz: Marking as resolved. Will leave comments on the new diff.

cherrypy/_cpwsgi_server.py
webknjaz: @jatalahd plz also bump Cheroot's minimum version to whatever introduced this interface first in `pyproject.toml`.

webknjaz (commented):
To be mergeable, this will need a change in the core packaging metadata (min Cheroot) and some test, perhaps. Also, rebase this branch as I've posted a few fixes on `main`.

cherrypy/_cpwsgi_server.py
jatalahd: Cheroot dependency version updated in latest push.

jatalahd:
@webknjaz; In latest push, updated the cheroot minimum version to 11.1.0. Added also one unit test for each ssl adapter, had to add pyopenssl as a test dependency. Possibly made a mistake with git when pushing, but hopefully did not mess up anything.

tox.ini
```diff
@@ -104,6 +104,8 @@ extras =
   routes_dispatcher
   memcached_session
 whitelist_externals = mkdir
+deps =
+  pyopenssl
```
webknjaz: Instead of a dep list, could you add an `ssl` extra a few lines above?

cherrypy/test/test_passwd.pem
webknjaz, line 1: How difficult would it be to integrate `trustme` instead of committing test data into the repo?

cherrypy/test/test_config_ssl_adapter.py
```diff
@@ -0,0 +1,117 @@
...
+                ),
+                'server.ssl_private_key': os.path.join(
+                    thisDir,
+                    'test_passwd.pem',
+                ),
+                'server.ssl_certificate_chain': os.path.join(
+                    thisDir,
+                    'test_passwd.pem',
+                ),
+                'server.ssl_private_key_password': '123456',
```
webknjaz: Looking at this, I'm starting to question the interface — keeping secrets in long-living objects might not be the most secure decision (and decreasing security would probably defeat the purpose of putting passwords on keys).
I think it might be a good idea to allow this to also be a callable, not just a string. So that the end-users would be able to pass something that would retrieve the secret from a secure location. A function with a signature like `() -> str`. WDYT?

webknjaz: A more concrete suggestion: https://github.com/cherrypy/cheroot/pull/798/files#r2566163502

cherrypy/_cpwsgi_server.py
webknjaz, line 1: Looks like the same needs to happen here: https://github.com/cherrypy/cherrypy/blob/1f75bc9eed8e0e385f64f368bd69f58d96fb8c2b/cherrypy/_cpnative_server.py#L170-L187

jatalahd: Done in latest push.

cherrypy/test/test_passwd.pem
jatalahd: Done in latest push, but also reworked the whole test suite to use the more modern pytest.fixture approach. See what you think about it.

tox.ini
jatalahd: Done in latest push.

cherrypy/test/test_config_ssl_adapter.py
jatalahd: The cheroot commit is now merged, so I am not sure if anything needs to be done in cherrypy side?

jatalahd:
@webknjaz ; In the latest push, reworked the whole test suite to use the pytest.fixture approach and using parametrization, added tests for both native and wsgi servers with pyopenssl and builtin adapters. There is quite a lot of code to review in the test suite, but please provide comments to make it better.

Also for the native server, the existing tests had excluded win32 platform, but I did not do that for the new tests. If this needs to be considered, please let me know.

webknjaz:
@jatalahd Thanks for working on this but I had a few health incidents over the past days that force me to stay away from GH for a while. Please keep iterating where you can, but JFYI it may take a while until I'll be able to get to another round of review.

